PR #5, opened 2019-11-11
MAT-373 Move to go-dep dependency

jdkanani:
We need to test it on different platform like Linux and windows before we merge this.

Please don't merge it for now.


vaibhavchellani:
Oh I don't think we should move to `go dep` at all, we should move `heimdall` and `bor` to go mods now.

PR #8, opened 2019-12-03
MAT-467 retry span and states in bor

jdkanani:


consensus/bor/bor.go
```diff
@@ -17,7 +17,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/maticnetwork/bor"
+	ethereum "github.com/maticnetwork/bor"
```
vaibhavchellani: rename to bor

PR #10, opened 2019-12-04
log usage standardised

vaibhavchellani:


jdkanani:
@vaibhavchellani looks good

PR #21, opened 2020-02-23
MAT-865  Add subscribe deposit event filter

ptsayli:


consensus/bor/bor.go
```diff
@@ -1174,6 +1179,16 @@ func (c *Bor) CommitStates(
 			"txHash", eventRecord.TxHash,
 		)
 
+		go func() {
+			stateData := types.StateData{
```
jdkanani: Create an object outside of the go routine.

eth/backend.go
```diff
@@ -75,6 +75,7 @@ type Ethereum struct {
 	// Handlers
 	txPool          *core.TxPool
 	blockchain      *core.BlockChain
+	bor             *bor.Bor
```
jdkanani: Why do you need bor object here? It will fail in case of `clique` consensus.

eth/filters/api.go
```diff
@@ -233,6 +232,38 @@ func (api *PublicFilterAPI) NewHeads(ctx context.Context) (*rpc.Subscription, er
...
+	rpcSub := notifier.CreateSubscription()
+	go func() {
+		stateData := make(chan *types.StateData)
+		stateDataSub := api.events.SubscribeNewDeposits(stateData)
+
+		for {
+			select {
+			case h := <-stateData:
+				if crit.Did == h.Did || crit.Contract == h.Contract ||
+					(crit.Did == 0 && crit.Contract == common.Address{}) {
```
jdkanani: Please test this. This might not work. You have to check the bytes equality.
```
crit.Contract == common.Address{}
```

ethclient/ethclient.go
```diff
@@ -24,7 +24,7 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/maticnetwork/bor"
+	ethereum "github.com/maticnetwork/bor"
```
vaibhavchellani: maybe rename everything to bor? To have consistency

eth/filters/filter_system.go
```diff
@@ -68,6 +72,8 @@ const (
 	logsChanSize = 10
 	// chainEvChanSize is the size of channel listening to ChainEvent.
 	chainEvChanSize = 10
+	// stateEvChanSize is the size of channel listening to ChainEvent.
+	stateEvChanSize = 10
```
vaibhavchellani: Are you sure the number of messages to this channel won't exceed 10? Because if they do there will be data loss.

jdkanani: Won't be data loss, just stuck there till the previous 10 messages get broadcasted. @ptsayli please test and check.

eth/backend.go
jdkanani: Also, are we using it somewhere?

eth/filters/filter_system.go
vaibhavchellani: Yeah that's what I meant, you won't be able to push to channel hence the value will be discarded.

ethclient/ethclient.go
ptsayli: Do we require this change
just making changes in this file will not be enough
there are around 24 files which imports `ethereum "github.com/maticnetwork/bor"`
we need to make changes there too
@vaibhavchellani

eth/backend.go
ptsayli: Removing this
We are using `Ethereum.engine.bor`

eth/filters/api.go
ptsayli: If `Did` or `Contract` values is not set then is condition is working file
any specific case you want to refer here @jdkanani

eth/filters/filter_system.go
ptsayli: Tested
Working Fine

atvanguard: Yeah, in `stateCh:  make(chan core.NewStateChangeEvent, stateEvChanSize)` , `stateEvChanSize` specifies the number of elements that will be buffered in the channel, before it starts to block on events being consumed. So this should work.

consensus/bor/bor.go
atvanguard: Just curious, why does it matter?

ethclient/ethclient.go
atvanguard: Planning to do this in another separate PR.

PR #22, opened 2020-02-27
MAT-355: Add bor.isValidatorAction and call it while validating underpriced tx

atvanguard:


jdkanani:
@atvanguard Need to test it with multiple validators and after span change.

atvanguard:
Closing in favour of #24

PR #23, opened 2020-03-06
Remove unwanted log

atvanguard:


jdkanani:
We shouldn't change the warning. Instead, we should fix our log message to comply with the logger.

jdkanani:
```
log.Info("Retrying again in 5 seconds for next Heimdall span", "path", u.Path)
```

PR #33, opened 2020-03-19
Fix go mod for make all

jdkanani:


Makefile
```diff
@@ -10,30 +10,33 @@
 
 GOBIN = $(shell pwd)/build/bin
 GO ?= latest
+GORUN = env GO111MODULE=on go run
```
atvanguard: Let's remove this since `GO111MODULE=on` is not necessary as discussed on slack?
Then can replace `$(GORUN)` with `go run` everywhere?

jdkanani: Done.

PR #37, opened 2020-03-25
Fix polluted path and check data length for validation action

jdkanani:


consensus/bor/rest.go
```diff
@@ -25,18 +25,13 @@ type IHeimdallClient interface {
 }
 
 type HeimdallClient struct {
-	u      *url.URL
-	client http.Client
+	urlString string
```
vaibhavchellani: We dont need to mention data type with variable name here I think

jdkanani: Added it because there is `url` package imported. So used `urlString` to have a meaningful name without conflicting with the imported package.

consensus/bor/bor.go
```diff
@@ -1253,13 +1253,21 @@ func (c *Bor) IsValidatorAction(chain consensus.ChainReader, from common.Address
...
+	if tx.Data() == nil || len(tx.Data()) < 4 {
+		return false
+	}
+
 	return bytes.Compare(proposeSpanSig, tx.Data()[:4]) == 0 &&
 		tx.To().String() == validatorContract
 }
 
 func isProposeStateAction(tx *types.Transaction, stateReceiverContract string) bool {
 	// keccak256('proposeState(uint256)').slice(0, 4)
```
vaibhavchellani: Can we remove this comment?

jdkanani: I think it's better to understand what that code means. Easier to test/verify.

vaibhavchellani:
LGTM

PR #39, opened 2020-03-27
fix: `commitState` to tolerate failures by adding `finalizeState`

jdkanani:


jdkanani:
As discussed with @atvanguard, we might not need it if we can have a require-less `onStateReceive` call from `StateReceiver.sol` genesis contract.

jdkanani:
Canceled due to https://github.com/maticnetwork/genesis-contracts/pull/13

PR #51, opened 2020-04-27
MAT-1248: Audit fixes

atvanguard:


consensus/bor/bor.go
```diff
@@ -430,15 +439,6 @@ func (c *Bor) verifyCascadingFields(chain consensus.ChainReader, header *types.H
 		for i, validator := range currentValidators {
```
jdkanani, line 439: We should use this `if-block` for `(number-1)  % c.config.Sprint == 0`. 

This checks if the incoming block header has valid validator bytes or not.

consensus/bor/bor.go
```diff
@@ -582,27 +582,9 @@ func (c *Bor) verifySeal(chain consensus.ChainReader, header *types.Header, pare
...
-	if proposerIndex != tempIndex && limit > 0 {
-		if tempIndex < proposerIndex {
-			tempIndex = tempIndex + len(validators)
-		}
-
-		if tempIndex-proposerIndex > limit {
-			return errRecentlySigned
-		}
+	if _, err = snap.getSignerSuccessionNumber(signer, proposer); err != nil {
+		return err
```
jdkanani, line 587: Nice! Thanks.

consensus/bor/validator.go
```diff
@@ -80,18 +82,6 @@ func ValidatorListString(vals []*Validator) string {
...
-// as its redundant with the pubkey. This also excludes ProposerPriority
-// which changes every round.
-func (v *Validator) Bytes() []byte {
-	b, err := json.Marshal(v)
-	if err != nil {
-		return b
-	}
-	return nil
-}
-
```
jdkanani, line 94: Why this method is removed? This can be used in encoding/decoding while the validator is getting stored in the store.

consensus/bor/validator_set.go
```diff
@@ -182,7 +183,7 @@ func computeMaxMinPriorityDiff(vals *ValidatorSet) int64 {
 func (vals *ValidatorSet) getValWithMostPriority() *Validator {
 	var res *Validator
 	for _, val := range vals.Validators {
-		res = res.CompareProposerPriority(val)
+		res = res.Cmp(val)
```
jdkanani, line 186: Why are we using new function name?

consensus/bor/snapshot.go
```diff
@@ -200,6 +186,40 @@ func (s *Snapshot) apply(headers []*types.Header) (*Snapshot, error) {
 	return snap, nil
 }
 
+// getSignerSuccessionNumber returns the relative position of signer in terms of the in-turn proposer
+func (s *Snapshot) getSignerSuccessionNumber(signer common.Address, proposer common.Address) (int, error) {
```
jdkanani, line 190: We should write test cases for this function as it is error-prone.

consensus/bor/snapshot.go
```diff
@@ -200,6 +186,40 @@ func (s *Snapshot) apply(headers []*types.Header) (*Snapshot, error) {
...
+	validators := s.ValidatorSet.Validators
+	// bor.verifySeal and bor.Seal has the following commented out.
+	// TODO DISCUSS WITH JD if this is required
+	// If it is, we will also need to send in header.Number as parameter
+	// if number%c.config.Sprint != 0 {
+	// 	proposer = snap.Recents[number-1]
+	// }
```
jdkanani, line 197: No need. You can remove it.

consensus/bor/bor.go
atvanguard: Done.

consensus/bor/snapshot.go
atvanguard: Done.

PR #52, opened 2020-05-05
new: Add bor.getRootHash(start, end) api method (MAT-1302)

atvanguard:
JSON RPC API Reference:
```
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"eth_getRootHash","params":["0x1", "0xa"], "id":1}' http://localhost:8545

Over websocket,
wscat -c ws://localhost:8585
{"id":1, "method":"eth_getRootHash", "params":["0x1", "0xa"]}
```
Over IPC:
```
bor attach node1/bor.ipc
> bor.getRootHash(1, 10)
```

Expected response:
```
{"jsonrpc":"2.0","id":1,"result":"BXy0SRpYza0Uu42EOY8dIHXlW94HBjLBcLQu+XbIA7Q="}
```

### DIFF
--- a/consensus/bor/api.go
+++ b/consensus/bor/api.go
@@ -35,6 +35,7 @@ import (
 var (
 	// MaxCheckpointLength is the maximum number of blocks that can be requested for constructing a checkpoint root hash
 	MaxCheckpointLength = uint64(math.Pow(2, 15))
+	once                sync.Once
 )
 
 // API is a user facing RPC API to allow controlling the signer and voting
@@ -123,11 +124,18 @@ func (api *API) GetCurrentValidators() ([]*Validator, error) {
 
 // GetRootHash returns the merkle root of the start to end block headers
 func (api *API) GetRootHash(start int64, end int64) ([]byte, error) {
-	key := getKey(start, end)
-	if root, err := api.lookupCache(key); err != nil {
+	var err error
+	once.Do(func() {
+		if api.rootHashCache == nil {
+			api.rootHashCache, err = lru.NewARC(10)
+		}
+	})
+	if err != nil {
 		return nil, err
-	} else if root != nil {
-		return root, nil
+	}
+	key := getRootHashKey(start, end)
+	if root, known := api.rootHashCache.Get(key); known {
+		return root.([]byte), nil
 	}
 	length := uint64(end - start + 1)
 	if length > MaxCheckpointLength {
@@ -176,19 +184,6 @@ func (api *API) GetRootHash(start int64, end int64) ([]byte, error) {
 	return root, nil
 }
 
-func (api *API) lookupCache(key string) ([]byte, error) {
-	var err error
-	if api.rootHashCache == nil {
-		if api.rootHashCache, err = lru.NewARC(10); err != nil {
-			return nil, err
-		}
-	}
-	if root, known := api.rootHashCache.Get(key); known {
-		return root.([]byte), nil
-	}
-	return nil, nil
-}
-
-func getKey(start int64, end int64) string {
+func getRootHashKey(start int64, end int64) string {
 	return strconv.FormatInt(start, 10) + "-" + strconv.FormatInt(end, 10)
 }
--- a/consensus/bor/api.go
+++ b/consensus/bor/api.go
@@ -17,9 +17,12 @@
 package bor
 
 import (
+	"math"
 	"math/big"
+	"strconv"
 	"sync"
 
+	lru "github.com/hashicorp/golang-lru"
 	"github.com/maticnetwork/bor/common"
 	"github.com/maticnetwork/bor/consensus"
 	"github.com/maticnetwork/bor/core/types"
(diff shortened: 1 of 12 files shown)


internal/web3ext/web3ext.go
```diff
@@ -149,6 +149,11 @@ web3._extend({
 			call: 'bor_getCurrentValidators',
 			params: 0
 		}),
+		new web3._extend.Method({
+			name: 'getRootHash',
+			call: 'bor_getRootHash',
+			params: 2,
+		}),
```
jdkanani, line 156: Will this be also available over WebSocket?

consensus/bor/api.go
```diff
@@ -105,3 +111,41 @@ func (api *API) GetCurrentValidators() ([]*Validator, error) {
 	}
 	return snap.ValidatorSet.Validators, nil
 }
+
+// GetRootHash gets the current validators
+func (api *API) GetRootHash(start uint64, end uint64) ([]byte, error) {
+	blockHeaders := make([]*types.Header, end-start+1)
+	wg := new(sync.WaitGroup)
+	// do we want to limit the # of concurrent go routines?
```
jdkanani, line 119: Yes. That would be great. 10 goroutines would be sufficient.

consensus/bor/api.go
```diff
@@ -105,3 +111,41 @@ func (api *API) GetCurrentValidators() ([]*Validator, error) {
...
+	for i := start; i <= end; i++ {
+		wg.Add(1)
+		go func(number uint64) {
+			blockHeaders[number-start] = api.chain.GetHeaderByNumber(number)
```
jdkanani, line 123: `api.chain.GetHeaderByNumber` reads from the state right?

consensus/bor/api.go
```diff
@@ -105,3 +111,41 @@ func (api *API) GetCurrentValidators() ([]*Validator, error) {
 	}
 	return snap.ValidatorSet.Validators, nil
 }
+
+// GetRootHash gets the current validators
+func (api *API) GetRootHash(start uint64, end uint64) ([]byte, error) {
```
jdkanani, line 116: It should also check the latest block height as part of validation.

consensus/bor/api.go
```diff
@@ -105,3 +111,41 @@ func (api *API) GetCurrentValidators() ([]*Validator, error) {
...
+		copy(arr[:], header)
+		headers[i] = arr
+	}
+
+	tree := merkle.NewTreeWithOpts(merkle.TreeOptions{EnableHashSorting: false, DisableHashLeaves: true})
+	if err := tree.Generate(convert(headers), sha3.NewLegacyKeccak256()); err != nil {
+		return nil, err
+	}
+
+	return tree.Root().Hash, nil
```
jdkanani, line 150: We should cache the result in case the same API gets hit multiple times with the same `start/end` value. 

LRU (size: 10): `append(start, end) => result-hash` would work.

consensus/bor/api.go
jdkanani: We should also limit `end - start + 1` to avoid abuse. Example: `end - start + 1 < 16384`

atvanguard: What limit do we want to keep? This will prevent being able to send large checkpoints.

jdkanani: @atvanguard Yeah. That's why we want to make it big but not too big where one can call these APIs with huge checkpoint and cause the node stoppage or something.

We can either limit APIs call time or something.

internal/web3ext/web3ext.go
atvanguard: Added a method in ethApi, so now it is. Checkout the PR description.

consensus/bor/api.go
atvanguard: Limited to 20 goRoutes.

consensus/bor/api.go
```diff
@@ -105,3 +120,75 @@ func (api *API) GetCurrentValidators() ([]*Validator, error) {
...
+		copy(arr[:], header)
+		headers[i] = arr
+	}
+
+	tree := merkle.NewTreeWithOpts(merkle.TreeOptions{EnableHashSorting: false, DisableHashLeaves: true})
+	if err := tree.Generate(convert(headers), sha3.NewLegacyKeccak256()); err != nil {
+		return nil, err
+	}
+	root := tree.Root().Hash
+	api.rootHashCache.Add(key, root)
```
jdkanani, line 175: `api.rootHashCache` can be nil. Let's create `getRootHashCache` which uniquely creates a root hash instance. Something like this: https://github.com/maticnetwork/heimdall/blob/c5aeaa63508219af573b64035265e4f8aace00a9/auth/types/pulp.go#L24-L33

consensus/bor/api.go
```diff
@@ -105,3 +120,75 @@ func (api *API) GetCurrentValidators() ([]*Validator, error) {
...
+			return nil, err
+		}
+	}
+	if root, known := api.rootHashCache.Get(key); known {
+		return root.([]byte), nil
+	}
+	return nil, nil
+}
+
+func getKey(start int64, end int64) string {
```
jdkanani, line 192:
```suggestion
func getRootHashKey(start int64, end int64) string {
```

atvanguard: Done.

consensus/bor/api.go
atvanguard: Done.

PR #54, opened 2020-05-08
chg: Cancel active sealing op when a new chain segment is received (MAT-1116)

atvanguard:


consensus/bor/bor.go
```diff
@@ -774,23 +783,62 @@ func (c *Bor) Seal(chain consensus.ChainReader, block *types.Block, results chan
...
 	}()
-
 	return nil
 }
 
+// WaitForSealingOp blocks until delay elapses or stop signal is received
+func (c *Bor) WaitForSealingOp(stop <-chan struct{}, shouldSeal chan bool, number uint64, delay time.Duration) {
+	ctx, cancel := context.WithCancel(context.Background())
+	ctx = context.WithValue(ctx, blockNumberKey, number)
+	ctx = context.WithValue(ctx, listenerKey, shouldSeal)
```
atvanguard, line 822: I could change these right?

consensus/bor/bor.go
```diff
@@ -774,23 +783,62 @@ func (c *Bor) Seal(chain consensus.ChainReader, block *types.Block, results chan
 
 	// Wait until sealing is terminated or delay timeout.
 	log.Trace("Waiting for slot to sign and propagate", "delay", common.PrettyDuration(delay))
+	shouldSeal := make(chan bool)
+	go c.WaitForSealingOp(stop, shouldSeal, number, delay)
 	go func() {
-		select {
-		case <-stop:
+		switch <-shouldSeal {
```
atvanguard, line 789: Should I `close(shouldSeal)` the channel immediately after this?

tokamak-git: if it is not going to be used anywhere else then yes, else defer would be a good idea.

consensus/bor/bor.go
```diff
@@ -138,6 +138,13 @@ var (
 	errRecentlySigned = errors.New("recently signed")
 )
 
+type ActiveSealingOp struct {
+	ctx    context.Context
+	cancel context.CancelFunc
```
jdkanani, line 143: Why it needs `ctx` and `cancel`? Why not just `blockNumber` and `shouldSeal`?

consensus/bor/bor.go
```diff
@@ -774,23 +783,62 @@ func (c *Bor) Seal(chain consensus.ChainReader, block *types.Block, results chan
 
 	// Wait until sealing is terminated or delay timeout.
 	log.Trace("Waiting for slot to sign and propagate", "delay", common.PrettyDuration(delay))
+	shouldSeal := make(chan bool)
+	go c.WaitForSealingOp(stop, shouldSeal, number, delay)
```
jdkanani, line 787: Shouldn't we cancel the previous `SealingOp` before we wait for new? Or is it done by the `stop` channel automatically?

atvanguard: Yes, the method that invokes this closes stop channel which causes the active one to stop before new sealing starts. https://github.com/maticnetwork/bor/blob/develop/miner/worker.go#L525

consensus/bor/bor.go
atvanguard: Aah I don't need to use a cancellable object at all. Changed.

consensus/bor/bor.go
atvanguard: Done.

consensus/bor/bor.go
atvanguard: Actually I don't need to use a context at all. Removed it.

core/types.go
```diff
@@ -49,3 +51,10 @@ type Processor interface {
 	// the processor (coinbase) and any included uncles.
 	Process(block *types.Block, statedb *state.StateDB, cfg vm.Config) (types.Receipts, []*types.Log, uint64, error)
 }
+
+// bor acts as a way to be able to type cast consensus.Engine;
+// since importing "github.com/maticnetwork/bor/consensus/bor" results in a cyclic dependency
+type bor interface {
```
tokamak-git, line 57: name could be `consense` instead of bor? interface being the verbe here.

atvanguard: Removed this altogether. Not the right place to be anyway.

consensus/bor/bor.go
```diff
@@ -774,23 +781,64 @@ func (c *Bor) Seal(chain consensus.ChainReader, block *types.Block, results chan
...
+		shouldSeal <- false
+	case <-time.After(delay):
+		shouldSeal <- true
+	}
+}
+
+// CancelActiveSealingOp cancels in-flight sealing process
+func (c *Bor) CancelActiveSealingOp() {
+	if c.activeSealingOp != nil {
+		log.Info("Discarding active sealing operation", "number", c.activeSealingOp.number)
```
atvanguard, line 837: Should this be a debug statement instead?

jdkanani: Yes

atvanguard: Done.

PR #55, opened 2020-05-11
miner/syncer fix

atvanguard:


atvanguard:
This PR keeps track of multiple sealing operations, but having concurrent sealing operations doesn't make sense.

PR #57, opened 2020-05-13
new: Add wiggle to block time (MAT-1353)

atvanguard:


consensus/bor/bor.go
```diff
@@ -334,17 +324,6 @@ func (c *Bor) verifyHeader(chain consensus.ChainReader, header *types.Header, pa
 	}
 	number := header.Number.Uint64()
 
-	var parent *types.Header
```
atvanguard, line 337: This has moved to verifySeal.

consensus/bor/snapshot.go
```diff
@@ -19,6 +19,7 @@ package bor
 import (
 	"bytes"
 	"encoding/json"
+	// "fmt"
```
jdkanani, line 22: Remove unwanted comments.

PR #71, opened 2020-05-27
new: add options to setup different node types

jdkanani:
With this PR, Bor Linux package will support different node types:

1. `sentry` - Sentry node
2. `validator` - Validator with sentry node
3. `validator-without-sentry` - Validator without sentry node

To select node type, edit `NODE_TYPE` in  `/etc/bor/metadata` after package is installed. Possible value for `NODE_TYPE`: `sentry`, `validator` and `vaidator-without-sentry`

### DIFF
--- a/.github/workflows/linuxpackage.yml
+++ b/.github/workflows/linuxpackage.yml
@@ -93,7 +93,7 @@ jobs:
             "
           fi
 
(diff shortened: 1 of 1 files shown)


jdkanani:
Merging on @anurag-arjun's approval.

PR #86, opened 2020-08-11
Fix deposit subscribe

ptsayli:


core/blockchain.go
```diff
@@ -1542,10 +1549,6 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 		if emitHeadEvent {
 			bc.chainHeadFeed.Send(ChainHeadEvent{Block: block})
 		}
-		syncData := block.StateSyncData()
```
jdkanani, line 1545: Why removed from here? Shouldn't be after `bc.chainFeed.Send`?

core/types/block.go
```diff
@@ -66,6 +66,14 @@ func (n *BlockNonce) UnmarshalText(input []byte) error {
 	return hexutil.UnmarshalFixedText("BlockNonce", input, n[:])
 }
 
+// StateData represents state received from Ethereum Blockchain
+type StateData struct {
```
jdkanani, line 70: Create new `core/types/state-data.go` file for struct. It's better we don't touch block.go

PR #90, opened 2020-10-19
Add state-sync logs into new tx/receipt

jdkanani:
This PR adds support for persistent state-sync logs. It contains:

* Introduce block receipt concept
* Store state-sync logs in block receipt (same log format as tx receipt log)
* Use ancient storage to store old block receipts (same as other entities like block, header, transactions, receipts, and logs)
* Introduce new methods `eth_getBorBlockReceipt` and `eth_getBorBlockLogs` (with filter) to retrieve block receipts and block logs
* Backward compatibility for already synced nodes. Only newly synced nodes will have block logs, but new block logs will be processed on all nodes. So, if the node needs all (including past) block logs, it has to sync again.
* `eth_getLogs`, `eth_getTransactionReceipt` and `eth_getBlockByNumber` will include new tx/receipt with hash`keccak256("matic-bor-receipt-" + block number + block hash)`


core/rawdb/bor_receipt.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	if has, err := db.Has(borReceiptKey(number, hash)); !has || err != nil {
+		return false
+	}
+
+	return true
+}
+
+// ReadBorReceiptRLP retrieves all the transaction receipts belonging to a block in RLP encoding.
```
nachomazzara, line 37: `all the transaction receipts` or `block receipts`?

eth/bor_api_backend.go
```diff
@@ -0,0 +1,42 @@
...
+		return nil, ethereum.NotFound
+	}
+
+	return receipt, nil
+}
+
+func (b *EthAPIBackend) GetBorBlockLogs(ctx context.Context, hash common.Hash) ([]*types.Log, error) {
+	receipt := b.eth.blockchain.GetBorReceiptByHash(hash)
+	if receipt == nil {
+		return nil, nil
```
nachomazzara, line 39: Should it be 
```go
return nil, ethereum.NotFound
```

eth/filters/bor_filter.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, nil
+	}
+	head := header.Number.Uint64()
+
+	if f.begin == -1 {
+		f.begin = int64(head)
+	}
+
+	// adjust begin for sprint
+	f.begin = nextSprintEnd(f.begin)
```
nachomazzara, line 100: It shouldn't be the `previousSprintEnd`?

eth/filters/bor_filter.go
```diff
@@ -0,0 +1,154 @@
...
+func (f *BorBlockLogsFilter) borBlockLogs(ctx context.Context, receipt *types.BorReceipt) (logs []*types.Log, err error) {
+	if bloomFilter(receipt.Bloom, f.addresses, f.topics) {
+		found := filterLogs(receipt.Logs, nil, nil, f.addresses, f.topics)
+		logs = append(logs, found...)
+	}
+	return logs, nil
+}
+
+func nextSprintEnd(n int64) int64 {
+	m := n % 64
```
nachomazzara, line 148: I didn't look through the whole codebase but if `64` is the sprint value, maybe it would be a good idea moving it to a variable/constant

ethclient/bor_ethclient.go
```diff
@@ -0,0 +1,30 @@
...
+		return "", err
+	}
+	return rootHash, nil
+}
+
+// GetBorBlockReceipt returns bor block receipt
+func (ec *Client) GetBorBlockReceipt(ctx context.Context, hash common.Hash) (*types.BorReceipt, error) {
+	var r *types.BorReceipt
+	err := ec.c.CallContext(ctx, &r, "eth_getBorBlockReceipt", hash)
+	if err == nil {
```
nachomazzara, line 24: Small comment (I'm not a golang dev):

Is not the same as :
```go
if err == nil && r == nil { 
   return nil, ethereum.NotFound
}
```

nachomazzara (commented):
🚀  Thanks JD & Matic team for all the effort! 

Just comments

ethclient/bor_ethclient.go
```diff
@@ -0,0 +1,30 @@
...
+
+	ethereum "github.com/maticnetwork/bor"
+	"github.com/maticnetwork/bor/common"
+	"github.com/maticnetwork/bor/core/types"
+)
+
+// GetRootHash returns the merkle root of the block headers
+func (ec *Client) GetRootHash(ctx context.Context, startBlockNumber uint64, endBlockNumber uint64) (string, error) {
+	var rootHash string
+	if err := ec.c.CallContext(ctx, &rootHash, "eth_getRootHash", startBlockNumber, endBlockNumber); err != nil {
```
itzmeanjan, line 14: Is it that `startBlockNumber` <= `endBlockNumber` ? If yes, we can put a check, if it's not handled in called function.

eth/filters/bor_filter.go
```diff
@@ -0,0 +1,154 @@
...
+		logs = append(logs, found...)
+	}
+	return logs, nil
+}
+
+// borBlockLogs returns the logs matching the filter criteria within a single block.
+func (f *BorBlockLogsFilter) borBlockLogs(ctx context.Context, receipt *types.BorReceipt) (logs []*types.Log, err error) {
+	if bloomFilter(receipt.Bloom, f.addresses, f.topics) {
+		found := filterLogs(receipt.Logs, nil, nil, f.addresses, f.topics)
+		logs = append(logs, found...)
```
imyourm8, line 142: This append here is unnecessary. Especially when it is followed by another append on the line 133.

eth/filters/bor_filter.go
imyourm8: This function name is confusing. It sounds like it gets the sprint end of the range, and nothing to do with previous or next sprint is that. If it gets 100 as a number, it will return the closest end of 128.

eth/filters/bor_filter.go
```diff
@@ -0,0 +1,154 @@
...
+		end = int64(head)
+	}
+
+	// Gather all indexed logs, and finish with non indexed ones
+	return f.unindexedLogs(ctx, uint64(end))
+}
+
+// unindexedLogs returns the logs matching the filter criteria based on raw block
+// iteration and bloom matching.
+func (f *BorBlockLogsFilter) unindexedLogs(ctx context.Context, end uint64) ([]*types.Log, error) {
```
imyourm8, line 113: Not sure that mixing fields and passing argument is a good idea long term. You are passing end here as a part of the range, but you have .end field in the filters. At the same time inside .begin field is used. I would stick with using fields in this case.

ethclient/bor_ethclient.go
itzmeanjan: Can we not do it this way ?

```go
var r types.BorReceipt
if err := ec.c.CallContext(ctx, &r, "eth_getBorBlockReceipt", hash); err != nil {
        return nil, err
}

return &r, err
```

eth/bor_api_backend.go
itzmeanjan, line 39: We can't return this error to as `nil`, we can do it 

```go
return nil, errors.New("...some meaning ful error message...")
```

Or as @nachomazzara suggested.

core/rawdb/bor_receipt.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"github.com/maticnetwork/bor/common"
+	"github.com/maticnetwork/bor/core/types"
+	"github.com/maticnetwork/bor/ethdb"
+	"github.com/maticnetwork/bor/log"
+	"github.com/maticnetwork/bor/rlp"
+)
+
+var (
+	borReceiptPrefix = []byte("bor-receipt-") // borReceiptPrefix + number + block hash -> bor block receipt
```
itzmeanjan, line 12: What this `number` is going to be ?

itzmeanjan (commented):
Added some comments

eth/filters/bor_filter.go
jdkanani: @nachomazzara @imyourm8 Sorry for the confusion. It's `currentSprintEnd` actually. For `number = 128`, it will return  128, but for 100 it will return 128 as @imyourm8 mentioned. I will make the change.

eth/bor_api_backend.go
jdkanani: I wanted to return `[]` (empty array) instead of error since it is possible that the receipt will not be present if block logs are not present.

core/rawdb/bor_receipt.go
jdkanani: It's a typo. I will fix it.

eth/filters/bor_filter.go
jdkanani: Genesis has a sprint variable. But it seems we cannot change it after this change because it will incur huge changes if sprint changes in the future. Either we remove it from genesis and make it permanent. Thoughts? @imyourm8

ethclient/bor_ethclient.go
jdkanani: I will make the change. Thanks.

eth/filters/bor_filter.go
jdkanani: Ah! got it. Thanks for catching it.

eth/filters/bor_filter.go
jdkanani: It's a common practice from normal log filter since `f.end` could be the latest block, and with subscription, `f.end` should remain the same throughout the subscription. That way same filter object can be used for future subscribed logs. I will look more into it and comment back.

I am making some changes in `eth_getLogs`, so this will be also used in the subscription. I will update you here.

ethclient/bor_ethclient.go
jdkanani: I will check and fix.

core/rawdb/bor_receipt.go
jdkanani: It's a block number.

itzmeanjan: I think we can make the name `blockNumber`, makes it more intuitive.

eth/filters/bor_filter.go
nachomazzara: It means that what happened or is expected on block 100 in Ethereum will be covered by a bor receipt at the beginning of 128 in Matic, right? 

I was wondering if it will be a common use case to set the begin block to get the current sprint start instead of the end (with 100, get 64 instead of 128)

ethclient/bor_ethclient.go
jdkanani: Check is there at https://github.com/maticnetwork/bor/blob/develop/consensus/bor/api.go#L156-L158

eth/filters/bor_filter.go
jdkanani: Fixed.

ethclient/bor_ethclient.go
jdkanani: Fixed.

eth/filters/bor_filter.go
jdkanani: Fixed.

core/blockchain.go
```diff
@@ -1460,8 +1462,9 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 		})
 
 		if len(blockLogs) > len(logs) {
-			rawdb.WriteBorReceipt(blockBatch, block.Hash(), block.NumberU64(), &types.BorReceiptForStorage{
-				Logs: blockLogs[len(logs):], // get state-sync logs from `state.Logs()`
+			rawdb.WriteBorReceipt(blockBatch, block.Hash(), block.NumberU64(), &types.ReceiptForStorage{
+				Status: types.ReceiptStatusSuccessful, // make receipt status successful
```
itzmeanjan, line 1475: Is `Status` something which denotes status of deposit/ state sync ?

core/rawdb/bor_receipt.go
```diff
@@ -9,15 +11,19 @@ import (
 )
 
 var (
-	borReceiptPrefix = []byte("bor-receipt-") // borReceiptPrefix + number + block hash -> bor block receipt
+	// bor receipt key
+	borReceiptKey = types.BorReceiptKey
+
+	// borTxLookupPrefix + hash -> transaction/receipt lookup metadata
+	borTxLookupPrefix = []byte("matic-bor-tx-lookup-")
```
itzmeanjan, line 21: Does it denote, we can query state sync log using `append([]byte("matic-bor-tx-lookup-"), hash.Bytes()...)` , where `hash` is bor internal ( i.e. system call ) tx hash 🤔 ?

core/types/bor_receipt.go
```diff
@@ -1,109 +1,73 @@
...
-		// r.Logs[j].TxHash = txHash
-		r.Logs[j].TxIndex = uint(0)
-		r.Logs[j].Index = uint(j)
-	}
-	return nil
+// MergeBorLogs merges receipt logs and block receipt logs
+func MergeBorLogs(logs []*Log, borLogs []*Log) []*Log {
+	result := append(logs, borLogs...)
+	sort.SliceStable(result, func(i int, j int) bool {
+		return (result[i].BlockNumber*100000 + uint64(result[i].Index)) < (result[j].BlockNumber*100000 + uint64(result[j].Index))
```
itzmeanjan, line 87: Why did we multiply `blocknumber` with 10^5 ? I think we can make it compile time constant.

core/blockchain.go
jdkanani: `Status` is about the tx status. In our case, it will be always successful as we are adding a new "fake" receipt for state-sync.

core/rawdb/bor_receipt.go
jdkanani: Yes, this is a reverse lookup to get the block number for the given bor receipt hash.

core/types/bor_receipt.go
jdkanani: We are sorting by `block * 100000 + logIndex`. A simple hack to sort easily. I will make it constant.

itzmeanjan: We need to  make `10 ** 5` constant.

itzmeanjan (changes requested):
Please check comments.

itzmeanjan (approved):
:+1:

eth/filters/bor_filter.go
imyourm8: Genesis is not supposed to be changed, right? I think it is fine to get sprint value from genesis. It also helps to avoid incorrect behaviour if somebody wants to run node with a different sprint value but wasn't aware of hardcoded constant in the code.

jdkanani: Fixed. Using genesis for sprint value.

imyourm8 (approved):

PR #101, opened 2021-01-06
fix: ignore block filter error for bor logs

jdkanani:
Issue: Blockhash based filter throwing an error when bor block receipt doesn't exist. 

This PR will fix by ignoring error while filtering logs by block hash.

### DIFF
--- a/eth/filters/bor_filter.go
+++ b/eth/filters/bor_filter.go
@@ -18,7 +18,6 @@ package filters
 
 import (
 	"context"
-	"errors"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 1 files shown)


eth/filters/bor_filter.go
```diff
@@ -76,13 +75,9 @@ func newBorBlockLogsFilter(backend Backend, sprint uint64, addresses []common.Ad
 func (f *BorBlockLogsFilter) Logs(ctx context.Context) ([]*types.Log, error) {
 	// If we're doing singleton block filtering, execute and return
 	if f.block != (common.Hash{}) {
-		receipt, err := f.backend.GetBorBlockReceipt(ctx, f.block)
-		if err != nil {
-			return nil, err
-		}
-
+		receipt, _ := f.backend.GetBorBlockReceipt(ctx, f.block)
 		if receipt == nil {
```
itzmeanjan, line 79: Why aren't we taking `error` under consideration ? Is it due to how the called function returns its value ?

jdkanani: Because most of the time, bor block receipt will be `nil` when there will be no logs for bor state sync. And it's only one type - `not found`

itzmeanjan: Oh ok 👍

itzmeanjan (approved):

PR #107, opened 2021-01-28
fix: use hash instead of block number incase of reorg

jdkanani:


imyourm8 (approved):

consensus/bor/bor.go
```diff
@@ -892,13 +895,16 @@ func (c *Bor) GetCurrentSpan(headerHash common.Hash) (*Span, error) {
...
-func (c *Bor) GetCurrentValidators(snapshotNumber uint64, blockNumber uint64) ([]*Validator, error) {
+func (c *Bor) GetCurrentValidators(headerHash common.Hash, blockNumber uint64) ([]*Validator, error) {
 	// block
-	blockNr := rpc.BlockNumber(snapshotNumber)
+	blockNr := rpc.BlockNumberOrHashWithHash(headerHash, false)
 
 	// method
 	method := "getBorValidators"
 
+	ctx, cancel := context.WithCancel(context.Background())
```
imyourm8, line 905: What is this for?

jdkanani: Normal `context.Background()` never gets cancelled automatically. It's better to cancel we are done with using it. Forgot to use it before!

imyourm8: Got it

imyourm8 (approved):

consensus/bor/bor.go
```diff
@@ -892,13 +895,16 @@ func (c *Bor) GetCurrentSpan(snapshotNumber uint64) (*Span, error) {
 }
 
 // GetCurrentValidators get current validators
-func (c *Bor) GetCurrentValidators(snapshotNumber uint64, blockNumber uint64) ([]*Validator, error) {
+func (c *Bor) GetCurrentValidators(headerHash common.Hash, blockNumber uint64) ([]*Validator, error) {
```
itzmeanjan, line 898: This function caller should rather pass `context` as first argument, instead of 

```go
ctx, cancel := context.WithCancel(context.Background())
```

But we can do it later too.

itzmeanjan (approved):
Just a comment

PR #110, opened 2021-03-09
Update Geth to v1.10.1

jdkanani:
- [x] Merge upstream v1.10.1 to master
- [x] Fix conflicts
- [x] Check following changes manually
     - [x] Consensus
     - [x] Fee receipts
     - [x] State sync
     - [x] Bor receipt logs
     - [x] Import/Export blocks
- [x] Sync node on current Mumbai/Mainnet


itzmeanjan (approved):
Too many file changes, haven't gone through all though.

Krishna-1604:
I have done few tests with this branch on my local, Here are the details.
1. I was only running bor nodes, No heimdall for now
2. Started with only one node but later added more and finally managed to run 4 bor nodes on local in sync.
3. Able to do basic value txns, SC deployment, SC related txns with events.
4. How do I know they were in sync ? Because I did txns on one of them and was able to fetch the event logs from other nodes.

Few things like state-sync which are dependent on heimdall still needs to be tested, Working on it.

cc @jdkanani @ssandeep

Krishna-1604:
I am able to setup this branch with heimdall node as well, I can confirm that **StateSync** is working.
Although I am not able to get ` Bor receipt logs`, Not sure how to exactly generate them on local.

ssandeep (approved):

Krishna-1604:
Update : Able to generate logs as well on the receiver side (bor) using event.

ssandeep:
@jdkanani @GraphicalDot , I have validator and sentry nodes running Mumbai but without any stake. Can try syncing there but I think we should test with validators having stake as well.
cc: @Krishna-1604

Krishna-1604:
As discussed with @jdkanani and @ssandeep, I am going to test the following test cases on local first
1. Run a bor node(s) on local with old code, Let it generate some data(blocks), Build the bor with berlin fork code now, Run a node(s) now and let is sync, Make sure it is able to sync-up with all the old data correctly.
2. Try to add some dummy txns as well in the old data to create a testnet scenario, Also run the above case with multiple nodes.
3. The nodes should also be running fine even after berlinBlock (which I will set to ~2-3K for testing purposes).

@ssandeep Please add if anything is missed.

0xKrishna:
Above mentioned cases are tested on local with multiple nodes and working fine, Also tested following cases
1. Tested multiple scenarios with 4 nodes in config [old:new] of [1:3], [2:2], [3:1]
2. Confirmation of berlin fork on new code node using local logger
3. Txn submission on old code node after berlin fork block
4. Txn submission on new code node after berlin fork block

0xKrishna:
**Update on Mumbai network :**

We started a node on the Mumbai network (with no stake), The initial sync worked fine, and then the node worked properly till the berlinBlock (12970000) for that node only. Exactly after that block fork happened on that node and it stopped working/syncing with other peers.

cc @jdkanani @ssandeep

PR #115, opened 2021-04-13
Use correct go mod path

af-afk:


af-afk:
Change import paths across the board to use `maticnetwork/bor` instead of `go-ethereum`, #117

PR #117, opened 2021-04-13
Change import paths to use maticnetwork/bor instead of ethereum/go-ethereum

af-afk:


nnqq:
Why it ignored? This fix should be in `master` as the first thing after the fork. Now there is no way to `go get github.com/maticnetwork/bor`

ssandeep:
Since we do not do `go get github.com/maticnetwork/bor` and considering that there will be too many merge conflicts every time we need to get changes from upstream geth, we will be sticking to geth imports.

PR #145, opened 2021-07-01
validate checkpoint

zhiqiangxu:
validate validators set for checkpoint, otherwise evil nodes can do bad things.

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -411,7 +411,7 @@ func (c *Bor) verifyCascadingFields(chain consensus.ChainHeaderReader, header *t
 		for _, validator := range newValidators {
 			validatorBytes = append(validatorBytes, validator.HeaderBytes()...)
 		}
-		if !bytes.Equal(validatorBytes, header.Extra[extraVanity:len(parent.Extra)-extraSeal]) {
+		if !bytes.Equal(validatorBytes, header.Extra[extraVanity:len(header.Extra)-extraSeal]) {
 			return errors.New("invalid validators")
 		}
 	}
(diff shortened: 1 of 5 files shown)


consensus/bor/bor.go
```diff
@@ -399,6 +399,22 @@ func (c *Bor) verifyCascadingFields(chain consensus.ChainHeaderReader, header *t
 		return err
 	}
 
+	if (number+1)%c.config.Sprint == 0 {
```
jdkanani, line 402: It's already checking it from previous block's header and that header is built from the `BorValidatorSet` contract. https://github.com/maticnetwork/bor/blob/master/consensus/bor/bor.go#L403-L417

consensus/bor/bor.go
```diff
@@ -399,6 +399,22 @@ func (c *Bor) verifyCascadingFields(chain consensus.ChainHeaderReader, header *t
...
+		newValidators, err := c.GetCurrentValidators(header.ParentHash, number+1)
```
jdkanani, line 403: I am not sure if this works in full sync mode. That was the reason we tweaked it to check validators from previous header.

jdkanani (commented):
Thanks a lot for the PR. Added few comments.

consensus/bor/bor.go
zhiqiangxu: the first block of a sprint will takes validators from last checkpoint, without validation.

consensus/bor/bor.go
zhiqiangxu: I think you mean the check of `validateHeaderExtraField`, but it only checks length, doesn't validate against the actual validators.

temaniarpit27:
@zhiqiangxu Please provide the latest update for this

temaniarpit27:
Closing this for now

PR #160, opened 2021-07-29
fix: use block hash from the query to get bor tx

imyourm8:


nachomazzara:
Any update on this?

miguelmota:
Wondering the same. Seems to be a blocker for https://github.com/graphprotocol/graph-node/issues/2552

imyourm8:
So far I'm not able to find a reason for this to happen. I will continue to look into it, but any help will be appreciated on this front. Specifically, a scenario when it might happen.

jdkanani (approved):

PR #164, opened 2021-08-10
Implement getTransactionReceipts

temaniarpit27:


jdkanani:
@temaniarpit27 few things -- 

1. Does it contain a bor receipt? If not, we should add it.
2. Can we convert it to `bor_` instead of `eth_`?
3. Allow it to have receipts by both block number and block hash by using `blockNrOrHash rpc.BlockNumberOrHash` as param instead of just `blockNumber`. We can rename the method to `bor_getTransactionReceiptsByBlock`

internal/ethapi/api.go
```diff
@@ -531,6 +531,129 @@ func NewPublicBlockChainAPI(b Backend) *PublicBlockChainAPI {
...
+			fields["status"] = hexutil.Uint(receipt.Status)
+		}
+		if receipt.Logs == nil {
+			fields["logs"] = [][]*types.Log{}
+		}
+		// If the ContractAddress is 20 0x0 bytes, assume it is not a contract creation
+		if receipt.ContractAddress != (common.Address{}) {
+			fields["contractAddress"] = receipt.ContractAddress
+		}
+		fields = s.appendRPCMarshalBorTransaction(ctx, block, fields, true)
```
jdkanani, line 589: This won't add bor tx receipt. It will just append bor transaction.

internal/ethapi/api.go
```diff
@@ -531,6 +531,129 @@ func NewPublicBlockChainAPI(b Backend) *PublicBlockChainAPI {
 	return &PublicBlockChainAPI{b}
 }
 
+// GetTransactionReceipt returns the transaction receipt for the given transaction hash.
+func (s *PublicBlockChainAPI) GetTransactionReceiptsByBlockNumber(ctx context.Context, blockNr rpc.BlockNumber) ([]map[string]interface{}, error) {
+	blockNumber := uint64(blockNr.Int64())
+	blockHash := rawdb.ReadCanonicalHash(s.b.ChainDb(), blockNumber)
```
jdkanani, line 537: Also, you don't need to methods (byBlockNumber or byBlockHash). One method can take block number or block hash. 

```
func (s *PublicBlockChainAPI) GetTransactionReceiptsByBlock(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) ([]map[string]interface{}, error) {
      // Retrieve the block to act as the gas ceiling
		block, err := b.BlockByNumberOrHash(ctx, blockNrOrHash)
		if err != nil {
			return nil, err
		}
                 ... 
}
```

jdkanani (changes requested):

temaniarpit27:
@jdkanani @ssandeep please review

jdkanani (approved):

ssandeep (approved):

PR #171, opened 2021-08-19
v1.10.7

temaniarpit27:


jdkanani:
@temaniarpit27 We also need to change docker changes you made in v1.10.6 branch

temaniarpit27:
> @temaniarpit27 We also need to change docker changes you made in v1.10.6 branch

which change
it is the same as our master docker file

PR #172, opened 2021-08-23
fix: delete check empty tx root hash

tn606024:
When i use ethclient to call BlockByHash or BlockByNumber at specify tx will get `server returned non-empty transaction list but block header indicates no transactions ` error,  like https://mumbai.polygonscan.com/block/17442944. That because if block only exist plasma deposit tx transaction will trigger this error. So if the block is legal, i think this check can be deleted.


0xKrishna:
I think this is valid because in our case the header might show empty (no txns) but there are state-sync txns in the block.
I will check, Others can also confirm.

0xKrishna (approved):

ssandeep (approved):

PR #185, opened 2021-09-02
Network flags

0xKrishna:
With this implemented, Now you can start bor with
Mumbai network
```
bor --bor-mumbai
```
and Mainnet
```
bor --bor-mainnet
```

Please use `--bor.heimdall` if not using local heimdall.

All the default values from [start.sh](https://github.com/maticnetwork/launch/blob/master/testnet-v4/sentry/sentry/bor/start.sh) are already added for both the networks, Any extra flag can be added as usual (Overwrite will also work)

Refer : 
https://github.com/maticnetwork/bor/blob/krishna/network-flags/cmd/geth/config.go#L342
https://github.com/maticnetwork/bor/blob/krishna/network-flags/cmd/geth/config.go#L365


ssandeep (approved):

cmd/geth/config.go
```diff
@@ -328,3 +338,49 @@ func setAccountManagerBackends(stack *node.Node) error {
...
+
+func setDefaultBorMainnetGethConfig(ctx *cli.Context, config *gethConfig) {
+	config.Node.P2P.ListenAddr = fmt.Sprintf(":%d", 30303)
+	config.Node.HTTPHost = "0.0.0.0"
+	config.Node.HTTPVirtualHosts = []string{"*"}
+	config.Node.HTTPCors = []string{"*"}
+	config.Node.HTTPPort = 8545
+	config.Node.IPCPath = utils.MakeDataDir(ctx) + "/bor.ipc"
+	config.Node.HTTPModules = []string{"eth", "net", "web3", "txpool", "bor"}
+	config.Eth.SyncMode = downloader.FullSync
```
jdkanani, line 373:
```suggestion
	config.Eth.SyncMode = downloader.SnapSync
```

cmd/geth/config.go
```diff
@@ -328,3 +338,49 @@ func setAccountManagerBackends(stack *node.Node) error {
...
+
+func setDefaultMumbaiGethConfig(ctx *cli.Context, config *gethConfig) {
+	config.Node.P2P.ListenAddr = fmt.Sprintf(":%d", 30303)
+	config.Node.HTTPHost = "0.0.0.0"
+	config.Node.HTTPVirtualHosts = []string{"*"}
+	config.Node.HTTPCors = []string{"*"}
+	config.Node.HTTPPort = 8545
+	config.Node.IPCPath = utils.MakeDataDir(ctx) + "/bor.ipc"
+	config.Node.HTTPModules = []string{"eth", "net", "web3", "txpool", "bor"}
+	config.Eth.SyncMode = downloader.FullSync
```
jdkanani, line 350:
```suggestion
	config.Eth.SyncMode = downloader.SnapSync
```

jdkanani (changes requested):

cmd/geth/config.go
0xKrishna: Okay, So we will use SnapSync as default ?
I used full because that was the default in our start.sh

ssandeep (dismissed):

cmd/geth/config.go
0xKrishna: @jdkanani Done, Please check.

0xKrishna:
@jdkanani I have done the requested changes, Review once available.

mobile/geth.go
```diff
@@ -179,6 +179,20 @@ func NewNode(datadir string, config *NodeConfig) (stack *Node, _ error) {
...
+				config.EthereumNetworkID = 137
+			}
+		}
+		// If we have the Mumbai testnet, hard code the chain configs too
+		if config.EthereumGenesis == MumbaiGenesis() {
+			genesis.Config = params.MumbaiChainConfig
+			if config.EthereumNetworkID == 1 {
+				config.EthereumNetworkID = 80001
+			}
+		}
```
jdkanani, line 195: Can you explain more why this condition is required?

jdkanani (changes requested):

mobile/geth.go
0xKrishna: As per the code and comments in above file, I can see that this comes into the effect when the
```
	if config.EthereumGenesis != "" {
```
and it parses the user-supplied genesis spec, Based on that it check what network hash it matches and update the config accordingly.

cmd/geth/consolecmd.go
```diff
@@ -134,6 +134,9 @@ func remoteConsole(ctx *cli.Context) error {
 				path = filepath.Join(path, "rinkeby")
 			} else if ctx.GlobalBool(utils.GoerliFlag.Name) {
 				path = filepath.Join(path, "goerli")
+			} else if ctx.GlobalBool(utils.MumbaiFlag.Name) || ctx.GlobalBool(utils.BorMainnetFlag.Name) {
```
nodakai, line 137: How about using different paths for Bor Mainnet vs Mumbai? Just like Go-ethereum does?

0xKrishna: I agree, But currently, in our default node setup, we use the same path for bor irrespective of the network, So to avoid that confusion I did the same.

jdkanani (approved):

PR #194, opened 2021-09-23
changed github.com imports to use maticnetwork/bor instead of ethereu…

ssandeep:


ssandeep:
Since we do not do `go get github.com/maticnetwork/bor` and considering that there will be too many merge conflicts every time we need to get changes from upstream geth, we will be sticking to geth imports.

PR #196, opened 2021-09-24
Sealing module

ferranbt:
This PR introduces the sealer module to replace the mining module as the main core logic that seals new blocks, it is described in detail in RFC-11.

It adds the next changes:
- Add a sealer module.
- Eth uses the new sealer module BOTH as the miner and consensus engine.
- The sealer module is specific for PoA/S consensus.


temaniarpit27:
Closing this for now. Will reuse when we will pick up the task

PR #198, opened 2021-09-30
Implemented fetchMessage and removed ethApi from bor consensus

0xKrishna:
The base for this PR was taken from `krishna/network-flags`.
We will merge this into master once that is merged to avoid conflicts.

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -2,7 +2,6 @@ package bor
 
 import (
 	"bytes"
-	"context"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -23,7 +22,6 @@ import (
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/consensus/misc"
 	"github.com/ethereum/go-ethereum/core"
@@ -33,7 +31,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/event"
-	"github.com/ethereum/go-ethereum/internal/ethapi"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rlp"
@@ -210,6 +207,7 @@ type Bor struct {
 	chainConfig *params.ChainConfig // Chain config
 	config      *params.BorConfig   // Consensus engine configuration parameters for bor consensus
 	db          ethdb.Database      // Database to store and retrieve snapshot checkpoints
+	bc          *core.BlockChain    // Blockchain to fetch state and chain config
 
 	recents    *lru.ARCCache // Snapshots for recent block to speed up reorgs
 	signatures *lru.ARCCache // Signatures of recent blocks to speed up mining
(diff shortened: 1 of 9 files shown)


consensus/bor/bor.go
```diff
@@ -455,14 +449,24 @@ func (c *Bor) snapshot(chain consensus.ChainHeaderReader, number uint64, hash co
 				// get checkpoint data
 				hash := checkpoint.Hash()
 
+				state, err := chain.StateAt(hash)
```
0xKrishna, line 452: Getting state `nil` here in case of 0 block number.
@jdkanani @ferranbt Please review.

eth/backend.go
```diff
@@ -209,11 +193,9 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	if err != nil {
 		return nil, err
 	}
-	eth.engine.VerifyHeader(eth.blockchain, eth.blockchain.CurrentHeader(), true) // TODO think on it
 
-	// BOR changes
-	eth.APIBackend.gpo.ProcessCache()
-	// BOR changes
+	// NOT IN GETH
+	eth.engine.VerifyHeader(eth.blockchain, eth.blockchain.CurrentHeader(), true) // TODO think on it
```
jdkanani, line 198: @ferranbt Can you check if we need this anymore?

ferranbt: Not sure what this is doing here, could it be used to preload the snapshots in Bor?

ferranbt:
I think this got more complex than I expected. There are many nested functions one inside another and variables are defined over the place. I think this is fault of the functional approach of the miner/consensus module. I would propose something different related to the work I am doing in #196.

At the end of the backend definition we could do something like [this](https://github.com/maticnetwork/bor/pull/196/files#diff-9cc7046826ce92d9cc5432cf41ed4fc14bb6907da44f771e3932302c40e8dbc2R260) for the consensus:
```
if c, ok := eth.engine.(*consensus.Bor); ok {
   c.Setup(eth.blockchain)
}
```
and then inside Bor:
```
type Bor struct {
   ...
   blockchain *core.Blockchain
   ...
}

func (b *Bor) Setup(blockchain *core.Blockchain) {
   b.blockchain = blockchain
}
```
Now, we do not need to use any of the chain readers, fetchMessage could be something like this:
```
func (b *Bor) fetchMessage(msg callmsg, header *types.Header) ([]byte, error) {
   state, err := b.blockchain.StateAt(header.StateRoot)
   if err != nil {
     return nil, err
   }
   blockContext := core.NewEVMBlockContext(header, chainContext, &header.Coinbase)
   vmenv := vm.NewEVM(blockContext, vm.TxContext{}, state, b.chainConfig, vm.Config{})
   res, _, err := vmenv.StaticCall(
	vm.AccountRef(msg.From()),
	*msg.To(),
	msg.Data(),
	msg.Gas(),
  )
  return res, err
}
```
and the rest of the functions become even simpler. We do not need to pass along all those awful chain readers anymore.

eth/backend.go
jdkanani: Can you remove this line and uncomment https://github.com/maticnetwork/bor/blob/krishna/updated-contract-call-flow/core/blockchain.go#L350-L353 ? I have commented it and added this line hear since `eth.engine` was null before.

0xKrishna:
Cool, I will work on this approach and update the PR.

eth/backend.go
0xKrishna: Got it

consensus/bor/bor.go
```diff
@@ -1256,6 +1230,27 @@ func applyMessage(
 	return nil
 }
 
+// fetch message
+func (c *Bor) fetchMessage(msg callmsg, header *types.Header) ([]byte, error) {
+	state, err := c.bc.StateAt(header.Root)
```
jdkanani, line 1235: What happens when the state is null. We had this issue and that's why we were using `header.parentHash` for `GetCurrentSpan` before. 

@ferranbt can you look into this once?

jdkanani (changes requested):

PR #215, opened 2021-10-25
Remove unsafe eth endpoints

ferranbt:
Some Ethereum endpoints in the eth namespace like eth_sendTransaction use the local unlocked accounts to sign transactions and perform operations. This is unsafe for validator nodes that by default have to unlock their node. A misconfiguration might expose the node to many attack vectors.

This PR directly removes those endpoints since they are not used anymore in the community. These endpoints were useful back when the node was both the client and the wallet manager.


ferranbt:
Closed in favor of #246

PR #223, opened 2021-11-02
Vendor protobuf and go extension issue PR

rajatlko13:
@ferranbt please checkout this


scripts/tools-protobuf.sh
```diff
@@ -0,0 +1,7 @@
+curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v3.12.0/protoc-3.12.0-linux-x86_64.zip
+unzip protoc-3.12.0-linux-x86_64.zip -d $HOME/.local
+export PATH="$PATH:$HOME/.local/bin"
+
+go get google.golang.org/grpc
```
ferranbt, line 5: I do not think we need to vendor grpc at all, that is already done in go modules.

scripts/tools-protobuf.sh
```diff
@@ -0,0 +1,7 @@
...
+
+sudo apt-get install -y golang-protobuf-extensions-dev
```
ferranbt, line 7: What is this for?

ferranbt (changes requested):

scripts/tools-protobuf.sh
rajatlko13: You asked me to add golang protobuf extension, I thought this is the one we needed.

rajatlko13: github.com/protocolbuffers/protobuf-go

ferranbt: Not that one either, please try this:
```

File: scripts/tools-protobuf.sh [hunk 3]
```diff
@@ -0,0 +1,7 @@
+curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v3.12.0/protoc-3.12.0-linux-x86_64.zip
+unzip protoc-3.12.0-linux-x86_64.zip -d $HOME/.local
+export PATH="$PATH:$HOME/.local/bin"
```
ferranbt, line 3: Do not add extra things on PATH, add it to /usr/local/bin or something like that

scripts/tools-protobuf.sh
ferranbt: Also please try to run ```make protoc``` after this to see if it works or not.

ferranbt:
So far looks good but there is something wrong if the protobuf files were regenerated completely. Please run the next commands to see which versions have been installed.
```
$ protoc-gen-go-grpc --version
$ protoc-gen-go --version
$ protoc --version
```

victorcoder (dismissed):
LGTM

ferranbt (approved):

PR #224, opened 2021-11-02
Feature grpc status

ferranbt:
This PR adds a new GRPC endpoint and CLI command to return the status of the client.


docs/cli/chain.md
```diff
@@ -0,0 +1,6 @@
+
+# Chain
+
+The ```chain``` command groups actions to interact with the blockchain in the client:
+
+- [```chain sethead```](./account_new.md): Set the current chain to a certain block.
```
victorcoder, line 6: Is this related to this PR?

victorcoder (commented):
What's the goal of knowing the current block and header? Is this info already reported using other methods?

docs/cli/chain.md
ferranbt: Nope, that is from another one I forgot to do before. Actually, the link of that command is incorrect. I will change it.

ferranbt:
> Should we use the same keys as eth.syncing uses?

I think we have much more freedom here since it is our internal API so we can add more info about the blocks. This is just an initial PR, later on we can add more data (i.e. syncing) to this status API.

ssandeep (dismissed):

vcastellm (approved):
LGTM

PR #226, opened 2021-11-02
Add chain set head command

ferranbt:
This PR adds command + GRPC method to set the head of the chain.


ssandeep (approved):

ferranbt:
I think change the flags/interface of the cli as we find it useful from user feedback.

PR #231, opened 2021-11-11
Added Prometheus flag on bor-cli

Shivam691999:
Added Prometheus Flag on the bor-cli with params { address string , enabled bool} and the workflow to enable the Prometheus server.


command/server/server.go
```diff
@@ -171,6 +181,26 @@ func (s *Server) setupGRPCServer(addr string) error {
...
 }
 
+func (s *Server) StartPrometheus(address string, enabled bool) error {
+	// Hook go-metrics into expvar on any /debug/metrics request, load all vars
+	// from the registry into expvar, and execute regular expvar handler.
+	if enabled {
+		exp.Exp(metrics.DefaultRegistry)
+	}
+
+	http.Handle("/memsize/", http.StripPrefix("/memsize", &Memsize))
```
ferranbt, line 191: Why are you exporting also memsize? I do not think it is part of Prometheus.

command/server/server.go
```diff
@@ -171,6 +181,26 @@ func (s *Server) setupGRPCServer(addr string) error {
...
+
+	log.Info("Starting pprof server", "addr", fmt.Sprintf("http://%s/debug/pprof", address))
```
ferranbt, line 193: You do not need to export pprof metrics either.

command/server/config.go
```diff
@@ -88,6 +88,9 @@ type Config struct {
 
 	// GRPC has the grpc server related settings
 	GRPC *GRPCConfig
+
+	// Prometheus Server
+	Prometheus *PrometheusConfig `hcl:"prometheus,optional"`
```
ferranbt, line 93: Add this config inside the 'Telemetry' group in Config instead of creating a new one.

command/server/config.go
```diff
@@ -236,6 +239,14 @@ type GRPCConfig struct {
 	Addr string
 }
 
+type PrometheusConfig struct {
+	// Enabled selects whether the api is enabled
+	Enabled bool `hcl:"enabled,optional"`
+
+	// Host is the address to bind the api
+	address string `hcl:"host,optional"`
```
ferranbt, line 247: Several things:
- For simplicity, use the same name for the var as the HCL/Json name in the tag. So this one should be called host.
- The variable should be in upper case like "Enabled". Otherwise, it is not public and it will not be set during config decoding.
- Lets use only one variable PrometheusAddr. If it is different from zero it is enabled.

command/server/server.go
```diff
@@ -171,6 +181,26 @@ func (s *Server) setupGRPCServer(addr string) error {
 	return nil
 }
 
+func (s *Server) StartPrometheus(address string, enabled bool) error {
+	// Hook go-metrics into expvar on any /debug/metrics request, load all vars
+	// from the registry into expvar, and execute regular expvar handler.
+	if enabled {
+		exp.Exp(metrics.DefaultRegistry)
```
ferranbt, line 188: Instead of calling this function from the metrics package and since the function is so little (3 lines). Just copy the content of the function here.

command/server/server.go
```diff
@@ -45,6 +50,11 @@ func NewServer(config *Config) (*Server, error) {
 		return nil, err
 	}
 
+	// start the Prometheus Server
+	if err := srv.StartPrometheus(config.Prometheus.address, config.Prometheus.Enabled); err != nil {
```
ferranbt, line 54: Do not use a different function here, we already have one called setupTelemetry to perform telemetry setup.

command/server/server.go
```diff
@@ -171,6 +181,26 @@ func (s *Server) setupGRPCServer(addr string) error {
...
+	if enabled {
+		exp.Exp(metrics.DefaultRegistry)
+	}
+
+	http.Handle("/memsize/", http.StripPrefix("/memsize", &Memsize))
+
+	log.Info("Starting pprof server", "addr", fmt.Sprintf("http://%s/debug/pprof", address))
+
+	go func() {
+		if err := http.ListenAndServe(address, nil); err != nil {
```
ferranbt, line 196: Here you are using the embed HTTP server in Go. Please create a new instance of an HTTP server and use that one instead. Example https://stackoverflow.com/questions/25583028/how-to-write-a-simple-custom-http-server-in-go.

command/server/config.go
ferranbt: Also, please add a flag to set this field as well.

ferranbt (changes requested):

command/server/server.go
```diff
@@ -149,6 +155,42 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
 	// Start system runtime metrics collection
 	go metrics.CollectProcessMetrics(3 * time.Second)
 
+	// Hook go-metrics into expvar on any /debug/metrics request, load all vars
+	// from the registry into expvar, and execute regular expvar handler.
+
+	if len(config.PrometheusAddr) != 0 {
```
ferranbt, line 161: For strings you can do:
```
if config.PrometheusAddr != ""
```

command/server/server.go
```diff
@@ -149,6 +155,42 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
...
+
+		prometheusMux := http.NewServeMux()
+
+		// this would cause a panic:
+		// panic: http: multiple registrations for /debug/vars
+		// http.HandleFunc("/debug/vars", e.expHandler)
+		// haven't found an elegant way, so just use a different endpoint
+
+		prometheusMux.HandleFunc("/debug/metrics", func(w http.ResponseWriter, r *http.Request) {
```
ferranbt, line 170: We do not need this metrics here, please remove.

command/server/server.go
```diff
@@ -149,6 +155,42 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
...
+			exp.ExpHandler(metrics.DefaultRegistry)
+		})
+
+		prometheusMux.HandleFunc("/debug/metrics/prometheus", func(w http.ResponseWriter, r *http.Request) {
```
ferranbt, line 174: Call the endpoint only "/metrics" since that is the Prometheus standard.

command/server/server.go
```diff
@@ -149,6 +155,42 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
...
+			prometheus.Handler(metrics.DefaultRegistry)
+		})
+
+		promServer := &http.Server{
+			Addr:           config.PrometheusAddr,
+			Handler:        prometheusMux,
+			ReadTimeout:    10 * time.Second,
```
ferranbt, line 181: You can remove ReadTimeout, WriteTimeout and MaxHeaderBytes since I think they are populated by default by Go (please double check).

ferranbt (changes requested):
Looking good, just some minor stuff.

command/server/server.go
```diff
@@ -18,12 +19,16 @@ import (
...
 	"github.com/ethereum/go-ethereum/metrics/influxdb"
+	"github.com/ethereum/go-ethereum/metrics/prometheus"
 	"github.com/ethereum/go-ethereum/node"
+	"github.com/fjl/memsize/memsizeui"
 	"github.com/mattn/go-colorable"
 	"github.com/mattn/go-isatty"
 	"google.golang.org/grpc"
 )
 
+var Memsize memsizeui.Handler
```
ferranbt, line 30: Remove

command/server/server.go
```diff
@@ -149,6 +154,35 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
 	// Start system runtime metrics collection
 	go metrics.CollectProcessMetrics(3 * time.Second)
 
+	// Hook go-metrics into expvar on any /debug/metrics request, load all vars
```
ferranbt, line 157: Can you remove this comment?

command/server/server.go
```diff
@@ -149,6 +154,35 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
...
+	// from the registry into expvar, and execute regular expvar handler.
+
+	if config.PrometheusAddr != "" {
+
+		prometheusMux := http.NewServeMux()
+
+		// this would cause a panic:
```
ferranbt, line 164: Could you remove this comment? It does not make sense here.

ferranbt (changes requested):

command/server/config.go
```diff
@@ -275,6 +275,9 @@ type TelemetryConfig struct {
 
 	// InfluxDB has the influxdb related settings
 	InfluxDB *InfluxDBConfig `hcl:"influx,block"`
+
+	// Prometheus Address
+	PrometheusAddr string `hcl:"prometheus-addr,optional"`
```
ferranbt, line 280: Still missing two things here:
- Add the default value for Prometheus in the default config (default="")
- Add the `prometheus-addr` flag in the cli.

ferranbt (changes requested):

ferranbt (approved):
LGTM!

PR #235, opened 2021-11-16
added opentelemetry config, flags, and setup

manav2401:
The tracer is being set in the server instance and hence have created a `shutdownTracer` function, which handles the shutdown case.


go.mod
```diff
@@ -70,14 +70,17 @@ require (
...
+	go.opentelemetry.io/otel/sdk v1.2.0
 	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
 	golang.org/x/net v0.0.0-20210805182204-aaa1db679c0d // indirect
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/sys v0.0.0-20210816183151-1e6c022a8912
 	golang.org/x/text v0.3.6
 	golang.org/x/time v0.0.0-20210220033141-f8bda1e9f3ba
-	google.golang.org/grpc v1.37.0
-	google.golang.org/protobuf v1.26.0-rc.1
+	google.golang.org/grpc v1.42.0
```
ferranbt, line 82: Did you run ```go mod tidy```? I do not like when protobuf and grpc update. Could this be a forced update from openTelemetry? @victorcoder

ferranbt (changes requested):

go.mod
manav2401: Yes. But, they got forced updated on importing the `go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc` package.

command/server/server.go
```diff
@@ -174,6 +182,45 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
...
+	)
+	if err != nil {
+		log.Error("Failed to create open telemetry resource", "service", "bor")
+	}
+
+	// Set up a trace exporter
+	traceExporter, err := otlptracegrpc.New(ctx,
+		otlptracegrpc.WithInsecure(),
+		otlptracegrpc.WithEndpoint("localhost:"+strconv.FormatUint(config.OpenCollectorPort, 10)),
+		otlptracegrpc.WithDialOption(grpc.WithBlock()),
```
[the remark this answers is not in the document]
manav2401: I am guessing I'll have to remove the `otlptracegrpc.WithDialOption(grpc.WithBlock())` line and not the whole initialisation as shown in the preview, right?

command/server/server.go
```diff
@@ -174,6 +182,45 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
...
+		),
+	)
+	if err != nil {
+		log.Error("Failed to create open telemetry resource", "service", "bor")
+	}
+
+	// Set up a trace exporter
+	traceExporter, err := otlptracegrpc.New(ctx,
+		otlptracegrpc.WithInsecure(),
+		otlptracegrpc.WithEndpoint("localhost:"+strconv.FormatUint(config.OpenCollectorPort, 10)),
```
victorcoder, line 201: I think we'll need to setup the OpenCollector address too, this could be just one param with the addr+port, as the collector can be also installed in a remote host.

manav2401: Yes, makes sense. So, should I change it to OpenTelemetry URL and use that instead of just port?

command/server/server.go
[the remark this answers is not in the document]
ferranbt: Correct

command/server/server.go
ferranbt: I would call it the OpenCollectorEndpoint.

command/server/server.go
```diff
@@ -174,6 +188,47 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
...
+		ctx := context.Background()
+
+		res, err := resource.New(ctx,
+			resource.WithAttributes(
+				// the service name used to display traces in backends
+				semconv.ServiceNameKey.String("bor"), // TODO: use the service name from the config
+			),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to create open telemetry resource for service " + "bor")
```
ferranbt, line 202: Could you write this instead:
```
return fmt.Errorf("failed to create open telemetry resource for service: %v", err)
```
We do not need to print the service name since the user should know already to which instance these logs belong, but, we do need to print the specific error.

command/server/server.go
```diff
@@ -174,6 +188,47 @@ func (s *Server) setupMetrics(config *TelemetryConfig) error {
...
+		}
+
+		// Set up a trace exporter
+		traceExporter, err := otlptracegrpc.New(
+			ctx,
+			otlptracegrpc.WithInsecure(),
+			otlptracegrpc.WithEndpoint(config.OpenCollectorEndpoint),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to create open telemetry tracer exporter for service " + "bor")
```
ferranbt, line 212: Same as the other one.

ferranbt (changes requested):
Some minor stuff left.

Shivam691999 (approved):
Looks good to me !

ferranbt (approved):
LGTM!

ferranbt:
Congrats on your first PR to Bor!

command/server/server.go
victorcoder: Why not maintain consistency with the existing vars? `PrometheusAddr` https://github.com/maticnetwork/bor/pull/235/files#diff-0b1642f753dfe53bd60b0d3f731e4e7728f66e550b8820013a46e48db9c317b5R280

victorcoder: But it's consistent with `WithEndpoint` so LGTM

victorcoder (commented):
Good job here!

PR #236, opened 2021-11-17
Added Reorg events to Ethstats.

Shivam691999:


ferranbt:
This PR shows more or less the path forward to track reorgs but we will leave it as a draft until we have the backend and see if we can figure out on the meantime how to make this change less invasive in the Geth codebase.

temaniarpit27:
@0xsharma Please provide an update on this

0xsharma:
@temaniarpit27 This was a PoC for our ethstats approach. We can close it as we have already implemented this in a better way.

PR #237, opened 2021-11-17
Adapt Dockerfile for development

victorcoder:
The Dockerfile for release using Alpine is in the included Dockerfile.release and this one was not used for release so doesn't make sense to have it using multistage nor Alpine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,17 @@
-# Build Geth in a stock Go builder container
-FROM golang:1.17-alpine as builder
+FROM golang:latest
 
-RUN set -x \
-	&& buildDeps='bash build-base musl-dev linux-headers git' \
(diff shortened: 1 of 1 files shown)


ferranbt:
LGTM!

Dockerfile
```diff
@@ -1,29 +1,17 @@
-# Build Geth in a stock Go builder container
-FROM golang:1.17-alpine as builder
+FROM golang:latest
```
ferranbt, line 1: Shouldn't we bind to a specific Golang version.

victorcoder: This is for development, I think that we can stick to latest so we can catch Go upgrade issues maybe from dev or CI

ferranbt (approved):

sreevalsanmr (approved):

PR #238, opened 2021-11-17
added service name in config and flags

manav2401:
- Adds service name in config and flags.
- Sets `os.Hostname()` to default
- Used in OpenCollector Config
- [Linear](https://linear.app/matic/issue/POS-68/add-a-new-name-field-in-bor)

### DIFF
--- a/command/server/config.go
+++ b/command/server/config.go
@@ -38,6 +38,9 @@ type Config struct {
 	// Chain is the chain to sync with
 	Chain string `hcl:"chain,optional"`
 
+	// Name, or identity of the node
+	Name string `hcl:"name,optional"`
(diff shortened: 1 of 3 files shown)


command/server/config.go
```diff
@@ -365,6 +368,7 @@ type AccountsConfig struct {
 func DefaultConfig() *Config {
 	return &Config{
 		Chain:     "mainnet",
+		Name:      Hostname(),
```
ferranbt, line 371: Smart.

ferranbt (approved):
LGTM!

vcastellm (commented):
Late LGTM!

PR #239, opened 2021-11-18
Ferranbt/london fork governance burn

ferranbt:
This PR moves adding the governance burn tokens outside of the TransitionDB function. Everything inside the state_transition.go file should be kept as close as possible to upstream Geth since that is the core consensus logic to process transactions. As much as possible we should move logic to some post hook operations. Hopefully, in this case it was possible.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -103,6 +103,12 @@ func applyTransaction(msg types.Message, config *params.ChainConfig, bc ChainCon
 		return nil, err
 	}
 
+	if result.BurnedAmount != nil {
+		// London fork enabled, send the burned amount to the governance contract that will burn the tokens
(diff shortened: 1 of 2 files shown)


jdkanani:
These changes look more complex and error-prone than it is before. Also, it doesn't change much since we need to change state transition in any case. Having things in one place will make it easier for upstream changes as well. 

@ferranbt would be great if you could share more rational behind the change.

ferranbt:
Everything that runs inside this function:
```
TransitionDb() (*ExecutionResult, error)
...
```
is what we call EVM compatible. That is the scope of the Ethereum state tests. If we change anything inside that function we might take the risk of not being compatible with transactions running fine in Ethereum mainnet (not saying that this change has that risk).

There are other consensus logic outside this function that are chain specific, for example, mining block rewards. As much as possible, we should strive towards keeping things that are specific to our chain outside that transition function.

PR #242, opened 2021-11-22
Geth v1.10.10

ssandeep:
Pulled changes from upstream geth v1.10.10


core/rawdb/accessors_chain.go
```diff
@@ -656,43 +769,59 @@ func WriteBlock(db ethdb.KeyValueWriter, block *types.Block) {
...
 	}
-	tdBlob, err := rlp.EncodeToBytes(td)
-	if err != nil {
-		log.Crit("Failed to RLP encode block total difficulty", "err", err)
+	if err := op.Append(freezerReceiptTable, num, receipts); err != nil {
+		return fmt.Errorf("can't append block %d receipts: %v", num, err)
 	}
-
-	borReceiptBlob := make([]byte, 0)
-	if borReceipt != nil {
```
ferranbt, line 683: Are we doing this safe check now? If there any chance that it is empty? what do we do then?

jdkanani: It's array now. So it will be nil in the worst case and now no need to Encode, so it's safe.

ferranbt (approved):

PR #245, opened 2021-11-23
eth/downloader: open telemetry tracing

manav2401:
NOTE: This is still WIP
This PR adds open tracing for some of the major processes involved in the synchronisation phase (in eth/downloader) of bor.
This is the first iteration for traces added in the module. As per @ferranbt's comment, I'll try to reduce certain sub-tracers added in the code. The initial commit is just for the reference.


manav2401:
I've removed few sub-traces and only kept the main ones. Do let me know your thoughts.

PR #246, opened 2021-11-29
Deprecate personal/wallet endpoints in bor

manav2401:
RFC: 21

This PR removes the usage of personal/wallet related RPC endpoints. This is achieved by removing the backend configuration using key store for the default account manager. The endpoints hence act as dummy with zero usability.
Moreover, it adds configuration of a local scoped account manager for signing/sealing the blocks while mining. This makes sure that the wallet is used only and only for signing/sealing the blocks and not anything else.

The signing/sealing changes are yet to be tested. Do not merge.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -30,7 +30,6 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/consensus"
-	"github.com/ethereum/go-ethereum/consensus/bor"
 	"github.com/ethereum/go-ethereum/consensus/clique"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/bloombits"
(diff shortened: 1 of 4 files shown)


manav2401:
This introduces a new cli flag named `disable-bor-wallet` which is false by default. It allows the node operator to disable the wallet endpoints (ones with prefix personal).

command/server/config.go
```diff
@@ -482,6 +485,7 @@ func DefaultConfig() *Config {
 			PasswordFile:        "",
 			AllowInsecureUnlock: false,
 			UseLightweightKDF:   false,
+			DisableBorWallet:    false,
```
vcastellm, line 488: Should this be `true`?

command/server/server.go
```diff
@@ -98,15 +100,56 @@ func NewServer(config *Config) (*Server, error) {
 	}
 
 	// setup account manager (only keystore)
-	{
-		keydir := stack.KeyStoreDir()
-		n, p := keystore.StandardScryptN, keystore.StandardScryptP
-		if config.Accounts.UseLightweightKDF {
-			n, p = keystore.LightScryptN, keystore.LightScryptP
+	// Not used for RPC endpoints as personal/wallet endpoints will be deprecated
+	/*
```
vcastellm, line 104: Remove commented code?

command/server/config.go
manav2401: I am not really sure of this. I think keeping `true` would be an explicit behaviour. We should instead keep default as usual behaviour. @ferranbt what do you think?

command/server/server.go
manav2401: Sure.

manav2401:
Closing this in favour of https://github.com/maticnetwork/bor/pull/380

PR #248, opened 2021-12-01
Created chain2head feed + protobuf

Shivam691999:
Created a protobuf service to send newHeads, Fork and Reorg event. The service can be used with new CLI using the new event feet Chain2Head event feed.

The commands to run bor through new CL:
` go run command/*.go server`

The command to view the chain2head event is :
`go run command/*.go chain watch`


core/blockchain.go
```diff
@@ -175,15 +175,16 @@ type BlockChain struct {
...
-	logsFeed      event.Feed
-	blockProcFeed event.Feed
-	scope         event.SubscriptionScope
-	genesisBlock  *types.Block
+	hc             *HeaderChain
+	rmLogsFeed     event.Feed
+	chainFeed      event.Feed
+	chainSideFeed  event.Feed
+	chainHeadFeed  event.Feed
+	chain2HeadFeed event.Feed
```
ferranbt, line 183: Put this at the end and separated from the rest and add a comment "// Bor fields" so that we can differentiate from upstream.

core/events.go
```diff
@@ -41,3 +41,9 @@ type ChainSideEvent struct {
 }
 
 type ChainHeadEvent struct{ Block *types.Block }
+
+type Chain2HeadEvent struct {
```
ferranbt, line 45: Add this in a different file called events_bor.go

ferranbt (changes requested):

command/server/service.go
```diff
@@ -107,3 +109,37 @@ func (s *Server) ChainSetHead(ctx context.Context, req *proto.ChainSetHeadReques
...
+			Number: block.NumberU64(),
+		}
+		blockStubs = append(blockStubs, blockStub)
+	}
+
+	return blockStubs
+}
+
+func (s *Server) ChainWatch(req *proto.ChainWatchRequest, reply proto.Bor_ChainWatchServer) error {
+	// 1. start the feed to the blcokchain events
```
vcastellm, line 129:
```suggestion
	// 1. start the feed to the bolckchain events
```

command/server/service.go
```diff
@@ -107,3 +109,37 @@ func (s *Server) ChainSetHead(ctx context.Context, req *proto.ChainSetHeadReques
...
+}
+
+func (s *Server) ChainWatch(req *proto.ChainWatchRequest, reply proto.Bor_ChainWatchServer) error {
+	// 1. start the feed to the blockchain events
+	// 2. for each event send a proto.ChainWatchResponse
+
+	chain2HeadChanSize := 10
+
+	chain2HeadCh := make(chan core.Chain2HeadEvent, chain2HeadChanSize)
+	s.headSub = s.backend.APIBackend.SubscribeChain2HeadEvent(chain2HeadCh)
```
ferranbt, line 135: This is not correct, each ```ChainWatch``` call should be independent from each other so the subscription reference should local to the call. Remove the ```headSub``` variable from Server too.

ferranbt: You should also close the subscription. Do:
```
defer headSub.Unsubscribe()
```
to unsubscribe the subscription once the function is over.

command/server/service.go
```diff
@@ -107,3 +109,37 @@ func (s *Server) ChainSetHead(ctx context.Context, req *proto.ChainSetHeadReques
...
+
+	for {
+		msg := <-chain2HeadCh
+
+		reply.Send(&proto.ChainWatchResponse{Type: msg.Type,
```
ferranbt, line 140: Close the connection if the client closes the connection. Otherwise, it might panic something:
```
err := reply.Send()
if err != nil {
   break
}
```

command/chain_watch.go
```diff
@@ -0,0 +1,71 @@
...
+
+	borClt, err := c.BorConn()
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	sub, err := borClt.ChainWatch(context.Background(), &proto.ChainWatchRequest{})
+	if err != nil {
+		panic(err)
```
ferranbt, line 48: Handle panic:
```
c.UI.Error(err.Error())
return 1
```

command/chain_watch.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	for {
+		msg, err := sub.Recv()
+		if err != nil {
+			// if err == EOF if finished on the other side
```
ferranbt, line 54: Something has happened with the connection in the other side, log it and get out of the loop:
```
c.UI.Info(err)
break
```

command/chain_watch.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	sub, err := borClt.ChainWatch(context.Background(), &proto.ChainWatchRequest{})
+	if err != nil {
+		panic(err)
+	}
+
```
ferranbt, line 50: Close the connection if the user does Ctrl-C:
```
signalCh := make(chan os.Signal, 1)
signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)

go func() {
     <-signalCh
     sub.Close()
}()
```
This will close the stream client connection and the for loop will break in ```Recv```.

core/bor_events.go
```diff
@@ -8,3 +8,10 @@ import (
 type StateSyncEvent struct {
 	Data *types.StateSyncData
 }
+
```
ferranbt, line 11: Add add here a specific type for each of the new chain events:
```
var (
    ChainHeadReorgEvent = "reorg"
    ChainHeadCanonicalEvent = "head"
    ChainHeadForkEvent = "fork"
)
```
Use these names instead of using a raw string each time you want to reference each event.

ferranbt (changes requested):

ferranbt (changes requested):

core/blockchain.go
```diff
@@ -1640,10 +1641,21 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
...
+				NewChain: []*types.Block{block},
+			})
+
 			// BOR
 		}
 	} else {
 		bc.chainSideFeed.Send(ChainSideEvent{Block: block})
+
+		bc.chain2HeadFeed.Send(Chain2HeadEvent{
+			Type:     "fork",
```
ferranbt, line 1656: Use ```core.Chain2HeadForkEvent```

core/blockchain.go
```diff
@@ -2262,6 +2279,13 @@ func (bc *BlockChain) reorg(oldBlock, newBlock *types.Block) error {
 	}
 	// Ensure the user sees large reorgs
 	if len(oldChain) > 0 && len(newChain) > 0 {
+
+		bc.chain2HeadFeed.Send(Chain2HeadEvent{
+			Type:     "reorg",
```
ferranbt, line 2284: Use ```core.Chain2HeadReorgEvent```.

core/blockchain.go
```diff
@@ -1737,6 +1749,11 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 	defer func() {
 		if lastCanon != nil && bc.CurrentBlock().Hash() == lastCanon.Hash() {
 			bc.chainHeadFeed.Send(ChainHeadEvent{lastCanon})
+
+			bc.chain2HeadFeed.Send(Chain2HeadEvent{
+				Type:     "head",
```
ferranbt, line 1754: Use ```core.Chain2HeadCanonicalEvent```.

command/server/service.go
```diff
@@ -107,3 +109,41 @@ func (s *Server) ChainSetHead(ctx context.Context, req *proto.ChainSetHeadReques
...
+			Number: block.NumberU64(),
+		}
+		blockStubs = append(blockStubs, blockStub)
+	}
+
+	return blockStubs
+}
+
+func (s *Server) ChainWatch(req *proto.ChainWatchRequest, reply proto.Bor_ChainWatchServer) error {
+	// 1. start the feed to the blockchain events
```
ferranbt, line 129: Remove these comments.

command/chain_watch.go
```diff
@@ -0,0 +1,91 @@
...
+func (c *ChainWatchCommand) Flags() *flagset.Flagset {
+	flags := c.NewFlagSet("chain watch")
+
+	return flags
+}
+
+// Synopsis implements the cli.Command interface
+func (c *ChainWatchCommand) Synopsis() string {
+	return ""
```
ferranbt, line 33: See the other commands and add ```Synopsis``` and ```Help``` text accordingly.

command/chain_watch.go
```diff
@@ -0,0 +1,91 @@
...
+			c.UI.Output("Reorg Detected")
+
+			out := fmt.Sprintf("Added : %v", msg.Newchain)
+			c.UI.Output(out)
+
+			out = fmt.Sprintf("Removed : %v", msg.Oldchain)
+			c.UI.Output(out)
+		}
+
+		// fmt.Println(msg)
```
ferranbt, line 87: Remove comment.

command/chain_watch.go
```diff
@@ -0,0 +1,91 @@
...
+
+	for {
+		msg, err := sub.Recv()
+		if err != nil {
+			// if err == EOF if finished on the other side
+			c.UI.Output(err.Error())
+			break
+		}
+		if msg.Type == core.Chain2HeadCanonicalEvent {
+			out := fmt.Sprintf("Block Added : %v", msg.Newchain)
```
ferranbt, line 72: This does not look nice. It would be better to create a print function that shows the event:
```
func printEvent(event *proto....) string {
   var out string
   if msg.Type == core.Chain2HeadCanonicalEvent {
       out = fmt.Sprintf("Block Added : %v", msg.Newchain)
   } else if ... {

   }
   return out
}
```
and then:
```
msg, err := sub.Recv()
c.UI.Output(printEvent(msg))
```

ferranbt (changes requested):
Please also make sure you include an entry for this command in the ```docs``` https://github.com/maticnetwork/bor/tree/master/docs/cli

command/server/service.go
```diff
@@ -107,3 +109,39 @@ func (s *Server) ChainSetHead(ctx context.Context, req *proto.ChainSetHeadReques
 	s.backend.APIBackend.SetHead(req.Number)
 	return &proto.ChainSetHeadResponse{}, nil
 }
+
+func ConvertBlockToBlockStub(blocks []*types.Block) []*proto.BlockStub {
```
vcastellm, line 113: This needs to be exported?

Shivam691999: No it did not need to be exported. Fixed it !

vcastellm (approved):
LGTM 🎉

Shivam691999:
Moving this PR to https://github.com/maticnetwork/bor/pull/274

PR #259, opened 2021-12-06
Implementation of SHA3-256 FIPS 202 hash precompile and EcRecover Uncompressed public key precompile for ICON's BTP interoperability protocol.

simsonraj:
### Description

- Implementation of SHA3-256 FIPS 202 hash precompile to interoperate with ICON blockchain
- EcRecover Uncompressed public key precompile to interoperate with ICON blockchain

### Rationale

- In order to validate the Merkle tree data from ICON, the SHA3-256 hashing function is required because all of ICON blockchain's hash and transaction hash is created using the SHA3-256 hashing function
- It is needed to validate the signatures of validators in ICON. Currently, ICON uses a similar address building scheme to Ethereum. ICON uses a part of the hash value of the public key. But we are using a different hash function, SHA3-256 FIPS.

### Example

Test cases performed to check the input and output of the functions for the given parameters below

sha3fips
params
data : ‘0x0448250ebe88d77e0a12bcf530fe6a2cf1ac176945638d309b840d631940c93b78c2bd6d16f227a8877e3f1604cd75b9c5a8ab0cac95174a8a0a0f8ea9e4c10bca’
returns : ‘0xc7647f7e251bf1bd70863c8693e93a4e77dd0c9a689073e987d51254317dc704’

ecrecoverPublicKey
params
hash : ‘0xc5d6c454e4d7a8e8a654f5ef96e8efe41d21a65b171b298925414aa3dc061e37’
v : ‘0x00’
r : ‘0x4011de30c04302a2352400df3d1459d6d8799580dceb259f45db1d99243a8d0c’
s : ‘0x64f548b7776cb93e37579b830fc3efce41e12e0958cda9f8c5fcad682c610795’
returns : ‘0x0448250ebe88d77e0a12bcf530fe6a2cf1ac176945638d309b840d631940c93b78c2bd6d16f227a8877e3f1604cd75b9c5a8ab0cac95174a8a0a0f8ea9e4c10bca’

### Packages Modified
Core, params


ferranbt:
Hi.

What is the reason to include this precompiled contracts besides that they are implemented in ICON blockchain?. IMHO, a hard fork change like this requires a more detailed conversation rather than just dropping the PR without any context. I am going to close this PR since as for now we keep compatibility only with EVM Ethereum mainnet chain and EIPs.

mohamedelshami:
Hi @ferranbt the bigger picture here is not limited to ICON blockchain, Polygon Matic is a multi-facet blockchain network with the primary goal to be EVM compatible as well as act layer 2 scaling provider support for blockchain interoperability.

This PR complements the Polygon goal and doesn't contradict the primary focus of being EVM compatible. 

What's the process of submitting enhancements proposal if this what needs to happen first?

PR #260, opened 2021-12-07
london fork

temaniarpit27:


ssandeep (approved):

params/config.go
```diff
@@ -253,7 +253,7 @@ var (
 			ValidatorContract:     "0x0000000000000000000000000000000000001000",
 			StateReceiverContract: "0x0000000000000000000000000000000000001001",
 			BurntContract: map[string]string{
-				"0": "0x0000000000000000000000000000000000000000f",
+				"0": "0x00000000000000000000000000000000000000000",
```
ssandeep, line 256: Why this change?

temaniarpit27: this is some random address which i had put earlier
its for only testcases

ssandeep (approved):

PR #262, opened 2021-12-07
POS-26 Benchmark current txpool

Stefan-Ethernal:
Implemented (whitebox) benchmark tests for these functions:
- `TxPool.AddRemotesSync`
- `TxPool.Content`
- `TxPool.runReorg` (with and without dummy `txpoolResetRequest` request)
- `TxPool.promoteExecutables`
- `TxPool.demoteUnexecutables`
- `TxPool.truncatePending`
- `TxPool.truncateQueued`
- `txPricedList.Reheap`
- `txSortedMap.Forward`

Separate `tx_pool` and `tx_list` tests into two files containing either unit tests or benchmark tests.

Fixed tests in `filter_system_test.go` (added missing parameter to `NewPublicFilterAPI` invocation). 
**_Note:_**  Seems that [this PR](https://github.com/maticnetwork/bor/pull/258) fixed that issue too, but somewhat different, so we would need to be careful during merge.
Fixed some static analysis warnings in `bor_api.go` and `api.go` (error object messages must not begin with upper case).

`txSortedMap` is covered with blackbox unit and benchmark tests.


temaniarpit27:
@Stefan-Ethernal Can you please provide the latest update for this?

Stefan-Ethernal:
> @Stefan-Ethernal Can you please provide the latest update for this?

As we have discussed with @ferranbt, initial aim of that task was to detect bottlenecks in transaction pool and give refactorings proposals which would give us significant performance improvements. 

It turned out that `tx_pool` already performs pretty well. There only could be some slight memory optimizations, for instance if we would merge two data structures in `tx_list.go` into a single sorted slice, but those would be up to 10% most (refer to these two commits for changes: [tx_list refactor part 1](https://github.com/Stefan-Ethernal/bor/commit/2a9164f08092f7b91b7b8494563ab46d388596ef), [tx_list refactor part 2](https://github.com/Stefan-Ethernal/bor/commit/b99186052c052744582238939f2600890441ce8a))
Large bottleneck which was recognized back then is in transactions signing (`signature_cgo.go`), but it is anyway used for testing purposes, since transactions are put into the transaction pool already signed from the outside.

Conclusion back then was that none of proposed changes would give us desired performance boost (think that Ferran said that we are expecting boost in performances of at least 50%). Therefore I think that we have gave up on merging this PR (although there were unit tests for `tx_list`, which would be probably useful to have if we come up with some fresh ideas in future how to refactor `tx_pool`), so I think that we can close this anyway. Can you confirm @ferranbt?

JekaMas:
> Conclusion back then was that none of proposed changes would give us desired performance boost (think that Ferran said that we are expecting boost in performances of at least 50%).

It is possible to optimize the txpool, however it is not about average case optimization, bit rather edge case optimizations(block was discarded, txpool lookups). https://github.com/ledgerwatch/erigon-lib/tree/main/txpool
Performance-wise the rule of thumb in Ethereum is like 1/3 of execution time is about io operations (mostly about latency and read/write amplification factors), 1/3 golang GC, 1/3 EVM.

PR #269, opened 2021-12-14
jaipur fork

temaniarpit27:


params/config.go
```diff
@@ -486,7 +486,8 @@ type BorConfig struct {
 	StateReceiverContract    string                 `json:"stateReceiverContract"`    // State receiver contract
 	OverrideStateSyncRecords map[string]int         `json:"overrideStateSyncRecords"` // override state records count
 	BlockAlloc               map[string]interface{} `json:"blockAlloc"`
-	BurntContract            map[string]string      `json:"burntContract"` // governance contract where the token will be sent to and burnt in london fork
+	BurntContract            map[string]string      `json:"burntContract"`         // governance contract where the token will be sent to and burnt in london fork
+	JaipurBlock              *big.Int               `json:"jaipurBlock,omitempty"` // Jaipur switch block (nil = no fork, 0 = already on jaipur)
```
ferranbt, line 490: Not sure if json can unmarshal a big.Int here.

consensus/bor/bor.go
```diff
@@ -148,15 +148,15 @@ func ecrecover(header *types.Header, sigcache *lru.ARCCache) (common.Address, er
...
 	hasher := sha3.NewLegacyKeccak256()
-	encodeSigHeader(hasher, header)
+	encodeSigHeader(hasher, header, c)
 	hasher.Sum(hash[:0])
 	return hash
 }
 
-func encodeSigHeader(w io.Writer, header *types.Header) {
-	err := rlp.Encode(w, []interface{}{
+func encodeSigHeader(w io.Writer, header *types.Header, c *params.BorConfig) {
```
ferranbt, line 158: Add unit tests for this so that we are sure we are encoding on every scenario.

consensus/bor/bor.go
```diff
@@ -148,15 +148,15 @@ func ecrecover(header *types.Header, sigcache *lru.ARCCache) (common.Address, er
 }
 
 // SealHash returns the hash of a block prior to it being sealed.
-func SealHash(header *types.Header) (hash common.Hash) {
+func SealHash(header *types.Header, c *params.BorConfig) (hash common.Hash) {
```
ferranbt, line 151: Could we do instead:
```
func (c *Bor) SealHash()...
```
and then the function signature would be the same. I think this function is not being used elsewhere.

consensus/bor/bor.go
```diff
@@ -123,7 +123,7 @@ var (
 type SignerFn func(accounts.Account, string, []byte) ([]byte, error)
 
 // ecrecover extracts the Ethereum account address from a signed header.
-func ecrecover(header *types.Header, sigcache *lru.ARCCache) (common.Address, error) {
+func ecrecover(header *types.Header, sigcache *lru.ARCCache, c *params.BorConfig) (common.Address, error) {
```
ferranbt, line 126: Could we do instead:
```
func (b *Bor) ecrecover()...
```

ferranbt (changes requested):

consensus/bor/bor.go
temaniarpit27: no
because snapshot.go doesn't have bor object

consensus/bor/bor.go
temaniarpit27: no
because snapshot.go doesn't have bor object

params/config.go
temaniarpit27: changed

consensus/bor/bor.go
```diff
@@ -172,8 +172,13 @@ func encodeSigHeader(w io.Writer, header *types.Header) {
 		header.Extra[:len(header.Extra)-65], // Yes, this will panic if extra is too short
 		header.MixDigest,
 		header.Nonce,
-	})
-	if err != nil {
+	}
+	if c.JaipurBlock >= int(header.Number.Int64()) {
```
jdkanani, line 176: Shouldn't we use big.Int to compare?

params/config.go
```diff
@@ -486,7 +486,8 @@ type BorConfig struct {
 	StateReceiverContract    string                 `json:"stateReceiverContract"`    // State receiver contract
 	OverrideStateSyncRecords map[string]int         `json:"overrideStateSyncRecords"` // override state records count
 	BlockAlloc               map[string]interface{} `json:"blockAlloc"`
-	BurntContract            map[string]string      `json:"burntContract"` // governance contract where the token will be sent to and burnt in london fork
+	BurntContract            map[string]string      `json:"burntContract"`         // governance contract where the token will be sent to and burnt in london fork
+	JaipurBlock              int                    `json:"jaipurBlock,omitempty"` // Jaipur switch block (nil = no fork, 0 = already on jaipur)
```
jdkanani, line 490: Shouldn't we use it outside bor config? like London fork flag

jdkanani (changes requested):

params/config.go
temaniarpit27: we tried that but there are few places where we wont have whole config
in snapshot.go basically

consensus/bor/bor.go
ferranbt: It does not matter here, Number in header is a big.int just for convenience but the block number will never be higher than int64.

consensus/bor/bor.go
temaniarpit27: added

params/config.go
jdkanani: Can we convert it to `big.Int` similar to other fork block numbers?

tests/bor/bor_test.go
```diff
@@ -502,3 +505,58 @@ func TestEIP1559Transition(t *testing.T) {
...
+		header.Difficulty,
+		header.Number,
+		header.GasLimit,
+		header.GasUsed,
+		header.Time,
+		header.Extra[:len(header.Extra)-65], // Yes, this will panic if extra is too short
+		header.MixDigest,
+		header.Nonce,
+	}
+	if c.JaipurBlock >= int(header.Number.Int64()) {
```
jdkanani, line 554:
```suggestion
	if c.IsJaipur(header.Number) {
	....
	}
```

And adding -- 

```
IsJaipurFork (blockNumber) {
  c.JaipurBlock.Cmp(blockNumber) or something
}
```

jdkanani (changes requested):

tests/bor/bor_test.go
```diff
@@ -502,3 +505,58 @@ func TestEIP1559Transition(t *testing.T) {
...
+
+// SealHash returns the hash of a block prior to it being sealed.
+func testSealHash(header *types.Header, c *params.BorConfig) (hash common.Hash) {
+	hasher := sha3.NewLegacyKeccak256()
+	testEncodeSigHeader(hasher, header, c)
+	hasher.Sum(hash[:0])
+	return hash
+}
+
+func testEncodeSigHeader(w io.Writer, header *types.Header, c *params.BorConfig) {
```
ferranbt, line 536: I would not do test this way because you have just copied the same code from consensus so If that code is wrong this code will be as well.

temaniarpit27: @ferranbt i agree
should i remove it then?

ssandeep (approved):

ferranbt (approved):

PR #270, opened 2021-12-17
Configure bor params in the unit itself

vcastellm:
## Motivation

This PR removes the necessity of having a startup script and an almost empty environment `metadata` file, less moving parts, and less files to watch.

## Implementation

As we are going to embed genesys data into the binary, this is the most similar method to do it before the new CLI release.

This way, each release gets tied to its corresponding genesys file.

This establish `bor` repository as the source of truth for the genesys file.


builder/files/bor.service
```diff
@@ -6,11 +6,39 @@
...
+    --datadir /var/lib/bor/data \
+    --port 30303 \
+    --http --http.addr '0.0.0.0' \
+    --http.vhosts '*' \
+    --http.corsdomain '*' \
+    --http.port 8545 \
+    --ipcpath /var/lib/bor/bor.ipc \
+    --http.api 'eth,net,web3,txpool,bor' \
+    --syncmode 'full' \
+    --networkid '80001' \
```
ssandeep, line 19: This is for mumbai, how do we specify mainnet networkid 137?

builder/files/bor.service
```diff
@@ -6,11 +6,39 @@
 [Service]
   Restart=on-failure
   RestartSec=5s
-  WorkingDirectory=$NODE_DIR
-  EnvironmentFile=/etc/matic/metadata
-  ExecStart=/usr/local/bin/bor $VALIDATOR_ADDRESS
+  ExecStart=/usr/local/bin/bor \
```
ssandeep, line 9: Can we make use of `--bor-mainnet` and `--bor-mumbai` instead?

.goreleaser.yml
```diff
@@ -95,7 +101,9 @@ dockers:
       - linux-amd64
     build_flag_templates:
       - --platform=linux/amd64
-    skip_push: true
+    extra_files:
+      - builder/files/genesis-mainnet-v1.json
+      - builder/files/genesis-testnet-v4.json
```
ssandeep, line 106: If I want to run a devnet, I just need to add a new genesis file here and use that? How do I specify starting bor using the devnet genesis?

builder/files/bor.service
vcastellm: I don't know, do they work correctly?

.goreleaser.yml
vcastellm: The files are just there, you don't need to add devnet one here. The idea is that you mount a volume with your file and use that to init `bor` using you own genesys file.

builder/files/bor.service
vcastellm: Gotcha, probably as you said it's better to use `--bor-mainnet` and `--bor-mumbai`

builder/files/bor.service
ssandeep: Yes, it works perfectly. We have collected all the common params in there. You can compare it with those configs. Also, if there are additional configs or want to override any values, you can specify them after `--bor-mainnet`.

vcastellm: this is great, then I think we can set this to `--bor-mumbai` by default and leave a comment on `--bor-mainnet`

vcastellm: According to https://github.com/maticnetwork/bor/blob/7368ca6888c6c4850dce537d4aebf9229ba1a55b/cmd/geth/config.go#L342

the deafult sync mode is `snap`, should be `full` right?

ssandeep: Yes, it should be `full`.

ssandeep (approved):

builder/files/bor.service
vcastellm: Changed in defaults

ferranbt (approved):

PR #279, opened 2022-01-03
Use stream to send debug files

### LINKED ISSUE #217: Debug traces longer than max Grpc message
# Rationale

Why should this feature exist?

With the ```bor debug``` command we get 'heap', 'cpu' and 'trace' traces from the running client. Both 'cpu' and 'trace' run for n seconds to get the trace, n is an flag in the cli. However, if the trace runs for more than 4 seconds, the output data is too big to be transferred over a single Grpc message.

# Implementation

Do you have ideas regarding the implementation of this feature?

Use something like https://github.com/mitchellh/go-grpc-net-conn to create a stream connection over the Grpc connection to send arbitrary amount of data and pipe that stream into a file.

ferranbt:
Closes #217.
Before, we would send debug trace files as a GRPC message. Now, we use a net.Conn interface on top of the GRPC message to send big files back.


command/server/service.go
```diff
@@ -13,14 +12,17 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/p2p/enode"
+	gproto "github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/empty"
+	grpc_net_conn "github.com/mitchellh/go-grpc-net-conn"
 )
 
-func (s *Server) Pprof(ctx context.Context, req *proto.PprofRequest) (*proto.PprofResponse, error) {
+func (s *Server) Pprof(req *proto.PprofRequest, stream proto.Bor_PprofServer) error {
```
vcastellm, line 20: Why remove context propagation here?

ferranbt: This is the Grpc spec, stream functions do not have context as parameter. I cannot change the signature of the function, it is given by Grpc.

vcastellm (approved):

PR #283, opened 2022-01-04
Do not attempt DNS discovery and change default sync flag

vcastellm:
* Polygon does not use DNS discovery so remove the cargo-culted init call
* Default sync should be `full`

### DIFF
--- a/cmd/geth/config.go
+++ b/cmd/geth/config.go
@@ -347,7 +347,7 @@ func setDefaultMumbaiGethConfig(ctx *cli.Context, config *gethConfig) {
 	config.Node.HTTPPort = 8545
 	config.Node.IPCPath = utils.MakeDataDir(ctx) + "/bor.ipc"
 	config.Node.HTTPModules = []string{"eth", "net", "web3", "txpool", "bor"}
-	config.Eth.SyncMode = downloader.SnapSync
+	config.Eth.SyncMode = downloader.FullSync
 	config.Eth.NetworkId = 80001
 	config.Eth.Miner.GasCeil = 20000000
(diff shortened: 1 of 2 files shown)


cmd/geth/config.go
```diff
@@ -347,7 +347,7 @@ func setDefaultMumbaiGethConfig(ctx *cli.Context, config *gethConfig) {
 	config.Node.HTTPPort = 8545
 	config.Node.IPCPath = utils.MakeDataDir(ctx) + "/bor.ipc"
 	config.Node.HTTPModules = []string{"eth", "net", "web3", "txpool", "bor"}
-	config.Eth.SyncMode = downloader.SnapSync
+	config.Eth.SyncMode = downloader.FullSync
```
0xKrishna, line 350: When this flag was implemented it was FULL by default but then @jdkanani asked me to use SNAP and we changed it for both Mumbai and Mainnet.

0xKrishna: And I think the reason behind this was that Geth is using SNAP as their default sync mode.

vcastellm: @jdkanani I got it correctly we're having issues with `snap` syncs right? Do we need to set this back to `full`?

manav2401: I was last working on the `snap` sync issue and it still remains unresolved for mainnet.

ssandeep (approved):

cmd/geth/config.go
ssandeep: Yes, we should change it to `full` until we get `snap` working on mainnet.

0xKrishna (approved):

PR #284, opened 2022-01-05
Bump maximum Gas Price Oracle gas price to 5000 Gwei

AlexSSD7:
Right now, whenever you do `eth.gasPrice` it will return the gas price estimate that will be capped at 500 Gwei. In reality, the gas price required to broadcast transactions is somewhere around 600-700 Gwei, but the node is returning 500 Gwei just because it has inherited the go-ethereum's gas price limiter.

### DIFF
--- a/eth/gasprice/gasprice.go
+++ b/eth/gasprice/gasprice.go
@@ -35,7 +35,7 @@ import (
 const sampleNumber = 3 // Number of transactions sampled in a block
 
 var (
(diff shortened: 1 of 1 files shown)


ferranbt (approved):

ssandeep (approved):

vcastellm (commented):
Late LGTM

ajb:
Is `eth_gasPrice` actually working for anyone? It always seem to return 30 gwei for me.

PR #290, opened 2022-01-08
Increase txfeecap - 10 ethers

moltam89:
Hi guys,

Recently the gas prices are so high, that it is quite easy to reach the 1 ether/matic limit.

I think "txfeecap" should be increased, I set it to 10.

Relates issue:
https://github.com/maticnetwork/bor/issues/289

Best,
Tamas

### DIFF
--- a/eth/ethconfig/config.go
+++ b/eth/ethconfig/config.go
@@ -93,7 +93,7 @@ var Defaults = Config{
 	RPCGasCap:     50000000,
 	RPCEVMTimeout: 5 * time.Second,
 	GPO:           FullNodeGPO,
-	RPCTxFeeCap:   1, // 1 ether
+	RPCTxFeeCap:   10, // 10 ethers
 }
 
 func init() {


ferranbt:
Hi, you are using the admin endpoints to send Ethereum transactions right? That is why you need to increase this fee value. Is there any reason why you rely on Bor to send transactions instead of signing and building yourself the txn?

moltam89:
Hi @ferranbt ,

Actually I'm using a Pocket Gateway endpoint, and in their [forum](https://forum.pokt.network/t/set-txfeecap-polygon/1520) this repository was  mentioned.

I just searched for the 1 ether value as a text, I thought it might be useful for someone else as well.

Best,
Tamas

ferranbt:
May I know which RPC endpoint are you calling that is giving you back this error message?

moltam89:
I'm using this one provided by Pocket Gateway:

https://poly-mainnet.gateway.pokt.network/v1/lb/<API_KEY>

ferranbt:
Sorry, I did not meant RPC endpoint provider but the actual RPC endpoint you were calling when it failed (i.e. eth_sendTransaction, eth_sendRawTransaction...).

moltam89:
Sorry, but I don't really understand your question.

I use etherjs to send the transaction:
https://docs.ethers.io/v5/api/signer/#Wallet--properties

temaniarpit27:
Please use the client settings instead of changing the default value
Reopen if required

PR #274, opened 2021-12-22
Feature reorg subscription

Shivam691999:
Continued from https://github.com/maticnetwork/bor/pull/248

Created a protobuf service to send newHeads, Fork and Reorg event. The service can be used with new CLI using the new event feet Chain2Head event feed.

The commands to run bor through new CL:
go run command/*.go server

The command to view the chain2head event is :
go run command/*.go chain watch

PR #303, opened 2022-01-19
Merge subs2

Shivam691999:
This is the rebased version of branch feature-subs2. 
For more info visit : #274


Shivam691999:
This branch is responsible to feed data to the Ethstats-Backend ( https://github.com/maticnetwork/ethstats-backend ). This data is then used to serve the Ethstats-Frontend.

ferranbt:
Sorry for this @Shivam691999. You have to rebase again since the layout of the cli was changed.

ferranbt (changes requested):
Please add tests to see if the events really catch what we expect, these are a couple of examples on how it is being done for other events (i.e. chain side event).

https://github.com/maticnetwork/bor/blob/master/core/blockchain_test.go#L1162

Please do the test in a different file called `core/blockchain_bor_test.go` to make it easy to upstream. Ping me if you need any help.

ferranbt (changes requested):

core/blockchain_bor_test.go
```diff
@@ -0,0 +1,144 @@
...
+	//then reorg happens and triggers 1 event, then last head block of Replacement chain triggers 1 event
+	totalEvents := 4
+
+	const timeoutDura = 10 * time.Second
+	timeout := time.NewTimer(timeoutDura)
+done:
+	for {
+		select {
+		case ev := <-chain2HeadCh:
+			if ev.Type == Chain2HeadReorgEvent {
```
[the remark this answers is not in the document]
Shivam691999: This logic would not work as every time readEvent() is called, it doesn't trigger case evnt. 
The current logic is pretty much similar to TestingSideChainEvents.

core/blockchain_bor_test.go
```diff
@@ -0,0 +1,123 @@
...
+			t.Fatalf("failed to create tx: %v", err)
+		}
+		gen.AddTx(tx)
+	})
+
+	if _, err := blockchain.InsertChain(replacementBlocks); err != nil {
+		t.Fatalf("failed to insert chain: %v", err)
+	}
+
+	i := 0
```
ferranbt, line 56: Do you still need this?

core/blockchain_bor_test.go
```diff
@@ -0,0 +1,123 @@
...
+			i++
+			return &ev
+		case <-time.After(2 * time.Second):
+			t.Fatal("timeout")
+		}
+		return nil
+	}
+
+	// head event
+	event1 := readEvent()
```
ferranbt, line 69: I would recommend you yet another iteration for this tests. Instead of having to manually write down every comparison, you can send to readEvent a struct of what you expect to be seen on the event. For example:

```
type eventTest struct {
    Type Chain2HeadType
    Added []common.Hash
    Removed []common.Hash
}

readEvent := func(expect *eventTest) {
    var ev *Chain2HeadEvent
    select {
        case ev = <-chain2HeadCh:
        case <-time.After(...:
            t.Fatal("timeout")
    }
    // compare ev with expect ...
    // compare Type
    // compare Added hashes
    // compare Removed hashes
}

// event 1
readEvent(&eventTest{
    Type: Chain2HeadCanonicalEvent,
    Added: []common.Hash{
        chain[2].Hash(),
    },
})

// event 2..
// event 3..
```

ferranbt (changes requested):

core/blockchain_bor_test.go
ferranbt: Ping again on this. Why do you need the `i` now?

core/blockchain_bor_test.go
```diff
@@ -0,0 +1,130 @@
...
+	i := 0
+
+	readEvent := func(expect *eventTest) {
+		select {
+		case ev := <-chain2HeadCh:
+			i++
+			if ev.Type != expect.Type {
+				t.Fatalf("%d : type mismatch", i)
+			}
+			for j := 0; j < len(ev.OldChain); j++ {
```
ferranbt, line 72: Please also check sizes for the arrays:
```
if len(ev.OldChain) != len(expect.Removed)...

if len(ev.NewChain) != len(expect.Added) ...
```

ferranbt (changes requested):
Looks good, just some minor stuff left

ferranbt (approved):

ferranbt:
Tests not working.

Shivam691999:
> Tests not working.

@ferranbt Tests were working.. There was some problem with Git.

0xKrishna (approved):

PR #304, opened 2022-01-20
fix for TestReorgSideEvent

Shivam691999:
The test _**https://github.com/maticnetwork/bor/blob/f3d6620eccb0c21225312de59f49ea83a66996c3/core/blockchain_test.go#L1162**_ fails.

This PR will fix this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,10 +48,9 @@ ios:
 	@echo "Done building."
 	@echo "Import \"$(GOBIN)/Geth.framework\" to use the library."
 
-test: all
-	# $(GORUN) build/ci.go test
+test:
 	# Skip mobile and cmd tests since they are being deprecated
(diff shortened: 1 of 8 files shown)


ferranbt:
Why was this working in CI then?

Makefile
```diff
@@ -48,10 +48,9 @@ ios:
 	@echo "Done building."
 	@echo "Import \"$(GOBIN)/Geth.framework\" to use the library."
 
-test: all
-	# $(GORUN) build/ci.go test
+test:
```
ferranbt, line 51: Just a note here, `go test` already does the compilation so we do not need to run `all`.

vcastellm (approved):
LGTM

ssandeep (approved):

PR #306, opened 2022-01-21
Add ability to run using chain json file

0xKrishna:
https://linear.app/matic/issue/POS-138/add-ability-to-run-any-genesis-chain-json-file


internal/cli/server/chains/chain.go
```diff
@@ -19,6 +22,29 @@ var chains = map[string]*Chain{
 }
 
 func GetChain(name string) (*Chain, bool) {
-	chain, ok := chains[name]
-	return chain, ok
+	chain, err := ImportFromFile(name)
+	if err != nil {
+		chain, ok := chains[name]
+		return chain, ok
+	}
```
manav2401, line 29: We should reverse the order by first checking `chains[name]` and then from the file because current implementation won't throw any file read/unmarshal error as it will be overridden with the map error.

vcastellm (approved):
LGTM

internal/cli/server/chains/chain.go
ferranbt: I agree with Manav. This is what polygon-sdk does but I think there are better ways to handle this. For example, right now if there is an error on the chain file, there is no output to the user.

I would check first if the input is a file, then try to unmarshal it like this:

```
func GetChain(name string) (*Chain, error) {
    if _, err := os.Stat(name); err == nil {
      // path exists
      // try to unmarshal
    } else if errors.Is(err, os.ErrNotExist) {
      // path does *not* exist
      // try to read from chains
    } else {
        // filesystem error
       return nil, err
    }
}
```

0xKrishna: I thought about it but it was written that way in the linear task and I honestly don't think it will make that much of a difference. But the point you have made is correct, In case of incorrect file format it will not give that error to user.

ferranbt: Yes, I gave you the link without checking it sorry.

0xKrishna: Updated

ferranbt (changes requested):
Changes look good! One last thing, could you add a test for `importChain` so that we are sure that we can decode a chain in JSON format?

0xKrishna:
> Changes look good! One last thing, could you add a test for `importChain` so that we are sure that we can decode a chain in JSON format?

Done, Please let me know if anything is wrong, I haven't written too many test cases lately.

internal/cli/server/chains/chain_test.go
```diff
@@ -0,0 +1,36 @@
+package chains
+
+import (
+	"testing"
+)
+
+func TestImportFromFile(t *testing.T) {
```
ferranbt, line 7: Rename to `TestChain_ImportFromFile`.

ferranbt (changes requested):
Looks good, only minor changes.

internal/cli/server/chains/chain_test.go
0xKrishna: Done

ferranbt (approved):

manav2401 (dismissed):

vcastellm (approved):
LGTM

manav2401 (approved):
LGTM

PR #307, opened 2022-01-21
Rename cli to bor in cmd

manav2401:


ferranbt:
Now that I think about this, `make all` builds and creates a binary of every folder inside `cmd` to `build` so it would create two binaries one for `geth` (old cli) and one for `bor` (new cli). But, if we run `make bor` it creates `geth` binary and copies it to a `bor` binary so this might create problems with the old cli.

manav2401:
Yeah right. We can merge this once we fully stop old cli and release the new one.

vcastellm (approved):

ferranbt:
I am moving this to draft until we are clear to merge.

PR #312, opened 2022-01-24
Add FeeConfigMap to ChainConfig

0xKrishna:


vcastellm (approved):
Lots of number castings but LGTM

0xKrishna:
> Lots of number castings but LGTM

I know, But we are going to change `ElasticityMultiplier` which is an `int` as of now to `float` and that's why all the calculations related to this are affected.

vcastellm (commented):
Remember to modify genesis files https://github.com/maticnetwork/bor/tree/master/builder/files https://github.com/maticnetwork/bor/blob/bae032365db29377e723c6b3d53d6de2e410db36/command/server/chains/

0xKrishna:
> Remember to modify genesis files https://github.com/maticnetwork/bor/tree/master/builder/files https://github.com/maticnetwork/bor/blob/bae032365db29377e723c6b3d53d6de2e410db36/command/server/chains/

Yup, I will once the block number and config settings are finalized.
Code has default check, If config is not added to genesis then it will take default fee config.

temaniarpit27:
@ssandeep can we close this for now?
i don't think its needed now
we can just keep track of branch and use if required in future

temaniarpit27:
@0xKrishna can we close this for now?

PR #318, opened 2022-02-03
Build and run the new CLI

vcastellm:
Build the new CLI in makefile and release pipeline.

### DIFF
--- a/builder/files/bor.service
+++ b/builder/files/bor.service
@@ -7,18 +7,18 @@
   Restart=on-failure
   RestartSec=5s
   ExecStart=/usr/local/bin/bor server \
-    --bor-mumbai \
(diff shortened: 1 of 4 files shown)


builder/files/bor.service
```diff
@@ -6,7 +6,7 @@
 [Service]
   Restart=on-failure
   RestartSec=5s
-  ExecStart=/usr/local/bin/bor \
+  ExecStart=/usr/local/bin/bor server \
     --bor-mumbai \
```
manav2401, line 10: Do we need to replace this with `--chain=mumbai`?

vcastellm: Yeah probably

manav2401 (approved):
Looks good to me

PR #323, opened 2022-02-10
Update default cli config

manav2401:
This PR does following things 
- Updates some default config parameters for the new cli.The parameters updated are 
    - HTTP flag
    - HTTP Modules
    - TxPool.GlobalSlots
    - TxPool.GlobalQueue
- Fixes a bug which arises while setting maxpeers flag to 0.
- Adds a new `no-snapshot` flag, which will be used to disable the snapshot database mode. 
- Fixes a bug while checking snapshot enabling. It was earlier checking in a wrong way as compared to old cli. The old cli condition can be found [here](https://github.com/maticnetwork/bor/blob/master/cmd/utils/flags.go#L1576).


vcastellm (approved):
LGTM

ferranbt:
Is Http enabled by default?

manav2401:
@ferranbt No it wasn't, but last week we enabled it, thinking that we can always disable by passing the `--http false` flag. But, that didn't work, so just reverting that change for now, making it disabled by default. 
Also, I'll get back with the test as this thing of not overriding the defaults might be something to look at.

ferranbt:
So, what is the end goal? Http enabled or not?

ssandeep (approved):

0xsharma (approved):

internal/cli/server/config.go
```diff
@@ -417,9 +421,9 @@ func DefaultConfig() *Config {
 			PriceLimit:   1,
 			PriceBump:    10,
 			AccountSlots: 16,
-			GlobalSlots:  4096,
+			GlobalSlots:  131072,
```
0xsharma, line 424: @manav2401 please confirm these numbers for globalSlots and globalQueue. I guess we decreased them to 32768.

0xsharma (approved):
LGTM !

internal/cli/server/config.go
```diff
@@ -58,9 +58,12 @@ type Config struct {
 	// GcMode selects the garbage collection mode for the trie
 	GcMode string `hcl:"gc-mode,optional"`
 
-	// XXX
+	// Snapshot enables the snapshot database mode
 	Snapshot bool `hcl:"snapshot,optional"`
```
ferranbt, line 62: Why do we have both `Snapshot` and `NoSnapshot` here? Shouldn't we only have one?

internal/cli/server/config.go
```diff
@@ -406,27 +409,28 @@ func DefaultConfig() *Config {
...
 			Journal:      "",
 			Rejournal:    time.Duration(1 * time.Hour),
-			PriceLimit:   1,
+			PriceLimit:   30000000000,
 			PriceBump:    10,
 			AccountSlots: 16,
-			GlobalSlots:  4096,
-			AccountQueue: 64,
-			GlobalQueue:  1024,
+			GlobalSlots:  32768,
```
ferranbt, line 424: When did we decide to change these values? Was this changed before in Bor?

ferranbt (changes requested):

internal/cli/server/config.go
manav2401: These values are currently being used in bor nodes. I think this change was when we faced an issue with the alchemy node with the transactions. We've been using this since then. 

@ssandeep can add more info here!

internal/cli/server/config.go
manav2401: Oh yes. I never thought of this as a redundant flag. I thought this is just an additional flag.

internal/cli/server/config.go
```diff
@@ -810,7 +810,7 @@ func (c *Config) buildEth(stack *node.Node) (*ethconfig.Config, error) {
 	}
 
 	// snapshot disable check
-	if c.Snapshot {
+	if c.NoSnapshot {
```
temaniarpit27, line 813: not sure but how come Snapshot and NoSnapshot checks are same

manav2401: Yes, so there was a bug in the new-cli implementation. It was earlier checking in a wrong way as compared to old cli. The old cli condition can be found [here](https://github.com/maticnetwork/bor/blob/master/cmd/utils/flags.go#L1576). 
If you see the (add: no-snapshot flag, handle snapshot disable](https://github.com/maticnetwork/bor/pull/323/commits/607645f490486fc70cba38ffae48dfd2d84d1adf) commit, it has the bug fix.

temaniarpit27 (approved):

ferranbt (approved):

PR #337, opened 2022-02-26
increase underpricedTxSet memory and increase blockLimit max received per peer

thogard785:
Two slight tweaks that I've been running on my nodes. Should help with the issues that polygon is currently having re; tx spam. The rest of the changes on my nodes are pretty extensive and would require significant review, but these two are pretty clear cut and the results should be pretty straight forward.  I hate github and have no idea how to use it - apologies if I did the PR wrong.


temaniarpit27:
@thogard785 We are closing this as we don't know the consequences for this change. We can revisit this if you want

PR #338, opened 2022-02-26
Add DataDog profiler

vcastellm:
Setup and start the DataDog continuous profiler.

New flags added for profiler config
```
--datadog.profiler
--datadog.profiler.expensive
--datadog.profiler.environment "env"
--datadog.profiler.tags "key:value,key:value"
```


metrics/metrics.go
```diff
@@ -93,6 +94,28 @@ func CollectProcessMetrics(refresh time.Duration) {
 		diskWriteBytes        = GetOrRegisterMeter("system/disk/writedata", DefaultRegistry)
 		diskWriteBytesCounter = GetOrRegisterCounter("system/disk/writebytes", DefaultRegistry)
 	)
+
+	option := profiler.WithProfileTypes(
```
jdkanani, line 98: Can we not enable datadog by default? Many node runners might not use it.

We can add command `--metrics.datadog` and `--metrics.datadog.expensive`

jdkanani (changes requested):

metrics/metrics.go
vcastellm: That was my first intention https://github.com/maticnetwork/bor/pull/338/commits/9a62679609622fca9c4b0236e96cb85d0efc4e26 to put it under the `pprof` config, but it didn't work, going to look to it carefully now, and add the config param.

jdkanani (approved):

internal/debug/flags.go
```diff
@@ -184,6 +208,16 @@ func Setup(ctx *cli.Context) error {
...
 		StartPProf(address, !ctx.GlobalIsSet("metrics.addr"))
 	}
+
+	if ctx.GlobalBool(datadogProfilerFlag.Name) {
+		StartDatadogProfiler(
+			"bor",
+			datadogProfilerEnvironmentFlag.Value,
+			params.VersionWithMeta,
+			datadogProfilerTagsFlag.Value,
+			ctx.GlobalBool(datadogProfilerExpensiveFlag.Name),
```
ssandeep, line 218: What will be the default option in this case?

ssandeep (approved):

ferranbt (commented):
I am not a fan of mixing go-ethereum code with Bor since it complicates upstream changes. Could we at least move the code to another file in `internal/debug`?

vcastellm:
> I am not a fan of mixing go-ethereum code with Bor since it complicates upstream changes. Could we at least move the code to another file in `internal/debug`?

Ok, agree

> 
> Which is the target branch for this change?

the target branch is [gethv1.10.15-merge](https://github.com/maticnetwork/bor/tree/gethv1.10.15-merge) this is the branch for the next minor release `v0.2.15`

internal/debug/profiler.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	datadogProfilerTagsFlag = cli.StringFlag{
+		Name:  "datadog.profiler.tags",
+		Usage: "Comma-separated tags (key:value) attached to all measurements",
+		Value: "",
+	}
+)
+
+func init() {
+	Flags = append(Flags,
```
vcastellm, line 31: I don't like this @ferranbt, can we just add these to the `var Flags` declaration even if it's changing geth upsteam?

ferranbt (approved):

PR #344, opened 2022-03-03
fix: modify the comment of txlookuplimit

shiziwen:
fix: modify the comment of txlookuplimit.

The comment of the txlookuplimit show as "Number of recent blocks to maintain transactions index for (default = about one year, 0 = entire chain)". It is correct for the Eth, but for bor, the default time is about 56 days.


temaniarpit27 (approved):

vcastellm (approved):
LGTM

PR #359, opened 2022-03-16
Handle graceful shutdown in bor

manav2401:
This PR handles graceful shutdown in bor client which is caused due to the rest client implementation.

Why does the error exist in the first place?
This is because the parent function is handled by the worker module in an endless loop and until the child function `FetchWithRetry` return a response, the parent loop will keep waiting even though the exit channel has been closed. 

This is solved by using a `closeCh` channel in the heimdall rest client used in bor consensus. This can be handled by the `Close()` function in consensus interface. This brings one major change in the closing pattern for the client, which is closing consensus engine ahead of the miner.

Linear: [POS-260 Investigate Bor Unclean Shutdown](https://linear.app/matic/issue/POS-260/investigate-bor-unclean-shutdown)


0xsharma (approved):
LGTM 🚀

ssandeep (approved):

consensus/bor/rest.go
```diff
@@ -96,13 +100,23 @@ func (h *HeimdallClient) FetchWithRetry(rawPath string, rawQuery string) (*Respo
...
+		case <-ctx.Done():
+			log.Info("Waiting for graceful shutdown")
+			time.Sleep(5 * time.Second)
+			return nil, fmt.Errorf("shutdown detected")
+		case <-time.After(5 * time.Second):
+			res, err := h.internalFetch(u)
+			if err == nil && res != nil {
+				return res, nil
+			}
+			log.Info("Retrying again in 5 seconds for next Heimdall span", "path", u.Path)
```
jdkanani, line 118:
```suggestion
			log.Info("Retrying again in 5 seconds for next Heimdall data", "path", u.Path)
```

jdkanani (changes requested):

ssandeep (approved):

jdkanani (approved):

consensus/bor/rest.go
```diff
@@ -96,13 +100,23 @@ func (h *HeimdallClient) FetchWithRetry(rawPath string, rawQuery string) (*Respo
 	u.Path = rawPath
 	u.RawQuery = rawQuery
 
+	// prepare a notify context to handle interrups
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
```
ferranbt, line 104: We should not be relying on `os.Interrupt` at the low level of the code flow. Instead, we should use `Close()` functions that propagate when the low-level code has to stop. Note that there is already a `Close()` function as part of the consensus interface, so we have already an entry point to notify `HeimdallClient` that it has to stop.

An extra benefit of this is that you can test the `HeimdallClient` implementation without having to call `os.Interrupt` which I am not sure it is possible with the test framework of Go.

Usually, I approach these problems with a `Close()` function and a `closeCh` channel.

```
type HeimdallClient struct {
   closeCh chan struct{}
}

...
  &HeimdallClient{
    closeCh: make(chan struct{}),
  }
...

func (h *HeimdallClient) Close() {
   close(h.closeCh)
}

func (h *HeimdallClient) FetchWithRetry() {
  ... 
  for {
     select {
        case <-h.closeCh:
            return
        case <-time.After(5*time.Second):
     }
   }
}
```

ferranbt (changes requested):

manav2401:
Have updated the approach with what ferran suggested. One thing to notice here is that we now need to close consensus `engine` first and then the `miner`.

ferranbt (approved):

consensus/bor/rest.go
```diff
@@ -97,12 +100,17 @@ func (h *HeimdallClient) FetchWithRetry(rawPath string, rawQuery string) (*Respo
 	u.RawQuery = rawQuery
 
 	for {
-		res, err := h.internalFetch(u)
-		if err == nil && res != nil {
-			return res, nil
+		select {
+		case <-h.closeCh:
+			log.Info("Waiting for graceful shutdown")
+			return nil, fmt.Errorf("shutdown detected")
```
JekaMas, line 106: It might be better not to use `f` if we don't have formatting. Maybe it's better to create an error variable?

consensus/bor/rest.go
```diff
@@ -97,12 +100,17 @@ func (h *HeimdallClient) FetchWithRetry(rawPath string, rawQuery string) (*Respo
 	u.RawQuery = rawQuery
 
 	for {
-		res, err := h.internalFetch(u)
-		if err == nil && res != nil {
-			return res, nil
+		select {
+		case <-h.closeCh:
```
JekaMas, line 104: The fix looks fine. How do you thing does it worth it to add a test with something like https://github.com/uber-go/goleak ?

JekaMas (commented):
A few comments to discuss

JekaMas (approved):

consensus/bor/rest.go
manav2401: Converting this to a separate task, as per our discussion.

temaniarpit27 (approved):

PR #361, opened 2022-03-22
New CLI console + attach IPC.

0xsharma:
Added bor console method. Also, refactored bor attach command. 

[https://linear.app/matic/issue/POS-271/checkout-ipc-console-and-implement-in-new-cli](url)


internal/cli/attach.go
```diff
@@ -14,63 +15,106 @@ import (
...
 func (c *AttachCommand) Run(args []string) int {
 
-	c.remoteConsole(args)
+	flags := c.Flags()
+
+	//check if first arg is flag or IPC location
+	if len(args) == 0 {
+		args = append(args, "")
+	}
+	if args[0] != "" && string(args[0][0:2]) == "--" {
```
ferranbt, line 75:
```suggestion
	if args[0] != "" && strings.HasPrefix(args, "--") {
```

ferranbt: Why do you use "--"? Can you add a description to the PR on how to use the command?

internal/cli/attach.go
```diff
@@ -14,63 +15,106 @@ import (
...
+		if err := flags.Parse(args); err != nil {
+			c.UI.Error(err.Error())
+			return 1
+		}
+	} else {
+		c.Endpoint = args[0]
```
ferranbt, line 81: So I am not sure if you really need this, is it because "--"?

internal/cli/attach.go
```diff
@@ -106,3 +148,19 @@ func dialRPC(endpoint string) (*rpc.Client, error) {
 	}
 	return rpc.Dial(endpoint)
 }
+
+// MakeConsolePreloads retrieves the absolute paths for the console JavaScript
+// scripts to preload before starting.
+func (c *AttachCommand) MakeConsolePreloads() []string {
```
ferranbt, line 154: This function does not need to be public.

internal/cli/server/command_flags.go
```diff
@@ -0,0 +1,504 @@
+package server
+
+import (
+	"github.com/ethereum/go-ethereum/internal/cli/flagset"
+)
+
+func (c *ConsoleCommand) Flags() *flagset.Flagset {
```
ferranbt, line 7: What is this file? It has so many flags.

internal/cli/server/console.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/ethereum/go-ethereum/cmd/utils"
+	"github.com/ethereum/go-ethereum/console"
+	"github.com/ethereum/go-ethereum/node"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/mitchellh/cli"
+)
+
+// ConsoleCommand is the command to Connect to remote Bor IPC console
+type ConsoleCommand struct {
```
ferranbt, line 17: Why is this command part of `server`?

internal/cli/server/console.go
```diff
@@ -0,0 +1,163 @@
...
+			return 1
+		}
+	}
+	if err := config.Merge(c.cliConfig); err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+	c.config = config
+
+	srv, err := NewServer(config)
```
ferranbt, line 72: Not sure if I understand `console` here. Can you add a description on this?

ferranbt (changes requested):

internal/cli/attach.go
```diff
@@ -14,63 +15,106 @@ import (
...
+			return 1
+		}
+	} else {
+		c.Endpoint = args[0]
+		if err := flags.Parse(args[1:]); err != nil {
+			c.UI.Error(err.Error())
+			return 1
+		}
+	}
+	c.remoteConsole()
```
ferranbt, line 87: Error is not being handled.

internal/cli/attach.go
ferranbt: Can you apply this https://github.com/maticnetwork/bor/pull/361/files#r832158323? Otherwise, it might panic if the size of the args is less than 2.

ferranbt (changes requested):

ferranbt (approved):

temaniarpit27 (approved):

PR #365, opened 2022-03-29
fix api GetTransactionReceiptsByBlock bug

wanyvic:
https://github.com/maticnetwork/bor/pull/164/ when block has bor tx, all `transactionHash` as same


temaniarpit27:
Please provide test cases for the change

PR #376, opened 2022-04-22
Automatically generate markdown pages from bor CLI

cffls:
Adding a script that can automatically generate markdown pages from bor
CLI, so we can avoid copy-pasting helper strings whenever a flag is created,
deleted, or modified.

CLI docs could be generated with command `make docs`.

In this commit, all markdown pages are re-generated with `make docs`. The major difference in the doc is the order of the flags, which are now ordered alphabetically. New cli docs could be previewed here: https://github.com/maticnetwork/bor/tree/jc/clidoc/docs/cli


manav2401:
As we don't support `light` mode in the new cli ([reference](https://github.com/maticnetwork/bor/blob/v0.3.0-dev/internal/cli/server/config.go#L785-L796)), should we remove it from the description ([here](https://github.com/maticnetwork/bor/blob/v0.3.0-dev/internal/cli/server/flags.go#L39))?

manav2401:
Really amazing work. Changes looks good. Could you open this against v0.3.0-dev? Also, you might have to rebase it as I've merged a PR in that branch, which has changes in a couple of flags.

cffls:
Thanks for the feedbacks! I removed the `light` mode and change the base branch to v0.3.0-dev.

manav2401 (approved):

0xsharma (approved):

PR #382, opened 2022-05-05
Fix : Appended Bor Tx in eth_getTransactionByBlockNumberAndIndex

0xsharma:
https://github.com/maticnetwork/bor/issues/369

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1433,6 +1433,7 @@ func newRPCPendingTransaction(tx *types.Transaction, current *types.Header, conf
 // newRPCTransactionFromBlockIndex returns a transaction that will serialize to the RPC representation.
 func newRPCTransactionFromBlockIndex(b *types.Block, index uint64, config *params.ChainConfig, backend Backend) *RPCTransaction {
 	txs := b.Transactions()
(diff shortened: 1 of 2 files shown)


internal/ethapi/api.go
```diff
@@ -1431,16 +1432,26 @@ func newRPCPendingTransaction(tx *types.Transaction, current *types.Header, conf
...
+		return nil
 	}
+
+	tx, _, _, _ := rawdb.ReadBorTransaction(db, borReceipt.TxHash)
+
+	if tx == nil {
+		return nil
+	}
+
+	txs = append(txs, tx)
```
JekaMas, line 1449: Is this slice is sorted after the operation? If not, how could we get by index?

0xsharma: The index of state-sync-transaction is always the last one in the block. That's why I believe it is safe to append it to the end of the list.

JekaMas: Got it. Should work, however could you add a small test to document expected behaviour?

0xsharma:
Fixed it in PR https://github.com/maticnetwork/bor/pull/480.

PR #385, opened 2022-05-06
Fix prometheus path and add prometheus to bor.service

cffls:
Prometheus metrics was not working for three reasons: 

1. path was set to `/metric`
2. we need to use `Handler` instead of `HandleFunc`. Also see the same fix in v3 https://github.com/maticnetwork/v3/pull/52
3. bor.service did not enable metrics and prometheus's address

This commit fixes these problems.


JekaMas (approved):

vcastellm (approved):
LGTM

PR #386, opened 2022-05-08
Rfc35/common ancestor approach units

manav2401:


eth/downloader/whitelist/service_test.go
```diff
@@ -30,3 +33,64 @@ func TestWhitelistCheckpoint(t *testing.T) {
...
+			return []*types.Header{&header}, []common.Hash{hash}, nil
+		case 2:
+			header.Number = big.NewInt(1) // sending wrong header for misamatch
+			return []*types.Header{&header}, []common.Hash{hash}, nil
+		default:
+			return nil, nil, errors.New("invalid number")
+		}
+	}
+
+	// case2: false fetchHeadersByNumber function provided, should consider the chain as valid
```
JekaMas, line 77: Could you make this cases as separate subtests using t.Run and table tests - https://go.dev/blog/subtests ?
Or this cases are part of one bigger single test?

manav2401: Oh I see, wasn't aware of this thing. Yes, my intention was to have these cases as sub-tests. Will make necessary changes.

JekaMas: Great!  Also please check how to use t.Parallel() with subtests.

JekaMas:
@manav2401 Please, check latest updates in the branch. I believe it might be enough for the first version to run it on Polygon testnet. Also please check ErrNoRemoteCheckoint logic, we might need a test for that case.

eth/downloader/whitelist/service.go
```diff
@@ -0,0 +1,116 @@
...
+	lastCheckpointBlockHash := w.checkpointWhitelist[lastCheckpointBlockNum]
+
+	// todo: we can extract this as an interface and mock as well or just test IsValidChain in isolation from downloader passing fake fetchHeadersByNumber functions
+	headers, hashes, err := fetchHeadersByNumber(lastCheckpointBlockNum, 1, 0, false)
+	if err != nil {
+		return false, fmt.Errorf("%w: last checkpoint %d, err %v", ErrNoRemoteCheckoint, lastCheckpointBlockNum, err)
+	}
+
+	if len(headers) == 0 {
+		return true, fmt.Errorf("%w: last checkpoint %d", ErrNoRemoteCheckoint, lastCheckpointBlockNum)
```
manav2401, line 59: Shouldn't we return false in this case?

eth/downloader/whitelist/service.go
```diff
@@ -0,0 +1,116 @@
...
+	if len(w.checkpointWhitelist) == 0 {
+		// worst case, we don't have the checkpoints in memory
+		return true, nil
+	}
+
+	// Fetch the last checkpoint entry
+	lastCheckpointBlockNum := w.checkpointOrder[len(w.checkpointOrder)-1]
+	lastCheckpointBlockHash := w.checkpointWhitelist[lastCheckpointBlockNum]
+
+	// todo: we can extract this as an interface and mock as well or just test IsValidChain in isolation from downloader passing fake fetchHeadersByNumber functions
```
manav2401, line 52: Now that this is handled with a unit test to check is valid chain in isolation, should we remove this?

JekaMas: The comment should be removed.

eth/downloader/whitelist/service.go
JekaMas: It is a kinda valid but empty... But you are right. Let's return false and retry later.

JekaMas (dismissed):
Merge release branch, fix CI and it us ready to be merged

eth/backend.go
```diff
@@ -582,8 +586,70 @@ func (s *Ethereum) Start() error {
...
+	err := s.handleWhitelistCheckpoint()
+	if err != nil {
+		if errors.Is(err, ErrBorConsensusWithoutHeimdall) || errors.Is(err, ErrNotBorConsensus) {
+			return
+		}
+
+		log.Warn("unable to whitelist checkpoint - first run", "err", err)
+	}
+
+	ticker := time.NewTicker(100 * time.Second)
```
JekaMas, line 611: Could you make a task for later to have this wicker duration value in config, rather then hard-coded?

eth/handler_bor.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	// fetch the end checkpoint block hash
+	block, err := h.ethAPI.GetBlockByNumber(context.Background(), rpc.BlockNumber(checkpoint.EndBlock.Uint64()), false)
+	if err != nil {
+		log.Debug("Failed to get end block hash of checkpoint while whitelisting")
+		return 0, common.Hash{}, errEndBlock
+	}
+
+	hash := fmt.Sprintf("%v", block["hash"])
```
JekaMas, line 72: Could we use Hex() method?

JekaMas (dismissed):
LGTM overall.

eth/backend.go
manav2401: Sure, will do

JekaMas: Thank. I believe the PR looks good. Let's move forward with it.

manav2401: Yes, just have one last thing to check for which is the failing test case. Will do it ASAP.

JekaMas (approved):

PR #394, opened 2022-05-11
internal/cli/server, eth: add option to disable personal wallet endpoints

manav2401:
This PR adds a new flag called disable-bor-wallet in the new-cli, which in-turn disables the rpc endpoints with personal namespace. Basically it won't allow the users to perform any wallet related operation. Incase it's a mining node, it will perform the authorisation using a dummy account manager.

Note: The endpoints will still exist and will also show up in the list of commands. It's just that the account manager has no keystore and hence those endpoints will be of no use.

The changes have been tested for developer mode as well as normal mode.

### DIFF
--- a/docs/cli/server.md
+++ b/docs/cli/server.md
@@ -36,6 +36,8 @@ The ```bor server``` command runs the Bor client.
 
 - ```gpo.ignoreprice```: Gas price below which gpo will ignore transactions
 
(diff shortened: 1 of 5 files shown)


cffls (approved):
LGTM

JekaMas (approved):

PR #397, opened 2022-05-12
internal/cli: add block tracing

manav2401:
This PR adds a new command to print a JSON output of the trace of a block (by number, tag, hash or bad block hash).

A helper utility which can create and destroy a mock server is added. This will abstract the port handling (primarily for the grpc server, http server and any other if added in future) from the tests.


eth/tracers/api_bor.go
```diff
@@ -56,7 +57,8 @@ func (api *API) traceBlock2(ctx context.Context, block *types.Block, config *Tra
 	if config != nil && config.Reexec != nil {
 		reexec = *config.Reexec
 	}
-	statedb, err := api.backend.StateAtBlock(ctx, parent, reexec, nil, true)
```
manav2401: @ferranbt would also love your thoughts on this change as well, which is due to the latest geth merge. Basically, if setting `false` to prefer disk flag looks good to go or not?

eth/tracers/api_bor.go
```diff
@@ -56,7 +57,8 @@ func (api *API) traceBlock2(ctx context.Context, block *types.Block, config *Tra
...
+	// TODO: discuss consequences of setting preferDisk false.
```
manav2401, line 60: @ferranbt would also love your thoughts on this change as well, which is due to the latest geth merge. Basically, if setting false to prefer disk flag looks good to go or not?

ferranbt: I think it does not matter since it is only used when provided a pre-state reference and we are not doing it.

internal/cli/server/service.go
```diff
@@ -255,3 +295,30 @@ func (s *Server) ChainWatch(req *proto.ChainWatchRequest, reply proto.Bor_ChainW
 		}
 	}
 }
+func (s *Server) TraceBlock(ctx context.Context, req *proto.TraceRequest) (*proto.TraceResponse, error) {
+	fmt.Println("- trace block -")
```
ferranbt, line 299: Remove print, change it with a `log.Info`

internal/cli/server/service.go
```diff
@@ -255,3 +295,30 @@ func (s *Server) ChainWatch(req *proto.ChainWatchRequest, reply proto.Bor_ChainW
...
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Use streams for this
+	raw, err := json.Marshal(res)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(res)
```
ferranbt, line 319: Remove prints.

ferranbt (approved):

internal/cli/server/proto/server_grpc.pb.go
```diff
@@ -1,4 +1,8 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
```
JekaMas, line 1: this code generation is it in Makefile or has been done by hands?

manav2401: I believe it is through Makefile using the command `make protoc`.

eth/tracers/api_bor.go
```diff
@@ -0,0 +1,143 @@
...
+
+type TraceBlockRequest struct {
+	Number     int64
+	Hash       string
+	IsBadBlock bool
+	Config     *TraceConfig
+}
+
+// If you use context as first parameter this function gets exposed automaticall on rpc endpoint
+func (api *API) TraceBlock2(req *TraceBlockRequest) (*BlockTraceResult, error) {
```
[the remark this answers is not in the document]
manav2401: I believe `TraceBlock` was an old function which ultimately called `TraceBlock2`. Now, we directly call `TraceBlock2` from `DebugBlock` function. We can safely remove the old function and only rely on the new one. And I guess, you're right. Renaming it makes more sense. Will do necessary changes.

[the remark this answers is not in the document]
manav2401: Also, if you see the `TraceBlock` function it is still incomplete. We don't really act upon the unmarshalled response in that function. You can find similar but complete implementation in the `DebugBlock` function for the same thing.

cffls (approved):
LGTM

JekaMas (approved):

PR #400, opened 2022-05-13
Lint consensus/bor

cffls:
- Cleanup, lint, and enable linters for consensus/bor module.
- Disabled linter "gomnd" and "tagliatelle" because they are not easy to fix and the return on the time investment is very low.
- Disabled linter "prealloc" because it is not easy to guess and pre-allocate the slice accurately in many cases.
- Changed nestif complexity from 3 to 5 to avoid changing too much existing logic.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,7 +28,7 @@ linters:
     - exportloopref
     - gocognit
     - gofmt
-    - gomnd
+    # - gomnd
     - gomoddirectives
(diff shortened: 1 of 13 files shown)


temaniarpit27 (approved):

cffls:
Closing this as the changes are also included in https://github.com/maticnetwork/bor/pull/410

PR #403, opened 2022-05-16
Implement and enable 'unlock' flag

cffls:
Currently in v0.3.0, validator couldn't mine new blocks, because the mining account is locked even when passed to "-unlock" flag. This commit will respect the functionality of flags "-unlock", "-password", and "-allow-insecure-unlock".


0xsharma (approved):
LGTM !

PR #410, opened 2022-05-20
Lint consensus/bor and internal/cli

cffls:
- Cleanup, lint, and enable linters for consensus/bor and internal/cli module.
- Disabled linter "gomnd" and "tagliatelle" because they are not easy to fix and the return on the time investment is very low.
- Changed nestif complexity from 3 to 5 to avoid changing too much existing logic.

### DIFF
--- a/internal/cli/bor_fingerprint.go
+++ b/internal/cli/bor_fingerprint.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/ethereum/go-ethereum/params"
+
 	"github.com/mitchellh/cli"
 	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/disk"
@@ -25,6 +26,7 @@ func (c *FingerprintCommand) MarkDown() string {
 		"# Fingerprint",
 		"Display the system fingerprint",
 	}
+
 	return strings.Join(items, "\n\n")
 }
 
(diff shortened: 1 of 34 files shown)


internal/cli/bor_fingerprint.go
```diff
@@ -101,9 +105,11 @@ func (c *FingerprintCommand) checkFio() error {
 // Run the IOPS benchmark for the node
 func (c *FingerprintCommand) benchmark() error {
 	var b []byte
+
 	err := c.checkFio()
+
 	if err != nil {
-		return nil
+		return err
```
cffls, line 112: @0xsharma Could you confirm if previous `return nil` was a typo, which is supposed to be `return err`?

temaniarpit27 (approved):

JekaMas (approved):

internal/cli/bor_fingerprint.go
JekaMas: @0xsharma Could you help with the question above?

0xsharma: @cffls @JekaMas 
It is supposed to be `return nil ` since it prints that "Fio is not installed"(in func checkFio) and does not perform benchmark if checkFio returns err. Also, we just want to gently exit the fingerprint test asking the user to install Fio and not output any error.

JekaMas: I see. Thank you!

@cffls I believe with this revert we can merge the pr.

cffls: Thanks @0xsharma @JekaMas ! I will revert this change and add a comment in the code.

PR #418, opened 2022-05-30
Increased default value of rpc.txfeecap to 5

0xKrishna:


ssandeep (approved):

0xKrishna:
There have been multiple requests from different RPC providers to increase the default from 1 to 5. The value of 1 makes sense for Ethereum but on our chain, we might need to increase it as the fee has crossed 1 on many occasions.

0xsharma (approved):
LGTM 🎉

PR #422, opened 2022-06-03
Mainnet block time analysis

manav2401:


miner/miner.go
```diff
@@ -68,9 +70,21 @@ type Miner struct {
 	stopCh   chan struct{}
 
 	wg sync.WaitGroup
+
+	log     *logrus.Logger
+	logFile *os.File
```
JekaMas, line 75: Can't we use standard ethereum logging?

consensus/bor/bor.go
```diff
@@ -779,6 +780,7 @@ func (c *Bor) FinalizeAndAssemble(chain consensus.ChainHeaderReader, header *typ
 				return nil, err
 			}
 		}
+		log.Info("[Mining Analysis] Commit span and state-syncs", "time", time.Since(start1))
```
JekaMas, line 783: Could you add block number and hash(if present) everywhere where it's about mining?

JekaMas: Also could you use one field name for elapsed time? Here it is "time" and below it's 'elapsed'.

consensus/bor/bor.go
```diff
@@ -1145,6 +1157,8 @@ func (c *Bor) fetchAndCommitSpan(
```
JekaMas, line 1157: That's a big piece of code to measure. Could you add intermediate points to measure?

consensus/bor/bor.go
manav2401: Sure, making relevant changes. Thanks!

JekaMas (dismissed):
LGTM!

PR #423, opened 2022-06-07
Add mining metrics for better understanding the time taken

temaniarpit27:


miner/worker.go
```diff
@@ -1172,6 +1184,11 @@ func (w *worker) commit(env *environment, interval func(), update bool, start ti
 	if update {
 		w.updateSnapshot(env)
 	}
+	if env.tcount == 0 {
+		minerZeroTxCommitMetrics.Update(time.Since(start))
```
JekaMas, line 1188: Now it looks as I expected)

miner/worker.go
```diff
@@ -77,6 +78,13 @@ const (
 	staleThreshold = 7
 )
 
+var (
+	minerZeroTxBroadcastMetrics    = metrics.NewRegisteredTimer("chain/miner/zerotx/broadcast", nil)
```
JekaMas, line 82: Why it is timer not histogram or gauge or counter?

temaniarpit27: counter and gauge basically don't calculate time

temaniarpit27: histogram I need to check but I think we can get that data from the timer itself on datadog I suppose

0xsharma (approved):

miner/worker.go
JekaMas: So it's mostly about block time. Ok then.

JekaMas (approved):

PR #425, opened 2022-06-08
RFC35/Common Ancestor: Modifying the forkchoice rule

manav2401:
This PR is the second step towards reducing reorgs and is a part of [RFC35](https://www.notion.so/polygontechnology/RFC-35-9f23b1ad5410451792e413640ffffca9). Refer https://github.com/maticnetwork/bor/pull/386 for the first step. 

It adds more functionality to the previously added checkpoint whitelist service. It performs some additional checks during the chain insertion process. The forkchoice module is responsible for checking against any incoming chain if we need to reorg or not using the `ReorgNeeded` function. It basically compared the total difficulty of the local chain and remote incoming chain. This PR adds some additional prevention by introducing a `ValidateReorg` which calls the checkpoint whitelist service to validate the chain. Some important validations are given below. 

- Past chain (where all blocks are <= last current block) case: check the chain against our local whitelist entry. 
- Future chain (where all blocks are > last current block) case: only accept a chain of some specific length (This number is fixed to 256 for now but will be made configurable later on). 

A minor modification over last iteration is that we now fetch N (latest) checkpoint entries instead of 1 in the first run to handle some sync cases for the client. Later on, the subsequent iterations only fetch 1 (latest) entry from heimdall. 

This will ideally prevent the node to get reorged to some wrong/invalid fork and will maintain consistency. Sufficient unit tests are added and devnet testing has been carried on for the same.


eth/downloader/whitelist/service.go
```diff
@@ -70,6 +72,96 @@ func (w *Service) IsValidChain(remoteHeader *types.Header, fetchHeadersByNumber
...
+	if from > to {
+		return nil
+	}
+
+	capacity := (to-from)/w.checkpointInterval + 1
+	res := make([]uint64, 0, capacity)
+
+	for n := from; n <= to; n++ {
+		//todo: check if it's correct
+		if n%w.checkpointInterval == 0 {
```
JekaMas, line 157: I believe we already have function like IsCheckpoint somewhere in the code. If not, we need to introduce one and remove code duplication.

eth/downloader/whitelist/service.go
```diff
@@ -70,6 +72,96 @@ func (w *Service) IsValidChain(remoteHeader *types.Header, fetchHeadersByNumber
...
+		return false
+	}
+
+	// Iterate over the chain and validate against the last checkpoint
+	// It will handle all cases where the incoming chain has atleast one checkpoint
+	for i := len(pastChain) - 1; i >= 0; i-- {
+		if _, ok := w.checkpointWhitelist[pastChain[i].Number.Uint64()]; ok {
+			if pastChain[i].Hash() != w.checkpointWhitelist[pastChain[i].Number.Uint64()] {
+				return false
+			} else {
```
JekaMas, line 132: you don't need `else` here as far as if-statement ends with `return`, just `break` if enough.

eth/downloader/whitelist/service.go
manav2401: I believe we can remove the functions `GetCheckpoints` and `GetCheckpointsBetween` from the whitelist service as they're no longer used. IsCheckpoint check I believe is just checking against the whitelist map, right? Should we also add it into function?

eth/downloader/whitelist/service.go
```diff
@@ -70,6 +72,96 @@ func (w *Service) IsValidChain(remoteHeader *types.Header, fetchHeadersByNumber
...
+	if last < oldestCheckpointNumber {
+		// We have future whitelisted entries, so no additional validation will be possible
+		// This case will occur when bor is in middle of sync, but heimdall is ahead/fully synced.
+		return true
+	}
+
+	// Start iterating over the chain and apply validation for past and future
+	// chain, referring to chain behind and after current header respectively.
+	for i := 0; i < len(chain); i++ {
+	}
```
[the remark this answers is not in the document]
manav2401: Nope, my bad

eth/downloader/whitelist/service.go
manav2401: You mean something like this?
`return pastChain[i].Hash() == w.checkpointWhitelist[pastChain[i].Number.Uint64()]`?

eth/downloader/whitelist/service.go
```diff
@@ -70,6 +72,96 @@ func (w *Service) IsValidChain(remoteHeader *types.Header, fetchHeadersByNumber
...
+		oldestCheckpointNumber uint64 = w.checkpointOrder[0]
+		first                  uint64 = chain[0].Number.Uint64()
+		last                   uint64 = chain[len(chain)-1].Number.Uint64()
+		current                uint64 = currentHeader.Number.Uint64()
+	)
+
+	// Check if we have checkpoints to validate incoming chain in memory
+	if len(w.checkpointWhitelist) == 0 {
+		// We don't have any entries, no additional validation will be possible
+		return true
```
JekaMas, line 88: won't it be possible to feed new peers with malicious chains?

interfaces.go
```diff
@@ -238,3 +238,13 @@ type StateSyncFilter struct {
 	ID       uint64
 	Contract common.Address
 }
+
+// interface for whitelist service
+type ChainValidator interface {
```
JekaMas, line 243: Why do we need 2 practically the same interfaces?

eth/downloader/whitelist/service.go
```diff
@@ -70,6 +72,96 @@ func (w *Service) IsValidChain(remoteHeader *types.Header, fetchHeadersByNumber
...
+
+	var (
+		pastChain   []*types.Header
+		futureChain []*types.Header
+	)
+	if current >= first {
+		if len(chain) == 1 {
+			pastChain = chain
+		} else {
+			pastChain = chain[:current-first]
```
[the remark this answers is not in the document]
manav2401: I treat the chain including all blocks till current block (inclusive) as past chain and remaining as future. 
So, if current block is 100, and we're receiving a chain from [90,105], then [90,100] is the past chain and [101,105] would be the future chain.

eth/downloader/whitelist/service.go
manav2401: So, do we want to consider the chain invalid, if we don't have local whitelist? As we just have a background process to gather whitelist and we only store them in memory I kind of think that there's good possibility of this map being empty or having old data. Wdyt?

interfaces.go
manav2401: Do you mean here and in `checker.go`?

eth/downloader/whitelist/service.go
JekaMas: ask from Heimdall for N checkpoints at the start of the service.

eth/downloader/whitelist/service_test.go
```diff
@@ -100,7 +102,105 @@ func TestIsValidChain(t *testing.T) {
...
+func createMockChain(start, end uint64) []*types.Header {
+	var i uint64
+	var chain []*types.Header
+	for i = start; i <= end; i++ {
+		header := &types.Header{
+			Number: big.NewInt(int64(i)),
+			Time:   uint64(time.Now().UnixMicro()),
+		}
+		chain = append(chain, header)
+		time.Sleep(time.Microsecond) // just to be safe
```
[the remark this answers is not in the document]
JekaMas: header := &types.Header{
			Number: big.NewInt(int64(i)),
			Time:   uint64(time.Now().UnixMicro())+1,
		}

eth/downloader/whitelist/service.go
manav2401: Done!

JekaMas (approved):
looks good!

eth/downloader/downloader.go
```diff
@@ -792,8 +792,10 @@ func (d *Downloader) getFetchHeadersByNumber(p *peerConnection) func(number uint
 // the head links match), we do a binary search to find the common ancestor.
 func (d *Downloader) findAncestor(p *peerConnection, remoteHeader *types.Header) (uint64, error) {
 	// Check the validity of peer from which the chain is to be downloaded
-	if _, err := d.IsValidPeer(remoteHeader, d.getFetchHeadersByNumber(p)); err != nil {
-		return 0, err
+	if d.ChainValidator != nil {
```
JekaMas, line 795: How could it be nil? Should we initiate it somewhere?

JekaMas (approved):

0xKrishna (approved):

PR #431, opened 2022-06-14
Reciept e2e test

temaniarpit27:


.github/workflows/ci.yml
```diff
@@ -61,9 +61,9 @@ jobs:
         #       echo >&2 "Reproducible build broken"; cat bor1.sha256; cat bor2.sha256; exit 1
         #     fi
 
-        - name: Lint
-          if: runner.os == 'Linux'
-          run: make lint
+        # - name: Lint
+        #   if: runner.os == 'Linux'
+        #   run: make lint
```
cffls, line 66: Why is linter disabled?

JekaMas: It was inconsistent comparing to linux.

cffls: Okay, but linter will only run when the os is linux right? Or are you saying even for linux we get inconsistent results?

JekaMas: Only mac on CI was inconsistent.

cffls: Okay, so we shouldn't comment this out right, since it is consistent on Linux?

JekaMas: Yeap, this is a mistake. Could you even try to run CI linters on both OSs?

cffls: Yes you can. It will be under two separate jobs. The code is doing the right thing, only allowing linter to run when os is linux.

JekaMas: Like remove if statement from the code and restore make lint. Let's try it on both oss once again.

temaniarpit27:
@JekaMas please fix linters
I see some deadcode but it is being used

JekaMas:
@temaniarpit27 @0xsharma looking for your review
* also added a task to rewrite fragile tests in `tests/bor` to the backlog

cffls (approved):

PR #433, opened 2022-06-15
Merge branch 'v0.3.0-dev' into develop

cffls:
This PR will merge everything from `v0.3.0-dev` into `develop` branch.


consensus/bor/bor.go
```diff
@@ -253,7 +254,11 @@ func New(
 		signatures:             signatures,
 		spanner:                spanner,
 		GenesisContractsClient: genesisContracts,
-		HeimdallClient:         heimdallClient,
+	}
+
+	// use reflection to check if the underlying value behind the interface is actually nil
+	if !(heimdallClient == nil || reflect.ValueOf(heimdallClient).IsNil()) {
```
JekaMas, line 260: I believe we don't need reflection here, since we are not checking all interfaces that way.
Just linters and tests should do the thing.

JekaMas (commented):
One comment about checking nil interface

consensus/bor/bor.go
cffls: Initially, I didn't have reflection here, but `heimdallClient == nil` is evaluated to `false` everywhere. I even print the heimdallClient and this is the output:

code:
```go
log.Info("OK", "heimdallClient", heimdallClient)
log.Info("OK", "heimdallClient", heimdallClient == nil)
```

output:
```
INFO [06-15|23:46:07.200] OK                                       heimdallclient=<nil>
INFO [06-15|23:46:07.200] OK                                       heimdallclient=false
```

Do you have a better solution for this?

JekaMas: But do we pass nil Heimdall client somewhere ? Is we do and pass incorrectly then it is possible to have that case. But it is just something to fix and pass interface correctly.
If we don't pass nil, then I am not sure, what you checked.

cffls: Yes, we do need to pass heimdallClient as nil when `withoutheimdall` flag is true. See here: https://github.com/maticnetwork/bor/pull/433/commits/1c1e61355d44fe3592d7496aeac9371b4d338237#diff-93626ed7f697f5674903a96e7153e88a4f2d8596df186489fb0c26557eb80889R248

JekaMas: And that exact place returns heimdallClient != Nil if it is set to nil?

cffls: No, `heimdallClietn == nil` is giving me the correct result. The result is only incorrect in [this function](https://github.com/maticnetwork/bor/pull/433/commits/1c1e61355d44fe3592d7496aeac9371b4d338237#diff-e2acd76a9636afbcc13d882e2f46ca9bf74e0f9fa6c190c7ba75328a98fd364fR229) and after. The cause is that `New` function is taking heimdallClient as an interface `IHeimdallClient`.

JekaMas: This is not the reason.

JekaMas: https://go.dev/play/p/lkE-NlO6Hht

JekaMas: The solution is in the link above.

cffls: The suggested solution works. Updated the commit, which looks much cleaner now. Thank you!

JekaMas: Will discuss golang interfaces and catches soon :)

temaniarpit27 (approved):

JekaMas (dismissed):

temaniarpit27 (approved):

JekaMas (approved):

PR #435, opened 2022-06-21
Adding correct miner address on API at the RPC level (POS-553)

pratikpatil024:
This PR changes the miner field in the block (0x0, Coinbase) with the original miner address.

[POS-553](https://linear.app/matic/issue/POS-533/adding-correct-miner-address-on-api-at-the-rpc-level).

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -814,16 +814,15 @@ func (s *PublicBlockChainAPI) GetHeaderByHash(ctx context.Context, hash common.H
 	return nil
 }
 
-// getHeaderWithAuthor: changes the miner (0x0, coinbase) with the original miner address
-func (s *PublicBlockChainAPI) getHeaderWithAuthor(head *types.Header) error {
+// getAuthor: returns the author of the Block
+func (s *PublicBlockChainAPI) getAuthor(head *types.Header) (*common.Address, error) {
 	// get author using From: Backend -> Engine -> Author
 	author, err := s.b.Engine().Author(head)
 	if err != nil {
-		return err
+		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


temaniarpit27 (approved):

internal/ethapi/api.go
```diff
@@ -848,6 +870,14 @@ func (s *PublicBlockChainAPI) GetBlockByHash(ctx context.Context, hash common.Ha
 		response, err := s.rpcMarshalBlock(ctx, block, true, fullTx)
 		// append marshalled bor transaction
 		if err == nil && response != nil {
+
+			author, err := s.getAuthor(block.Header())
+			if err != nil {
+				return nil, err
```
manav2401, line 876: Do we want to return an error from here or just return the block with `0x00` miner address?

manav2401: I agree. What I mean to say is `getBlockByNumber` is highly used function. Any errors with our `getAuthor` implementation should not ideally cause error in this function. We're coupling these 2 components for this.

pratikpatil024: > I agree. What I mean to say is `getBlockByNumber` is highly used function. Any errors with our `getAuthor` implementation should not ideally cause error in this function. We're coupling these 2 components for this.

I agree.

temaniarpit27: good point manav
ideally the functionality should not change of the public facing function
if we are able to find the value from getAuthor, it should return the value otherwise it should return "0x0000..." (previous functionality)

pratikpatil024: Okay, will do that change rn.

PR #441, opened 2022-06-28
core/forkchoice: abstract chain validator from reorg check, validate reorg while inserting

manav2401:


JekaMas (commented):
Lgtm, do we need to change forkChoiseRule interface as well?

manav2401:
> Lgtm, do we need to change forkChoiseRule interface as well?

I don't think so. It has chain validator which we can consume in another function, right? Or are you saying in some other context?

manav2401:
Closing this as have cherry picked the commit.

PR #443, opened 2022-07-06
Create MVHashMap and use it in StateDB

cffls:
The implementation of MVHashMap slightly differs from the original implementation by storing an instance of statedb, instead of the actual data bytes, as the value of a key in the map.

### DIFF
--- a/core/blockstm/executor.go
+++ b/core/blockstm/executor.go
@@ -1,0 +1,241 @@
+package blockstm
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type ExecResult struct {
+	err      error
+	ver      Version
+	txIn     TxnInput
+	txOut    TxnOutput
+	txAllOut TxnOutput
+}
+
+type ExecTask interface {
+	Execute(mvh *MVHashMap, incarnation int) error
+	MVReadList() []ReadDescriptor
+	MVWriteList() []WriteDescriptor
+	MVFullWriteList() []WriteDescriptor
+}
+
+type ExecVersionView struct {
+	ver Version
+	et  ExecTask
+	mvh *MVHashMap
+}
+
+func (ev *ExecVersionView) Execute() (er ExecResult) {
+	er.ver = ev.ver
+	if er.err = ev.et.Execute(ev.mvh, ev.ver.Incarnation); er.err != nil {
+		log.Debug("blockstm executed task failed", "Tx index", ev.ver.TxnIndex, "incarnation", ev.ver.Incarnation, "err", er.err)
+		return
+	}
+
+	er.txIn = ev.et.MVReadList()
+	er.txOut = ev.et.MVWriteList()
+	er.txAllOut = ev.et.MVFullWriteList()
+	log.Debug("blockstm executed task", "Tx index", ev.ver.TxnIndex, "incarnation", ev.ver.Incarnation, "err", er.err)
+
+	return
+}
+
+var ErrExecAbort = fmt.Errorf("execution aborted with dependency")
+
+const numGoProcs = 4
+
+// nolint: gocognit
+func ExecuteParallel(tasks []ExecTask) (lastTxIO *TxnInputOutput, err error) {
+	if len(tasks) == 0 {
+		return MakeTxnInputOutput(len(tasks)), nil
+	}
+
+	chTasks := make(chan ExecVersionView, len(tasks))
+	chResults := make(chan ExecResult, len(tasks))
+	chDone := make(chan bool)
+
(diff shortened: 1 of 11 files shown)


core/state/mvhashmap.go
```diff
@@ -0,0 +1,180 @@
...
+	return &MVHashMap{
+		rw: sync.RWMutex{},
+		m:  make(map[string]*TxnIndexCells),
+	}
+}
+
+type WriteCell struct {
+	flag        uint
+	incarnation int
+	data        *StateDB
```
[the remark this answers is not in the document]
paulgoleary: LGTM

core/state/txio.go
```diff
@@ -0,0 +1,75 @@
...
+type ReadDescriptor struct {
+	Path []byte
+	Kind int
+	V    Version
+}
+
+type WriteDescriptor struct {
+	Path []byte
+	V    Version
+	Val  *StateDB
```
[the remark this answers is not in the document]
paulgoleary: As above, I understand how this works now and it looks good.

core/state/statedb.go
```diff
@@ -154,6 +162,161 @@ func New(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB, error)
...
+
+	switch res.Status() {
+	case blockstm.MVReadResultDone:
+		{
+			v = readStorage(res.Value().(*StateDB))
+			rd.Kind = blockstm.ReadKindMap
+		}
+	case blockstm.MVReadResultDependency:
+		{
+			return defaultV
```
paulgoleary, line 242: As discussed, we will need some way to capture the fact that the read hit a dependency so we can abort the execution later. There is a `setError` method on StateDB that accomplishes something similar for db errors. Something simple like a `setReadDep` method that sets a boolean will likely do just fine.

paulgoleary (commented):
LGTM now. See the comment about setting a flag to indicate that execution hit a read dependency.

core/state/statedb.go
cffls: Added `HadInvalidRead` function to indicate the read error.

core/blockstm/executor.go
```diff
@@ -0,0 +1,241 @@
...
+type ExecResult struct {
+	err      error
+	ver      Version
+	txIn     TxnInput
+	txOut    TxnOutput
+	txAllOut TxnOutput
+}
+
+type ExecTask interface {
+	Execute(mvh *MVHashMap, incarnation int) error
```
cffls, line 18: Since we can't perfectly abstract MVHashmap from statedb, changed the `Execute` interface a bit by replacing the read/write interface with MVHashmap directly.

core/blockstm/executor.go
```diff
@@ -0,0 +1,241 @@
...
+	MVReadList() []ReadDescriptor
+	MVWriteList() []WriteDescriptor
+	MVFullWriteList() []WriteDescriptor
```
cffls, line 21: Added a new write set that represent the full write list. `MVFullWriteList` will be a parent set of `MVWriteList`. This is needed because we had to store state objects in MVHashmap (the records whose key is only an address without any subpath), but we don't want to detect it as a write conflict during validation. Therefore, in validation, only `MVWriteList` will be checked, while when recording writes to MVHashmap upon the completion of a task, we will write MVFullWriteList to MVHashmap.

core/parallel_state_processor.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Process processes the state changes according to the Ethereum rules by running
+// the transaction messages using the statedb and applying any rewards to both
+// the processor (coinbase) and any included uncles.
+//
+// Process returns the receipts and logs accumulated during the process and
+// returns the amount of gas that was used in the process. If any of the
+// transactions failed to execute due to insufficient gas it will return an error.
+func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg vm.Config) (types.Receipts, []*types.Log, uint64, error) {
```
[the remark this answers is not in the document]
pratikpatil024: Resolved, thanks.

pratikpatil024 (approved):

paulgoleary:
how can we get @pratikpatil024's branch / work on here too? do we need to add a flag a/o parameter to the API to make the dump optional? would be useful to be able to specify the target location of the output file.

cffls:
> how can we get @pratikpatil024's branch / work on here too? do we need to add a flag a/o parameter to the API to make the dump optional? would be useful to be able to specify the target location of the output file.

I will merge this one first. Then cherry-pick @pratikpatil024 's commit with some minor changes. We can specify the location by passing a parameter through the API. I will double check this.

PR #448, opened 2022-07-08
internal/cli: update protobuf dependency, fix linters

manav2401:
This PR updates the deprecated [protobuf](https://pkg.go.dev/github.com/golang/protobuf/proto) dependency used in the new-cli to this [updated](https://pkg.go.dev/google.golang.org/protobuf) package. Moreover, it also updates some types consumed from the old library with the new ones. 

Another point to note is that earlier we used to wrap the stream into a `net.Conn` instance using the [go-grpc-net-conn](https://github.com/mitchellh/go-grpc-net-conn) library but unfortunately it didn't support the new protobuf version. So, we've migrated to a [fork](https://github.com/hashicorp/go-grpc-net-conn) of the same library which is maintained by hashicorp and has support for the new protobuf version.  

It also includes some fixes suggested by linters in other packges. 

Jira ticket: [POS-497 - Update protobuf/grpc dependencies in new-cli](https://polygon.atlassian.net/browse/POS-497)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ devtools:
 	$(GOBUILD) -o $(GOBIN)/codecgen github.com/ugorji/go/codec/codecgen
 	$(GOBUILD) -o $(GOBIN)/abigen ./cmd/abigen
 	$(GOBUILD) -o $(GOBIN)/mockgen github.com/golang/mock/mockgen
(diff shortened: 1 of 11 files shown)


manav2401:
Closing in favour of https://github.com/maticnetwork/bor/pull/449.

Issue #447, opened 2022-07-08
Unexpected trx hashes in eth_getTransactionReceiptsByBlock


dsoldatow:
#### System information

Geth version: `0.2.14-stable-28963d5c`
OS & Version: Linux
Commit hash : 28963d5ccd3028e0ac6f1915845c747862d649d1
eth_getTransactionReceiptsByBlock with params: ["0x1cfe800"]

#### Expected behaviour
All transaction hashes exist in receipt objects

#### Actual behaviour
First receipt:
    "transactionHash": "0x615141657f68a41c33ae2ed087f38782812e1fc386aea0b8a51a88fbd0e86d47",
    "transactionIndex": "0x0"
Second receipt:
   "transactionHash": "0x615141657f68a41c33ae2ed087f38782812e1fc386aea0b8a51a88fbd0e86d47",
    "transactionIndex": "0x1"
Indexes are different, but hashes are the same

And in all receipts indexes are different, but hash only one.

Full response: [receipts.txt](https://github.com/maticnetwork/bor/files/9070408/receipts.txt)
#### Steps to reproduce the behaviour
curl --location --request POST 'NODE_URL' \
--header 'Content-Type: application/json' \
--data-raw '{
   "jsonrpc": "2.0",
   "id": 1,
   "method": "eth_getTransactionReceiptsByBlock",
   "params": ["0x1cfe800"]
}'

I am writing to you on behalf of the [WatchData.io](https://watchdata.io) (blockchain data provider). We are adding polygon to our system now and this bug is important for us.

dsoldatow:
Ok. I will check, but not today

kfastov:
I found that if there is a "bor transaction" in a block, each transaction gets a hash of that bor transaction due to an error in the condition.
https://github.com/maticnetwork/bor/blob/04e90e2426ac257b7db812478d2370a8b8d0fab2/internal/ethapi/api.go#L683-L685

dsoldatow:
yeah, i tryed tag: 0.2.16 on commit hash: f083705e00b6497e575a6e6c800aad661a4a2eca and got the same.

dsoldatow:
Polygon API is available now on wachdata.io. @JekaMas, What about add WatchData.io to https://docs.polygon.technology/docs/develop/network-details/network/

0xKrishna:
I think the issue is resolved, You can contact support team on discord for the above request, They will help you with that for sure.
Closing the issue for now.

PR #453, opened 2022-07-11
Fix transaction hash assignment in getTransactionReceiptsByBlock

kfastov:
There is a bug in getTransactionReceiptsByBlock which causes every transaction in a block take the same transaction hash if that block contains a bor transaction because of a loose condition in `if` expression. I've added an additional check to fix that. Since the bor transaction is always the last in a block, we can check if it's the bor transaction by comparing its index with length of the receipts array minus one.
The related issue is #447

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -680,7 +680,7 @@ func (s *PublicBlockChainAPI) GetTransactionReceiptsByBlock(ctx context.Context,
 		if receipt.Logs == nil {
 			fields["logs"] = [][]*types.Log{}
 		}
-		if borReceipt != nil && idx == len(receipts) - 1 {
+		if borReceipt != nil && idx == len(receipts)-1 {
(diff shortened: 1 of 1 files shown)


kfastov:
Fixed linter error

JekaMas (approved):

shiziwen:
Hi, I found this issue has been fixed at https://github.com/maticnetwork/bor/pull/365.

manav2401:
> Hi, I found this issue has been fixed at https://github.com/maticnetwork/bor/pull/365.

Hi, yes this issue has been address by our team and it should be available in the next release. Thanks for the PR. Meanwhile you can use this branch (a fork of v0.2.16) if you want: https://github.com/maticnetwork/bor/tree/PR-365. Closing this.

JekaMas:
@kfastov Many thanks from my side! The release will be soon.

PR #456, opened 2022-07-14
Cover Bor Filters with testcases

0xsharma:
#### The PR contains the following changes : 
- Workaround to create state-sync tx manually for testcases.
- Func TestBorFilters() : test filters for bor-transactions.
- Some miner changes with sprint size. Sprint size will be determined from the genesis where as it was hardcoded to 64 before.

Link to [Jira Task](https://polygon.atlassian.net/browse/POS-636?atlOrigin=eyJpIjoiZmViOWU1OWM0YzUzNDdhMDkxMjAyZTQ3ZGQwZmRiMzQiLCJwIjoiaiJ9).


eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	dir, err := ioutil.TempDir("", "filtertest")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer os.RemoveAll(dir)
+
+	var (
+		db, _   = rawdb.NewLevelDBDatabase(dir, 0, 0, "", false)
```
JekaMas, line 31: Could we stick with in memory DB?

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,189 @@
...
+		hash3 = common.BytesToHash([]byte("topic3"))
+		hash4 = common.BytesToHash([]byte("topic4"))
+	)
+
+	defer db.Close()
+
+	genesis := core.GenesisBlockForTesting(db, addr, big.NewInt(1000000))
+	sprint := params.TestChainConfig.Bor.Sprint
+
+	chain, receipts := core.GenerateChain(params.TestChainConfig, genesis, ethash.NewFaker(), db, 1000, func(i int, gen *core.BlockGen) {
```
JekaMas, line 47: Is it possible to test bor_filter in isolation with mocks?

eth/filters/bor_filter_test.go
0xsharma: Yes, we can do that too.

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (backend *MockBackend) expectBorReceiptsFromMock(hashes []*common.Hash) {
+	for i := range hashes {
+		if hashes[i] == nil {
+			backend.EXPECT().GetBorBlockReceipt(gomock.Any(), gomock.Any()).Return(nil, nil)
+			continue
+		}
+
+		backend.EXPECT().GetBorBlockReceipt(gomock.Any(), gomock.Any()).Return(newTestReceipt(addr, *hashes[i]), nil)
```
JekaMas, line 56: could we use less gomock.Any, rather then specify expected input values where it's possible?

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,164 @@
...
+	addr    = crypto.PubkeyToAddress(key1.PublicKey)
+
+	hash1 = common.BytesToHash([]byte("topic1"))
+	hash2 = common.BytesToHash([]byte("topic2"))
+	hash3 = common.BytesToHash([]byte("topic3"))
+	hash4 = common.BytesToHash([]byte("topic4"))
+)
+
+func newTestHeader(blockNumber uint) *types.Header {
+	head := types.Header{
```
JekaMas, line 28: return &types.Header{
		Number: big.NewInt(int64(blockNumber)),
	}

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,164 @@
...
+		Number: big.NewInt(int64(blockNumber)),
+	}
+
+	return &head
+}
+
+func newTestReceipt(contractAddr common.Address, topicAddress common.Hash) *types.Receipt {
+	receipt := types.NewReceipt(nil, false, 0)
```
JekaMas, line 36: if you have the same logs for all receipts, then you might not catch issues in the code.

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+		backend.EXPECT().GetBorBlockReceipt(gomock.Any(), gomock.Any()).Return(newTestReceipt(addr, *hashes[i]), nil)
+	}
+}
+
+func TestBorFilters(t *testing.T) {
+	t.Parallel()
+
+	var (
+		db = rawdb.NewMemoryDatabase()
```
JekaMas, line 64: Why do we need it? could we remove db from test or mock it, if it's not possible to remove?

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/ethereum/go-ethereum/params"
+
+	gomock "github.com/golang/mock/gomock"
+)
+
+var (
+	key1, _ = crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
+	addr    = crypto.PubkeyToAddress(key1.PublicKey)
+
+	hash1 = common.BytesToHash([]byte("topic1"))
```
JekaMas, line 21: could we remove package-level variables? I believe we only need a bunch of hashes that can be made inside a test or in a function.

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+	}
+
+	receipt.Bloom = types.CreateBloom(types.Receipts{receipt})
+
+	return receipt
+}
+
+func (backend *MockBackend) expectBorReceiptsFromMock(hashes []*common.Hash) {
+	for i := range hashes {
```
JekaMas, line 50:
```
	for i, h := range hashes {
		expect := backend.EXPECT().GetBorBlockReceipt(gomock.Any(), gomock.Any())

		if h == nil {
			exp.Return(nil, nil)
			continue
		}

		expect.Return(newTestReceipt(addr, *h), nil)
	}
}
```

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,164 @@
...
+	backend.expectBorReceiptsFromMock([]*common.Hash{&hash1, &hash3})
+
+	filter = NewBorBlockLogsRangeFilter(backend, sprint, 990, 999, []common.Address{addr}, [][]common.Hash{{hash3}})
+	logs, _ = filter.Logs(context.Background())
+
+	if len(logs) != 1 {
+		t.Error("expected 1 log, got", len(logs))
+	}
+
+	if len(logs) > 0 && logs[0].Topics[0] != hash3 {
```
JekaMas, line 104: do we need `len(logs) > 0` here?

eth/filters/bor_filter_test.go
0xsharma: The logs differ, the topic address is different for every log.

0xsharma: https://github.com/maticnetwork/bor/blob/58ef6fdbb7ecc37a3eeebd0c0410917b00268327/eth/filters/bor_filter_test.go#L56

eth/filters/bor_filter_test.go
0xsharma: I also had to add a case for nil. Added it.

eth/filters/bor_filter_test.go
0xsharma: Yes, to save from panic. In case if len(logs)=0, the above t.Error on `len(logs) != 1 ` would although fail the case but it would still continue execution and panic on if `len(logs) > 0` is not used.

eth/filters/bor_filter_test.go
0xsharma: It doesn't rely on the input values. And yes I tried sending in the expected input but it is a lot of hardcoding even though the output is irrespective of the inputs in this case.

eth/filters/bor_filter_test.go
0xsharma: Done, used mock db.

eth/filters/bor_filter_test.go
```diff
@@ -0,0 +1,158 @@
...
+		},
+	}
+
+	receipt.Bloom = types.CreateBloom(types.Receipts{receipt})
+
+	return receipt
+}
+
+func (backend *MockBackend) expectBorReceiptsFromMock(hashes []*common.Hash) {
+	for _, h := range hashes {
```
JekaMas, line 42: I believe something like that already was somewhere in the comments. Could you remove code duplication?

Also is it possible to review all gomock.Any() usage and use it as least as possible?

eth/filters/test_backend.go
```diff
@@ -0,0 +1,173 @@
+package filters
```
JekaMas, line 1: is it a generated file? if so where the comments at the beginning of the file?

0xsharma: No, it is not a generated file.

eth/filters/bor_filter_test.go
0xsharma: https://github.com/maticnetwork/bor/pull/456#discussion_r924367178

https://github.com/maticnetwork/bor/pull/456#discussion_r924368597

This might answer this.

0xsharma: For gmock.Any() : https://github.com/maticnetwork/bor/pull/456#discussion_r924383586

0xsharma: Also, https://github.com/maticnetwork/bor/pull/456#discussion_r924378572

eth/filters/bor_filter_test.go
JekaMas: I see now.  `backend.EXPECT().GetBorBlockReceipt(gomock.Any(), gomock.Any()).Return(newTestReceipt(addr, *hashes[i]), nil)` hashes doesn't stand for hash, but rather for topics.
No questions then

JekaMas (approved):

JekaMas:
@0xsharma Could you fix the CI?

0xKrishna (approved):

PR #458, opened 2022-07-18
consensus/bor/heimdall: add mock heimdall rest-server, tests for timeout and cancellation

manav2401:
This PR adds a setup for using mock heimdall rest-server for testing various components of bor heimdall interaction. It creates a simple http-server with a mock handler for fetching checkpoints (for now), which can be updated throughout to the test case for simulating different scenarios. Moreover, additional handlers for different routes (span, state-sync, etc) can be added as per requirement. 

It tests timeout and cancellations in the heimdall client's `FetchWithRetry` function using timeout based context (added in https://github.com/maticnetwork/bor/pull/445)

Some of the cases tests are as follows: 
1. Handling delay in response from server causing error due to context timeout.
2. Handling cancellation of context
3. Handling user interrupt (leading to shutting of heimdall client). 

Moreover, this PR also adds tests for some additional scenarios for context timeout and cancellations independent of the client module. 

Jira: [POS-640 - Use mock heimdall client for testing and test context cancellation/timeout](https://polygon.atlassian.net/browse/POS-640)

### DIFF
--- a/consensus/bor/bor_test.go
+++ b/consensus/bor/bor_test.go
@@ -1,15 +1,13 @@
 package bor
 
 import (
-	"context"
 	"math/big"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/hexutil"
-	"github.com/ethereum/go-ethereum/consensus/bor/heimdall" //nolint:typecheck
+	"github.com/ethereum/go-ethereum/common/hexutil" //nolint:typecheck
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/state"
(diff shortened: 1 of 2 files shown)


consensus/bor/heimdall/client_test.go
```diff
@@ -1,9 +1,317 @@
...
+	"github.com/ethereum/go-ethereum/consensus/bor/heimdall/checkpoint"
+
+	"github.com/stretchr/testify/require"
 )
 
+var maxPortCheck int32 = 100
+
+// HttpHandlerFake defines the handler functions required to serve
+// requests to the mock heimdal server for specific functions. Add more handlers
+// according to requirements.
```
JekaMas, line 26: Can we try https://pkg.go.dev/net/http/httptest ?

consensus/bor/bor_test.go
```diff
@@ -142,48 +140,3 @@ func TestEncodeSigHeaderJaipur(t *testing.T) {
 	hash = SealHash(h, &params.BorConfig{JaipurBlock: 10})
 	require.Equal(t, hash, hashWithoutBaseFee)
 }
-
-// TestCheckpoint can be used for to fetch checkpoint
-// count and checkpoint for debugging purpose.
-// Also, this is kept only for local use.
-func TestCheckpoint(t *testing.T) {
```
JekaMas, line 149: Why do you removed this test?

manav2401: It was only for debugging purpose (checking if all entities are correct, marshalling/unmarshalling takes place correctly or not, etc). Now that we have the mock server, all these checks can be easily added in the same test and can also be used for debugging by changing the heimdall client url.

consensus/bor/heimdall/client_test.go
manav2401: Will check it.

JekaMas (approved):

consensus/bor/heimdall/client_test.go
manav2401: Resolving this as per our conversation. This package is used to create mock requests and test the server side logic. As we have a heimdall client which has additional methods, it's good to test things using those methods. In order to do that, the easiest way for now is a mock server.

0xKrishna (approved):

PR #465, opened 2022-07-22
eth: fix multiple checkpoint fetch logic

manav2401:
This PR fixes a bug that was introduced in the RFC35 forkchoice implementaiton (https://github.com/maticnetwork/bor/pull/425). When the whitelisting service tries to fetch checkpoints, it fetches them in decreasing order starting from latest checkpoint to the oldest one. So, when a new checkpoint arrives, instead of eliminating the oldest checkpoint, it eliminated the latest one. We now fetch the checkpoints in an increasing order to prevent this from happening.


JekaMas:
Could you add a test for the issue?

manav2401:
> Could you add a test for the issue?

I actually tested this on a running node. But sure, I'll add some tests for this whole function to make it more deterministic.

temaniarpit27 (approved):

eth/bor_checkpoint_verifier.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+type checkpointVerifier struct {
+	verify func(ctx context.Context, handler *ethHandler, checkpoint *checkpoint.Checkpoint) (string, error)
+}
+
+func newCheckpointVerifier(verifyFn func(ctx context.Context, handler *ethHandler, checkpoint *checkpoint.Checkpoint) (string, error)) *checkpointVerifier {
+	if verifyFn == nil {
```
JekaMas, line 18: could you revert logic to reduce complexity?
like

```
if if verifyFn != nil {
    return &checkpointVerifier{verifyFn}
} 
```

eth/bor_checkpoint_verifier.go
```diff
@@ -0,0 +1,53 @@
...
+			}
+
+			// fetch the end checkpoint block hash
+			block, err := handler.ethAPI.GetBlockByNumber(ctx, rpc.BlockNumber(checkpoint.EndBlock.Uint64()), false)
+			if err != nil {
+				log.Debug("Failed to get end block hash of checkpoint while whitelisting", "err", err)
+				return "", errEndBlock
+			}
+
+			hash := fmt.Sprintf("%v", block["hash"])
```
JekaMas, line 46: is it equal to `hash := block["hash"].String()` ?

JekaMas (approved):

eth/bor_checkpoint_verifier.go
manav2401: Nope. I don't think it has `String()` method.

JekaMas (approved):

JekaMas:
@temaniarpit27 I believe, we need that one in v0.3.0

temaniarpit27 (approved):

PR #468, opened 2022-07-25
Implemented gRPC client for heimdall

0xKrishna:
This PR contains :

1. gRPC client implementation for the Heimdall server
2. Implemented all the methods in the current Heimdall interface

To use gRPC you have to pass gRPC server address using
```--bor.heimdallgRPC=<address>```
If you give the above flag then it will use gRPC, Otherwise it will continue to use REST APIs.

Default port in Heimdall is `3132`
So for local the gRPC address is `localhost:3132`.

https://polygon.atlassian.net/browse/POS-648

### CODE REVIEW

**Review by JekaMas — APPROVED** [review]

File: consensus/bor/heimdallgrpc/client.go [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+type HeimdallGRPCClient struct {
+	conn   *grpc.ClientConn
+	client proto.HeimdallClient
+}
+
+func NewHeimdallGRPCClient(address string) *HeimdallGRPCClient {
+	conn, err := grpc.Dial(address, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Error("Failed to connect to Heimdall gRPC server: %v", err)
+		panic(err)
```
manav2401, line 24: Can we avoid `panic`?

consensus/bor/heimdallgrpc/span.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func (h *HeimdallGRPCClient) Span(ctx context.Context, spanID uint64) (*span.HeimdallSpan, error) {
+	req := &proto.SpanRequest{
+		ID: spanID,
+	}
+
+	res, err := h.client.Span(context.Background(), req)
+	if err != nil {
+		return nil, err
```
manav2401, line 20: Won't we need a retry mechanism, specifically for `span`? Also, I think it's good to have it also for state-sync & checkpoint methods.

consensus/bor/heimdallgrpc/state_sync.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/consensus/bor/clerk"
+
+	proto "github.com/maticnetwork/polyproto/heimdall"
+)
+
+func (h *HeimdallGRPCClient) StateSyncEvents(ctx context.Context, fromID uint64, to int64) ([]*clerk.EventRecordWithTime, error) {
+	eventRecords := make([]*clerk.EventRecordWithTime, 0)
+
+	req := &proto.StateSyncEventsRequest{
```
JekaMas, line 15: I am more than sure that it's possible to remove all this manual unmarshal. By the time, as we discussed, let's keep it as it as. Could you create a task for bor and heimdall to find a bit better solution? We can rethink that place later.

consensus/bor/heimdallgrpc/span.go
JekaMas: Why you use blank context here?

consensus/bor/heimdallgrpc/state_sync.go
JekaMas: @0xKrishna could we update PRs (Heimdall and Bor), fix CI and merge gRPC PRs?

consensus/bor/heimdallgrpc/client.go
0xKrishna: Can use `log.Fataf`
I will check more options.

consensus/bor/heimdallgrpc/span.go
0xKrishna: Fixed the context

0xKrishna: Good point regarding retry, I will check

consensus/bor/heimdallgrpc/state_sync.go
0xKrishna: Lint and CI and fixed, Working on few comments above

consensus/bor/heimdallgrpc/client.go
0xKrishna: Using `log.Fatal` now, Will push soon.

consensus/bor/heimdallgrpc/span.go
0xKrishna: Fixed the context, Looking to implement retry mechanism.

0xKrishna: Implemented retry middleware, Now need to add supported errors on heimdall side.

0xsharma (approved):

manav2401 (approved):

PR #470, opened 2022-07-26
consensus/ethhash: skip TestRemoteMultiNotifyFull in sealer test

manav2401:
This PR skips one of the tests named `TestRemoteMultiNotifyFull` in the `consensus/ethhash/sealer_test.go` package. It is not a deterministic test and hence is one of primary cause of the failing CI in recent PRs. As we aren't dependent on the ethhash consensus, we can disable it for the time being. Later, according to bandwidth, we can identify the reasons and make it deterministic. A task for the same can be found [here](https://polygon.atlassian.net/browse/POS-654).


pratikpatil024 (approved):

JekaMas (approved):

consensus/ethash/sealer_test.go
```diff
@@ -167,6 +167,8 @@ func TestRemoteMultiNotify(t *testing.T) {
 // Tests that pushing work packages fast to the miner doesn't cause any data race
 // issues in the notifications. Full pending block body / --miner.notify.full)
 func TestRemoteMultiNotifyFull(t *testing.T) {
+	t.Skip()
```
JekaMas, line 170: Could you use t.SkipNow("and mention a reason of skipping here") ?

manav2401: I'll still need to use `t.Skip("reason")` as `t.Skip()` is log followed by `t.SkipNow()` call.

PR #478, opened 2022-08-02
internal/cli: add support for removedb

manav2401:
This PR adds support for removedb in the new-cli with similar functionality as in the old cli. 

It can be used as `bor removedb <datadir>` and is responsible to clear the blockchain and state databases. 

Note: It assumes that the `ancient` db lies inside the `chaindata` folder in datadir as we don't have access to the `DatabaseFreezer` path in the code. 

Jira - [POS-699: Support removedb in new-cli](https://polygon.atlassian.net/browse/POS-699)

### DIFF
--- a/internal/cli/removedb.go
+++ b/internal/cli/removedb.go
@@ -1,0 +1,145 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/internal/cli/flagset"
+	"github.com/ethereum/go-ethereum/internal/cli/server"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/node"
+
+	"github.com/mitchellh/cli"
+)
+
+// RemoveDBCommand is for removing blockchain and state databases
+type RemoveDBCommand struct {
+	*Meta2
+
+	datadir string
+}
+
+// MarkDown implements cli.MarkDown interface
+func (c *RemoveDBCommand) MarkDown() string {
+	items := []string{
+		"# RemoveDB",
+		"The ```bor removedb``` command will remove the blockchain and state databases at the given datadir location",
+		c.Flags().MarkDown(),
+	}
+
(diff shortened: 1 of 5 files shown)


internal/cli/removedb.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// create ethereum node config with just the datadir
+	nodeCfg := &node.Config{DataDir: datadir}
+
+	// Remove the full node state database
+	path := nodeCfg.ResolvePath("chaindata")
+	if common.FileExist(path) {
+		confirmAndRemoveDB(c.UI, path, "full node state database")
+	} else {
+		log.Info("Full node state database missing", "path", path)
```
JekaMas, line 84: we'd like to continue execution in that case?

internal/cli/removedb.go
```diff
@@ -0,0 +1,145 @@
...
+	// under the "chaindata" folder.
+	path = filepath.Join(nodeCfg.ResolvePath("chaindata"), "ancient")
+	if common.FileExist(path) {
+		confirmAndRemoveDB(c.UI, path, "full node ancient database")
+	} else {
+		log.Info("Full node ancient database missing", "path", path)
+	}
+
+	// Remove the light node database
+	path = nodeCfg.ResolvePath("lightchaindata")
```
JekaMas, line 99: could you define a bunch of constants for string values?

internal/cli/removedb.go
```diff
@@ -0,0 +1,145 @@
...
+			case "y":
+				start := time.Now()
+				_ = filepath.Walk(database, func(path string, info os.FileInfo, err error) error {
+					// If we're at the top level folder, recurse into
+					if path == database {
+						return nil
+					}
+					// Delete all the files, but not subfolders
+					if !info.IsDir() {
+						os.Remove(path)
```
JekaMas, line 130: do we need handle the error here?

JekaMas (approved):

pratikpatil024 (approved):

internal/cli/removedb.go
manav2401: Resolving this as per our last conversation.

cffls (approved):

PR #481, opened 2022-08-03
internal/cli, cmd/geth: replaced package naoina/toml with BurntSushi/toml and updated config name-tags

pratikpatil024:
This PR,

- Replaced the `naoina/toml` package in `dumpconfig` with `BurntSushi/toml` package.
- Updated the toml name tags for the raw field types in the config structure.
- Updated/modified tests accordingly.

[Jira Task Link](https://polygon.atlassian.net/browse/POS-687)

### DIFF
--- a/internal/cli/dumpconfig.go
+++ b/internal/cli/dumpconfig.go
@@ -1,23 +1,23 @@
 package cli
 
 import (
-	"reflect"
+	"os"
 	"strings"
 
-	"github.com/naoina/toml"
+	"github.com/BurntSushi/toml"
 
 	"github.com/ethereum/go-ethereum/internal/cli/server"
 )
 
-// These settings ensure that TOML keys use the same names as Go struct fields.
-var tomlSettings = toml.Config{
-	NormFieldName: func(rt reflect.Type, key string) string {
-		return key
-	},
-	FieldToKey: func(rt reflect.Type, field string) string {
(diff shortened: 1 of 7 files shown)


internal/cli/dumpconfig.go
```diff
@@ -69,14 +60,9 @@ func (c *DumpconfigCommand) Run(args []string) int {
 	userConfig.Gpo.IgnorePriceRaw = userConfig.Gpo.IgnorePrice.String()
 	userConfig.Cache.RejournalRaw = userConfig.Cache.Rejournal.String()
 
-	// Currently, the configurations (userConfig) is exported into `toml` file format.
-	out, err := tomlSettings.Marshal(&userConfig)
-	if err != nil {
-		c.UI.Error(err.Error())
-		return 1
+	if err := toml.NewEncoder(os.Stdout).Encode(userConfig); err != nil {
+		log.Fatal(err)
```
manav2401, line 64: Can we use `c.UI.Error(err.Error())` here instead of log.Fatal? And also return 1?

manav2401:
Let's also remove the dependency in the old-cli (in `cmd/geth/config.go`). Just remove the toml config there and use the new package's encoder instead.

pratikpatil024:
Created a new PR #482 which merges into `qa`.

PR #482, opened 2022-08-04
internal/cli, cmd/geth: replaced package naoina/toml with BurntSushi/toml and updated config name-tags

pratikpatil024:
This PR,

- Replaced the `naoina/toml` package in `dumpconfig` with `BurntSushi/toml` package.
- Updated the toml name tags for the raw field types in the config structure.
- Updated/modified tests accordingly.
- Replaced the `naoina/toml` package in `cmd/geth/config.go` with `BurntSushi/toml` package.

[Jira Task Link](https://polygon.atlassian.net/browse/POS-687)

### DIFF
--- a/cmd/geth/config.go
+++ b/cmd/geth/config.go
@@ -17,16 +17,15 @@
 package main
 
 import (
-	"bufio"
-	"errors"
 	"fmt"
+	"io/ioutil"
 	"math/big"
 	"os"
-	"reflect"
 	"time"
-	"unicode"
 
 	"gopkg.in/urfave/cli.v1"
+
+	"github.com/BurntSushi/toml"
 
 	"github.com/ethereum/go-ethereum/accounts/external"
 	"github.com/ethereum/go-ethereum/accounts/keystore"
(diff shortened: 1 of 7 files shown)


cmd/geth/config.go
```diff
@@ -61,27 +59,27 @@ var (
...
-			return nil
-		}
-		var link string
-		if unicode.IsUpper(rune(rt.Name()[0])) && rt.PkgPath() != "main" {
-			link = fmt.Sprintf(", see https://godoc.org/%s#%s for available fields", rt.PkgPath(), rt.Name())
-		}
-		return fmt.Errorf("field '%s' is not defined in %s%s", field, rt.String(), link)
-	},
-}
+// // These settings ensure that TOML keys use the same names as Go struct fields.
```
cffls, line 62: Let's keep the codebase clean. We can delete this code if it is not used anymore. If for some reason we need it in the future, we can still look at git history to restore this function.

manav2401 (approved):

cffls (approved):

PR #485, opened 2022-08-04
A thread safe way to take a free port

JekaMas:


internal/cli/server/server.go
```diff
@@ -327,22 +365,26 @@ func (s *Server) setupMetrics(config *TelemetryConfig, serviceName string) error
...
+	return s.gRPCServerByListener(lis)
+}
+
+func (s *Server) gRPCServerByListener(listener net.Listener) error {
+	s.grpcServer = grpc.NewServer(s.withLoggingUnaryInterceptor())
+	proto.RegisterBorServer(s.grpcServer, s)
+
 	go func() {
-		if err := s.grpcServer.Serve(lis); err != nil {
+		if err := s.grpcServer.Serve(listener); err != nil {
```
manav2401, line 382: Should we handle all of this outside of the go-routine synchronously?

manav2401: Because the logs becomes contradictory in case of error.

internal/cli/server/server.go
```diff
@@ -50,20 +50,40 @@ type Server struct {
...
 	}
 
 	// start the logger
 	setupLogger(config.LogLevel)
 
-	if err := srv.setupGRPCServer(config.GRPC.Addr); err != nil {
-		return nil, err
+	var err error
+
+	for _, opt := range opts {
```
manav2401, line 78: Should we handle the case where `len(opts) = 0`?

JekaMas: nope )

JekaMas: it'll skip it safely if len(opts) == 0

manav2401: But then it won't start grpc server, right?

JekaMas: if opts is empty then we just don't have to run grpc server.

internal/cli/server/server.go
JekaMas: `s.grpcServer.Serve()` will be running forever in case if everything is fine, so if we'd wait for an error then it'll block all execution.

manav2401 (approved):

marcello33 (approved):

PR #487, opened 2022-08-05
Txn prioritizer implemented using mutex map

pratikpatil024:


core/blockstm/executor.go
```diff
@@ -55,6 +59,13 @@ func ExecuteParallel(tasks []ExecTask) (lastTxIO *TxnInputOutput, err error) {
 	chTasks := make(chan ExecVersionView, len(tasks))
 	chResults := make(chan ExecResult, len(tasks))
 	chDone := make(chan bool)
+	mutMap := map[common.Address]*sync.RWMutex{}
+
+	for _, t := range tasks {
+		if _, ok := mutMap[t.Sender()]; !ok {
+			mutMap[t.Sender()] = &sync.RWMutex{}
```
JekaMas, line 66: I believe that without pointer and by value rwmutex in map still will do the thing.

pratikpatil024: @JekaMas understood, so is passing by value better in this case? Or should I keep pointer?

pratikpatil024: Also, the `TryLock` and `Unlock` method is a pointer method and if I pass by value, I cannot call these pointer method (`TryLock` and `Unlock`) on sync.RWMutex.

JekaMas: if we don't copy a mutex then it's safe to have it by value.
https://go.dev/play/p/XlKeK9vA5tH - still works fine. You can keep by pointer if you'd like to.

pratikpatil024: Right, so once I pass it by value, this is the work around, but here the problem is that we are copying lock value to `m: sync.RWMutex`, I wonder if this is safe in multiple threads?

```
m := mutMap[task.sender]
if !m.TryLock() {
	// why not this? -> chTasks <- task
	t <- task
} else {
	res := task.Execute()
	chResults <- res
	m.Unlock()
}
```

Opposed to this, as this is giving error: `cannot call pointer method TryLock on sync.RWMutex`
```
if !mutMap[task.sender].TryLock() {
	// why not this? -> chTasks <- task
	t <- task
} else {
	res := task.Execute()
	chResults <- res
	mutMap[task.sender].Unlock()
}
```

pratikpatil024: [This](https://github.com/maticnetwork/bor/blob/ed8cd53ce3370172cb23e8c6ff6b483166d6b52d/core/blockstm/executor.go#L72-L94) is is the parallel thread which I am talking about.

JekaMas: Why do you need mutex in that pieces of code at all? I don't see any reads or writes to mutMap? What's the purpose of mutMap?

JekaMas: Ok, it'll work fine then. The operations above are thread-safe.

pratikpatil024: I see, thank you. Updated the code.

JekaMas: block-stm-txn-prioritizer-1 - here are a few thoughts how it's possible to simplify loops a bit.

pratikpatil024: Thank you for this.
Mentioning @cffls to include him in the loop as well.

cffls: Thanks! Incorporated the simplification in this PR: https://github.com/maticnetwork/bor/pull/490

cffls (approved):

core/blockstm/executor.go
pratikpatil024: > I believe that without pointer and by value rwmutex in map still will do the thing.

Hi @JekaMas, actually when I do by value, I am getting these 2 lint errors

1. `Error: core/blockstm/executor.go:65:15: copylocks: assignment copies lock value to _: (sync.RWMutex, bool) contains sync.RWMutex (govet)
		if _, ok := mutMap[t.Sender()]; !ok {
`
2.  `Error: core/blockstm/executor.go:79:12: copylocks: assignment copies lock value to m: sync.RWMutex (govet)
						m := mutMap[task.sender]
 `

And these errors goes away when I change
this:
```
	mutMap := map[common.Address]sync.RWMutex{}

	for _, t := range tasks {
		if _, ok := mutMap[t.Sender()]; !ok {
			mutMap[t.Sender()] = sync.RWMutex{}
		}
	}
```
to this:
```
	mutMap := map[common.Address]*sync.RWMutex{}

	for _, t := range tasks {
		if _, ok := mutMap[t.Sender()]; !ok {
			mutMap[t.Sender()] = &sync.RWMutex{}
		}
	}
```

I searched for some other solution, but couldn't find any. 
Do you have any other solution to this?

JekaMas: it means that somewhere you pass the whole map of mutexes. In that case you either need *map or *Mutex.

pratikpatil024: I see, thanks. Using a pointer now.

PR #488, opened 2022-08-05
Milestone Implementation in the Bor

VAIBHAVJINDAL3012:


consensus/bor/heimdall/client_test.go
```diff
@@ -150,6 +211,26 @@ func TestFetchShutdown(t *testing.T) {
 		}
 	}
 
+	handler.handleFetchMilestone = func(w http.ResponseWriter, _ *http.Request) {
+		time.Sleep(6 * time.Second)
```
JekaMas, line 215: Can we get rid of sleep completely or at least make it as little as possible?

eth/handler_bor.go
```diff
@@ -39,80 +39,148 @@ var (
...
+	head := h.ethAPI.BlockNumber()
+	if head < hexutil.Uint64(checkpoint.EndBlock.Uint64()) {
+		log.Debug("Head block behind checkpoint block", "head", head, "milestone end block", checkpoint.EndBlock)
+		return blockNum, blockHash, errMissingMilestone
+	}
+
+	// verify the root hash of checkpoint
+	roothash, err := h.ethAPI.GetRootHash(ctx, checkpoint.StartBlock.Uint64(), checkpoint.EndBlock.Uint64())
+	if err != nil {
+		log.Debug("Failed to get root hash of checkpoint while whitelisting", "err", err)
```
JekaMas, line 102: could you describe the scenario when Bor node is behind the latest checkpoint?

eth/handler_bor.go
```diff
@@ -39,80 +39,148 @@ var (
...
+	if roothash != checkpoint.RootHash.String()[2:] {
+		log.Warn("Checkpoint root hash mismatch while whitelisting", "expected", checkpoint.RootHash.String()[2:], "got", roothash)
+		return blockNum, blockHash, errCheckpointRootHashMismatch
+	}
+
+	// fetch the end checkpoint block hash
+	block, err := h.ethAPI.GetBlockByNumber(ctx, rpc.BlockNumber(checkpoint.EndBlock.Uint64()), false)
+	if err != nil {
+		log.Debug("Failed to get end block hash of checkpoint while whitelisting", "err", err)
+		return blockNum, blockHash, errEndBlock
```
JekaMas, line 115: why don't what an error and remove logging on that level?

eth/handler_bor.go
```diff
@@ -39,80 +39,148 @@ var (
...
+	}
+
+	hash := fmt.Sprintf("%v", block["hash"])
```
JekaMas, line 118: is it equal to block["hash"].String()?

consensus/bor/heimdall/client_test.go
VAIBHAVJINDAL3012: Okay! I will reduce the time.

eth/handler_bor.go
```diff
@@ -39,80 +39,148 @@ var (
...
+		log.Debug("Head block behind milestone block", "head", head, "milestone end block", milestone.EndBlock)
+		return blockNum, blockHash, errMissingMilestone
 	}
 
-	// If we're in the first iteration, we'll fetch last 10 checkpoints, else only the latest one
-	iterations := 1
-	if first {
-		iterations = 10
+	// verify the root hash of checkpoint
+	roothash, err := h.ethAPI.GetRootHash(ctx, milestone.StartBlock.Uint64(), milestone.EndBlock.Uint64())
```
JekaMas, line 162: it looks like code duplication, doesn't it?

eth/handler_bor.go
VAIBHAVJINDAL3012: Yes, When Bor node is not in sync or on the wrong fork, but heimdall is in complete sync. At that point this case will arise.

eth/handler_bor.go
```diff
@@ -39,80 +39,148 @@ var (
...
+
+	blockNum = checkpoint.EndBlock.Uint64()
+	blockHash = common.HexToHash(hash)
+
+	return blockNum, blockHash, nil
+}
+
+// fetchWhitelistMilestones fetches the latest milestone/s from it's local heimdall
+// and verifies the data against bor data.
+func (h *ethHandler) fetchWhitelistMilestone(ctx context.Context, bor *bor.Bor, first bool) (uint64, common.Hash, error) {
```
JekaMas, line 128: could we have a bunch of unit tests for milestone feature? could you ask @manav2401 about how he tested checkpoints?

VAIBHAVJINDAL3012: I have included the similar test for the milestone as was done in the Checkpoint. But will add more after the manual testing.

eth/handler_bor.go
VAIBHAVJINDAL3012: Yes

eth/handler_bor.go
VAIBHAVJINDAL3012: Oh I forgot to change the comment. Will do it. Actually we can make the single function also but would need to give sort of flag to it. What will be best way- To keep the single function with flag or Two seperate functions?

eth/handler_bor.go
VAIBHAVJINDAL3012: Do you mean we should return the err itself rather than creating errEndBlock

eth/handler_bor.go
JekaMas: Could you stick with solutions that doesn't introduce additional calls like fmt?

eth/handler_bor.go
JekaMas: Like add the context to the error with fmt Errorf + %w

eth/handler_bor.go
JekaMas: Could retrying be  a separate function? I'd like to have a general solution without flags

eth/handler_bor.go
JekaMas: I am looking for more units mostly.

eth/backend.go
```diff
@@ -641,16 +643,53 @@ var (
...
 // checkpoint whitelist map.
 func (s *Ethereum) startCheckpointWhitelistService() {
+	// first run the checkpoint whitelist
+	firstCtx, cancel := context.WithTimeout(context.Background(), whitelistTimeout)
+	err := s.handleWhitelistCheckpoint(firstCtx)
+
+	cancel()
+
+	if err != nil {
+		if errors.Is(err, ErrBorConsensusWithoutHeimdall) || errors.Is(err, ErrNotBorConsensus) {
```
JekaMas, line 653: I see this pattern quite often. Could we try to introduce in `common` package something like

```
func AnyError(src error, errs ...error) bool {
	for _, err := range errs {
		if errors.Is(src, err) {
			return true
		}
	}

	return false
}
``` 
?

eth/backend.go
```diff
@@ -695,21 +734,47 @@ func (s *Ethereum) handleWhitelistCheckpoint(ctx context.Context, first bool) er
...
+// handleMilestone handles the milestone mechanism.
+func (s *Ethereum) handleMilestone(ctx context.Context) error {
+	ethHandler := (*ethHandler)(s.handler)
+
+	bor, ok := ethHandler.chain.Engine().(*bor.Bor)
+	if !ok {
+		return ErrNotBorConsensus
+	}
+
+	if bor.HeimdallClient == nil {
```
JekaMas, line 761: I believe we have something like `IsHeimdall` or something like that.

eth/bor_checkpoint_verifier.go
```diff
@@ -2,59 +2,120 @@ package eth
...
+			log.Warn("Root hash mismatch while whitelisting", "expected", localRoothash, "got", rootHash)
+
+			ethHandler := (*ethHandler)(eth.handler)
+
+			var (
+				rewindTo uint64
+				doExist  bool
+			)
+
+			if doExist, rewindTo, _ = ethHandler.downloader.GetWhitelistedMilestone(); doExist {
```
JekaMas, line 67: I believe this part should be extracted as a function with a batch of units on it.

eth/backend.go
VAIBHAVJINDAL3012: No, We don't have IsHeimdall function linked to Bor struct.

eth/downloader/whitelist/checkpoint.go
```diff
@@ -0,0 +1,145 @@
...
+		// We don't have whitelisted checkpoint, no additional validation will be possible
+		return true
+	}
+
+	// Return if we've received empty chain
+	if len(chain) == 0 {
+		return false
+	}
+
+	w.m.Lock()
```
JekaMas, line 82: What fields we're guarding?

VAIBHAVJINDAL3012: w.checkpointNumber, It might be possible that when we are reading this,  other routine might be writing to it.

eth/downloader/whitelist/milestone.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type milestone struct {
```
JekaMas, line 12: is it the same structure as checkoint? if so please check could we use alias or embedding here?

```
// alias
type milestone checkpoint

// embedding
type milestone struct {
    checkpoint
}

JekaMas: and remove most of copy-paste below

VAIBHAVJINDAL3012: Yes, Same structure. I will check the both approach,

eth/bor_checkpoint_verifier.go
VAIBHAVJINDAL3012: Do you mean to put this if else if else in one function? And use that function here?

eth/downloader/whitelist/milestone.go
VAIBHAVJINDAL3012: To implement this approach, we need to change the name of all the functions and variable we have used for milestone.

PR #490, opened 2022-08-08
Block-stm optimization

cffls:
Added tests for executor and some improvements:

    Added tests for executor and some improvements:

    1. Add a dependency map during execution. This will prevent aborted tasks from being sent for execution immedaitely after failure.
    2. Change the key of MVHashMap from string to a byte array. This will reduce time to convert byte slices to strings.
    3. Use sync.Map to reduce the time spent in global mutex.
    4. Skip applying intermediate states.
    5. Estimate dependency when an execution fails without dependency information.
    6. Divide execution task queue into two separate queues. One for relatively certain transactions, and the other for speculative future transactions.
    7. Setting dependencies of Txs coming from the same sender before starting parallel execution.
    8. Process results in their semantic order (transaction index) instead of the order when they arrive. Replace result channel with a priority queue.

Sample test output:

```
Running tool: /home/fan/.go/bin/go test -timeout 30s -run ^TestAlternatingTx$ github.com/ethereum/go-ethereum/core/blockstm

=== RUN   TestAlternatingTx
=== PAUSE TestAlternatingTx
=== CONT  TestAlternatingTx
INFO [09-01|20:55:12.860] Executing block                          numTx=200 numRead=20 numWrite=20 numNonIO=100
INFO [09-01|20:55:12.905] blockstm exec summary                    execs=200 success=200 aborts=0 validations=200 failures=0 #tasks/#execs=100.00%
exec duration 44.722667ms, serial duration 75.476825ms, time reduced 30.754158ms 40.75%, ✅
Improved:  1 Total:  1 success rate:  100
Total exec duration: 44.722667ms, total serial duration: 75.476825ms, time reduced: 30.754158ms, time reduced percent: 40.75%
--- PASS: TestAlternatingTx (0.05s)
PASS
ok      github.com/ethereum/go-ethereum/core/blockstm   0.050s
```

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -185,8 +185,6 @@ func (s *StateDB) MVWriteList() []blockstm.WriteDescriptor {
 	for _, v := range s.writeMap {
 		if !v.Path.IsAddress() {
 			writes = append(writes, v)
-		} else if _, ok := s.newStateObjects[common.BytesToAddress(v.Path[:common.AddressLength])]; ok {
-			writes = append(writes, v)
 		}
 	}
 
--- a/core/state/statedb_test.go
+++ b/core/state/statedb_test.go
@@ -885,7 +885,7 @@ func TestApplyMVWriteSet(t *testing.T) {
 	states[0].SetBalance(addr1, balance1)
 	states[0].SetState(addr2, key2, val2)
 	states[0].GetOrNewStateObject(addr3)
-	states[0].Finalise(false)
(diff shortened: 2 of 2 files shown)


pratikpatil024 (approved):

JekaMas:
https://github.com/ethereum/go-ethereum/issues/21350

It seems like any randomization might cause quite a few problems.
Need to rethink whole txpool ordering.

PR #494, opened 2022-08-17
eth, cli: prevent snap sync mode migration - v0.3.x

manav2401:
We've seen a couple of scenarios where some of our internal nodes have migrated to snap sync mode. Until we make it working fully, this PR will prevent that from happening. 

It adds a preventive layer in the cli (old and new both) so that it's not enabled through flag and also handles cases where is migrates to snap sync.


JekaMas (approved):

temaniarpit27 (approved):

temaniarpit27:
@manav2401 create a new PR from v0.2.16 and raise it against that as well

pratikpatil024 (approved):

PR #496, opened 2022-08-17
eth: fix whitelist/requiredblocks regression - v0.3.x

manav2401:
This PR fixes a bug in checking blocks (and hash) set using the `--requiredblocks` flag in the new cli. 

Reference: https://github.com/ethereum/go-ethereum/pull/24817

Jira - [[v0.2.16 and v0.3.0] Required block mismatch when whitelist blocks are provided](https://polygon.atlassian.net/browse/POS-726)


temaniarpit27 (approved):

JekaMas (approved):

JekaMas:
How do you think that kind of test do we need to catch that issue?

PR #501, opened 2022-08-23
shuffle pending accounts in the transaction pool

JekaMas:
Prevents possible vulnerability with pre-mined accounts


ssandeep (approved):
Should we also propose this change in geth?

temaniarpit27 (approved):

temaniarpit27:
@JekaMas can you please look into the CI failure?

common/random/shuffle.go
```diff
@@ -0,0 +1,28 @@
...
+	//nolint:gosec
+	r := rand.New(CryptoSource{}) //CryptoSource is a cryptographic source of randomness
+
+	r.Shuffle(len(slice), func(i, j int) {
+		slice[i], slice[j] = slice[j], slice[i]
+	})
+}
+
+// ShuffleMap returns a shuffled slice of keys to iterate by
+func ShuffleMap[K comparable, V any](m map[K]V) []K {
```
cffls, line 17: This isn't deterministically random I guess?

JekaMas: As far as it uses crypto/rand, it isn't.

cffls: This looks like a good explanation: https://stackoverflow.com/a/55925880

cffls (approved):

common/random/shuffle.go
JekaMas: Yeap, although it only depends on non cryptographically strong hash and a seed. The PR introduces cryptographically strong solution.

0xsharma:
@JekaMas Can you please have a look at the CI ? Thank you !

temaniarpit27:
@JekaMas CI is failing. Can you please look into this?

JekaMas:
I believe, we need to rethink whole idea of random ordering of transactions - https://github.com/ethereum/go-ethereum/issues/21350

JekaMas:
Will redo with proving tests.

PR #502, opened 2022-08-23
Changing default txpool pricelimit to 1

0xsharma:
Prior to this PR, 
Bor client running on ^v0.3.0 were setting 30Gwei  as default txpool.pricelimit which was not suitable for mumbai testnet. 

In this PR, the bor client will set 1 wei as default txpool.pricelimit.

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -425,7 +425,7 @@ func DefaultConfig() *Config {
 			NoLocals:     false,
 			Journal:      "",
 			Rejournal:    1 * time.Hour,
-			PriceLimit:   30000000000,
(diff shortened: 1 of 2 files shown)


pratikpatil024 (approved):

temaniarpit27 (approved):

JekaMas (approved):

pratikpatil024 (approved):

cffls (commented):
I think we should configure this parameter through the CLI flag or a toml config on Mumbai nodes. We want to default code to work as mainnet sentry, so this change will probably break it.

cffls (approved):

PR #503, opened 2022-08-24
internal/cli/server: update default cli values

0xsharma:
This PR sets some of the default flags in the new-cli referencing from bor v0.2.16 defaults. 

Some of the flags changed are
- txpool.journal
- miner.gaslimit
- miner.gasprice (It was set to 30 gwei which is not ideal for testnets. It is also later translated to txpool.pricelimit on a miner node and hence txpool is not able get transactions from other nodes as well)
- cors/vhost configurations for http, ws, graphql
- disable-bor-wallet (This change was already done in the release qa branch but was not reflected in develop).


manav2401 (approved):

manav2401:
Should we also change [here](https://github.com/maticnetwork/bor/blob/shivam/sealer-gasPrice/builder/files/config.toml#L55) and in launch repo?

temaniarpit27 (approved):

temaniarpit27:
> Should we also change [here](https://github.com/maticnetwork/bor/blob/shivam/sealer-gasPrice/builder/files/config.toml#L55) and in launch repo?

isn't this deprecated?

pratikpatil024 (approved):

temaniarpit27 (approved):

manav2401 (approved):

PR #507, opened 2022-08-31
add : TestValidatorWentOffline

0xsharma:
In this PR, we add a red test where we test a validator coming back online after a crash and reorging their backup blocks generated by the secondary validator.


tests/bor/bor_reorg_test.go
```diff
@@ -0,0 +1,303 @@
...
+
+var (
+	// addr1 = 0x71562b71999873DB5b286dF957af199Ec94617F7
+	pkey1, _ = crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
+	// addr2 = 0x9fB29AAc15b9A4B7F17c3385939b007540f4d791
+	pkey2, _ = crypto.HexToECDSA("9b28f36fbd67381120752d6172ecdcf10e06ab2d9a1367aac00cdcd6ac7855d3")
+	keys     = []*ecdsa.PrivateKey{pkey1, pkey2}
+)
+
+func initMiner(genesis *core.Genesis, privKey *ecdsa.PrivateKey) (*node.Node, *eth.Ethereum, error) {
```
JekaMas, line 41: Why you didn't reuse code from tests with NewFakeBor() ?

tests/bor/bor_reorg_test.go
```diff
@@ -0,0 +1,303 @@
...
+	go func() {
+		for {
+			// Stop when interrupted.
+			select {
+			case <-interruptCh:
+				for _, node := range stacks {
+					node.Close()
+				}
+				return
+			default:
```
JekaMas, line 183: It'll consume all CPU

tests/bor/bor_reorg_test.go
0xsharma: Because, I had to use a Node object and Ethereum object so that I can interact on peer level and can also use account manager to seal blocks using those addresses. NewFakeBor returns only consensus.Engine.

tests/bor/bor_reorg_test.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+
+		// reorg happens here, node1 has higher difficulty, it will replace blocks by node0
+
+		if blockHeaderVal0.Number.Uint64() == 30 {
+
+			break
+		}
+
+		time.Sleep(1 * time.Second)
```
JekaMas, line 215: could we go without sleeps?

tests/bor/bor_reorg_test.go
0xsharma: Will rectify this. Thank you !

tests/bor/bor_reorg_test.go
0xsharma: Yes, we can.

JekaMas (approved):

tests/bor/bor_reorg_test.go
JekaMas: So the only problem is to have manageable fake transport layer. Will do one in a new task.

Raneet10 (approved):

tests/bor/bor_reorg_test.go
```diff
@@ -0,0 +1,282 @@
...
+	fdlimit.Raise(2048)
+
+	// Generate a batch of accounts to seal and fund with
+	faucets := make([]*ecdsa.PrivateKey, 128)
+	for i := 0; i < len(faucets); i++ {
+		faucets[i], _ = crypto.GenerateKey()
+	}
+
+	// Pre-generate the ethash mining DAG so we don't race
+	ethash.MakeDataset(1, ethconfig.Defaults.Ethash.DatasetDir)
```
manav2401, line 130: Do we need this?

manav2401 (approved):

tests/bor/bor_reorg_test.go
0xsharma: Actually yes this can be removed from here.

manav2401 (approved):

JekaMas (approved):

PR #520, opened 2022-09-22
Refactor blockstm executor

cffls:
Refactor blockstm executor into a structure that can execute transactions step by step. This enables us to write property based tests for the executor.


pratikpatil024 (commented):
Why does it show that the complete file is added? (`dag.go`, `executor.go`, ....)

PR #529, opened 2022-09-23
internal/cli/server: fix flag behaviour

manav2401:
This PR achieves the following things

1. When the flag `nodiscover` was enabled in the old-cli, the node would not discover other peers but still connected to static and trusted peers if required, which was much convenient for our validator and sentry architecture. But, in the new-cli, when we set this flag, we also used to set `maxpeers` to 0. This would not allow the node to connect to any static peers. Ideally, we've suggested people to use [this](https://github.com/maticnetwork/launch/blob/master/mainnet-v1/sentry/validator/bor/start.sh#L39) setting of flags (`nodiscover` along with `maxpeers=0`). Now, we can either get rid of this flag and keep the same behaviour in client as it is now by not translating it the new toml file using @pratikpatil024's migration script or we can keep the same behaviour of `nodiscover` in the client (as geth does). I prefer the latter one. 

2. We were unable to send metrics to datadog from some of our internal validator nodes running the new-cli. On investigating further, we found out that in the old-cli, prometheus address was set on based on `metrics.addr` and `metrics.port` flags (or `pprof.addr` and `pprof.port` flag if the earlier ones are not available). The defaults address in any case is `127.0.0.1:7071`. So, ideally even if the users have not provided these flags, the prometheus metrics are exported on the default address. In case of the new-cli, currently we need to explicitly set the address for prometheus. Instead we can set them as default so that the changes on our end (w.r.t to metrics collector) are minimized. Also, if these flags are available, use them to derive prometheus address in the migration script. 

3. As finalized in (2), the default port for prometheus is 7071 and with the migration script, we're translating the pprof address + port combination to grpc address. Most of the nodes have pprof port set to 7071 and hence it would not allow prometheus to work. My suggestion is to ignore all pprof flags and just rely on default grpc port which is 3131.

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -996,8 +996,7 @@ func (c *Config) buildNode() (*node.Config, error) {
 	}
 
 	if c.P2P.NoDiscover {
-		// Disable networking, for now, we will not even allow incomming connections
-		cfg.P2P.MaxPeers = 0
+		// Disable peer discovery
 		cfg.P2P.NoDiscovery = true
 	}
 
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -479,8 +479,8 @@ func DefaultConfig() *Config {
 		Telemetry: &TelemetryConfig{
 			Enabled:               false,
 			Expensive:             false,
-			PrometheusAddr:        "",
-			OpenCollectorEndpoint: "",
+			PrometheusAddr:        "0.0.0.0:7071",
+			OpenCollectorEndpoint: "0.0.0.0:4317",
 			InfluxDB: &InfluxDBConfig{
 				V1Enabled:    false,
 				Endpoint:     "",
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -479,8 +479,8 @@ func DefaultConfig() *Config {
 		Telemetry: &TelemetryConfig{
 			Enabled:               false,
 			Expensive:             false,
-			PrometheusAddr:        "0.0.0.0:7071",
-			OpenCollectorEndpoint: "0.0.0.0:4317",
+			PrometheusAddr:        "127.0.0.1:7071",
+			OpenCollectorEndpoint: "127.0.0.1:4317",
 			InfluxDB: &InfluxDBConfig{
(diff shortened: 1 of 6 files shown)


scripts/getconfig.go
```diff
@@ -658,7 +659,7 @@ func main() {
 	outOfDateFlags := checkFlag(allFlags, flagsToCheck)
 	updatedArgs := updateArgsClean(args, outOfDateFlags)
 	updatedArgs = updateArgsAdd(updatedArgs)
-	updatedArgs = handleGRPC(args, updatedArgs)
+	// updatedArgs = handleGRPC(args, updatedArgs)
```
JekaMas, line 662: why we'd like to comment out gRPC?

manav2401: Well this is the migration script which when provided with an old start.sh spits out a new cli based toml config file. Earlier we used to set `grpc.addr = pprof.addr + ":" + pprof.port" as the grpc address. But now as that port is taken by default prometheus config, we would have to rely on a different port. I suggest relying on the default port (3131) instead of anything else. 

Also, it would be difficult to change prometheus port as all of our nodes have service running on that port.

cffls (approved):

pratikpatil024 (approved):

pratikpatil024 (approved):

cffls (approved):
LGTM.

internal/cli/server/config.go
```diff
@@ -996,8 +996,7 @@ func (c *Config) buildNode() (*node.Config, error) {
 	}
 
 	if c.P2P.NoDiscover {
-		// Disable networking, for now, we will not even allow incomming connections
-		cfg.P2P.MaxPeers = 0
+		// Disable peer discovery
 		cfg.P2P.NoDiscovery = true
```
VAIBHAVJINDAL3012, line 1000: What is the difference in outcome in MaxPeers=0 and NoDiscovery=true?

temaniarpit27: max peers 0 will not let anyone to connect with it and it wont connect to everyone
no discovery will not let anyone recognize the node but the node can connect to others
good for validators as we don't want any unknown nodes to connect to it even if somehow they get our node details

PR #530, opened 2022-09-26
Fixed a small bug in the Report function

pratikpatil024:
We only used to sort the roots, we should also modify the rootIds in the same order.

### DIFF
--- a/core/blockstm/dag.go
+++ b/core/blockstm/dag.go
@@ -74,13 +74,19 @@ func BuildDAG(deps TxnInputOutput) (d DAG) {
 func (d DAG) Report(out func(string)) {
 	roots := make([]int, 0)
 	rootIds := make([]string, 0)
+	rootIdMap := make(map[int]string, len(d.GetRoots()))
 
 	for k, i := range d.GetRoots() {
 		roots = append(roots, i.(int))
-		rootIds = append(rootIds, k)
+		rootIdMap[i.(int)] = k
 	}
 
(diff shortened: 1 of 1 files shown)


core/blockstm/dag.go
```diff
@@ -80,7 +100,7 @@ func (d DAG) Report(out func(string)) {
 		rootIds = append(rootIds, k)
 	}
 
-	sort.Ints(roots)
+	roots, rootIds = bubbleSort(roots, rootIds)
```
cffls, line 103: Not sure if this is the best way to sort this. Can we create a map (roots -> rootIds), sort roots, and re-create a sorted rootId slace based on the sorted roots and the map?

pratikpatil024: Yes sure, will do this.

core/blockstm/dag.go
```diff
@@ -74,13 +74,19 @@ func BuildDAG(deps TxnInputOutput) (d DAG) {
 func (d DAG) Report(out func(string)) {
 	roots := make([]int, 0)
 	rootIds := make([]string, 0)
+	rootIdMap := make(map[int]string, 0)
```
cffls, line 77: Maybe consider preallocating the map with the size of `len(d.GetRoots())`?

cffls (approved):

core/blockstm/dag.go
pratikpatil024: Done

PR #547, opened 2022-10-12
Add state sync transaction to debug_trace functions.

0xsharma:
# Description

Because state sync transaction is added by bor consensus, it will not be included in `debug_traceByBlockNumber` and `debug_tranceByBlockHash`. The problem seems to be that block.Transactions() doesn’t contain state sync transaction.

In this PR, tracing can be possible on state-sync transactions as well. All the debug_trace functions will contain the traces of state-sync transactions, if present.

# Changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests
- [x] I have tested this code manually on mainnet

### DIFF
--- a/core/vm/interface.go
+++ b/core/vm/interface.go
@@ -74,6 +74,8 @@ type StateDB interface {
 	AddPreimage(common.Hash, []byte)
 
 	ForEachStorage(common.Address, func(common.Hash, common.Hash) bool) error
+
+	Finalise(bool)
 }
 
 // CallContext provides a basic interface for the EVM calling conventions. The EVM
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -28,9 +28,11 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/consensus"
+	"github.com/ethereum/go-ethereum/consensus/bor/statefull"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/state"
@@ -80,6 +82,9 @@ type Backend interface {
 	// so this method should be called with the parent.
 	StateAtBlock(ctx context.Context, block *types.Block, reexec uint64, base *state.StateDB, checkLive, preferDisk bool) (*state.StateDB, error)
 	StateAtTransaction(ctx context.Context, block *types.Block, txIndex int, reexec uint64) (core.Message, vm.BlockContext, *state.StateDB, error)
+
+	// Bor related APIs
+	GetBorBlockTransactionWithBlockHash(ctx context.Context, txHash common.Hash, blockHash common.Hash) (*types.Transaction, common.Hash, uint64, uint64, error)
 }
 
 // API is the collection of tracing APIs exposed over the private debugging endpoint.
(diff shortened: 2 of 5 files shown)


pratikpatil024 (approved):

core/vm/interface.go
```diff
@@ -74,6 +74,8 @@ type StateDB interface {
 	AddPreimage(common.Hash, []byte)
 
 	ForEachStorage(common.Address, func(common.Hash, common.Hash) bool) error
+
+	Finalise(bool)
```
cffls, line 78: Why is this needed? Changes in this commit only use `Finalise` from statedb right?

eth/tracers/api.go
```diff
@@ -705,7 +795,9 @@ func (api *API) standardTraceBlockToFile(ctx context.Context, block *types.Block
 			canon = false
 		}
 	}
-	for i, tx := range block.Transactions() {
+
+	txs, stateSyncPresent := api.getAllBlockTransactions(ctx, block)
```
cffls, line 799: As discussed, we don't want to change the current behavior, which won't include state sync transaction. Could we introduce a new flag to trace API and only include state sync transactions when the flag is set to true?

eth/tracers/api_bor.go
```diff
@@ -115,7 +124,11 @@ func (api *API) traceBorBlock(ctx context.Context, block *types.Block, config *T
 	}
 
 	for indx, tx := range txs {
-		res.Transactions = append(res.Transactions, traceTxn(indx, tx))
+		if stateSyncPresent && indx == len(txs)-1 {
```
cffls, line 127: Similar to the comment above, by default, we don't want to include state sync tx even when it presents in the block.

eth/tracers/api.go
0xsharma: Done

eth/tracers/api_bor.go
0xsharma: Done

core/vm/interface.go
0xsharma: We need to use Finalise method of the current vmenv state which is possible only if this is enabled. You can try removing this in code and will get the reason why this was added. We can discuss this offline if needed ✅

0xsharma:
> debug.traceBlockByNumber("0x20FE900",{"borTraceEnabled": true}).length
203
> debug.traceBlockByNumber("0x20FE900",{"borTraceEnabled": false}).length
202
> debug.traceBlockByNumber("0x20FE900").length
202

We can now specify inside the rpc call, if we want the trace of stateSyncTx as well. By default, it is false. This works for all debug trace methods.

pratikpatil024 (approved):

eth/tracers/api.go
```diff
@@ -911,9 +1126,22 @@ func (api *API) traceTx(ctx context.Context, message core.Message, txctx *Contex
 	// Call Prepare to clear out the statedb access list
 	statedb.Prepare(txctx.TxHash, txctx.TxIndex)
 
-	result, err := core.ApplyMessage(vmenv, message, new(core.GasPool).AddGas(message.Gas()))
-	if err != nil {
-		return nil, fmt.Errorf("tracing failed: %w", err)
+	var result *core.ExecutionResult
+
+	if config.BorTx == nil {
+		config.BorTx = newBoolPtr(false)
```
cffls, line 1132: Maybe I am missing something, why do we need `newBoolPtr`?

0xsharma: Hey, config.BorTx is a *bool and we cannot define it directly AFAIK. So instead of using 3 same lines at multple places. I created a func to give *bool from bool.

0xsharma: Yes, I too actually was going with `bool`, but then also wanted to keep the same standard as before, therefore used `*bool`.

cffls (approved):

MRabenda:
@0xsharma @pratikspatil024 should state-sync traces be also available in trace_block?

0xsharma:
@MRabenda , yes they will be available but as empty traces as state-syncs are system transactions.

PR #556, opened 2022-10-18
interrupting commit experiment

JekaMas:
I am experimenting with interrupting and committing current being mined block using an empty block as a signal to that (we also should remove committing this empty block complitely). The idea is to have partly filled blocks rather than empty ones.


ferranbt (approved):

miner/worker.go
```diff
@@ -1293,6 +1326,33 @@ func (w *worker) commitWork(ctx context.Context, interrupt *int32, noempty bool,
...
+
+	blockNumber := current.NumberU64() + 1
+	interruptCh := make(chan struct{})
+
+	go func() {
+		select {
+		case <-timeoutTimer.C:
+			log.Info("an interrupt event. cancelling the current block",
+				"block", blockNumber,
+				"hash", current.Hash().Hex(),
```
b00ris, line 1345: Do you expect to have block number from the next block, but block number from the previous one?

miner/worker.go
```diff
@@ -911,14 +916,22 @@ func (w *worker) commitTransaction(env *environment, tx *types.Transaction) ([]*
...
 	if env.gasPool == nil {
 		env.gasPool = new(core.GasPool).AddGas(gasLimit)
 	}
 	var coalescedLogs []*types.Log
 
 	for {
+		// case of interrupting by timeout
+		select {
+		case <-interruptCh:
+			break
```
b00ris, line 931: Do you need to do something with interrupt variable here?

miner/worker.go
```diff
@@ -1259,6 +1281,17 @@ func (w *worker) commitWork(ctx context.Context, interrupt *int32, noempty bool,
...
 	}
 
+	var interruptCh chan struct{}
+
+	stopFn := func() {}
+	defer func() {
+		stopFn()
+	}()
+
+	if !noempty {
```
b00ris, line 1291: You'll exit earlier

marcello33 (approved):

0xsharma (approved):

PR #561, opened 2022-10-20
Block stm miner dependency

pratikpatil024:
1. Miner calculates the transaction dependencies while creating a block, and puts them in the block header.
2. Updated the parallel executor to support this, and use these dependencies.
3. Added tests in `executor_test` to benchmark these changes.
4. Updated the `applyTransaction` and `commitTransactions` functions to support MVHashMap.


core/state/statedb.go
```diff
@@ -227,6 +227,18 @@ func (s *StateDB) ensureWriteMap() {
...
+	s.readMap = make(map[blockstm.Key]blockstm.ReadDescriptor)
+}
+
+func (s *StateDB) ClearWriteMap() {
+	s.writeMap = make(map[blockstm.Key]blockstm.WriteDescriptor)
+}
+
+func (s *StateDB) SetMVHashMapNil() {
+	s.mvHashmap = nil
+}
```
cffls, line 240: Can we just call `SetMVHashmap(nil)` instead of creating a new function?

core/blockstm/executor_test.go
```diff
@@ -294,6 +306,66 @@ func testExecutorComb(t *testing.T, totalTxs []int, numReads []int, numWrites []
 	fmt.Printf("Total exec duration: %v, total serial duration: %v, time reduced: %v, time reduced percent: %.2f%%\n", totalExecDuration, totalSerialDuration, totalSerialDuration-totalExecDuration, float64(totalSerialDuration-totalExecDuration)/float64(totalSerialDuration)*100)
 }
 
+// // nolint: gocognit
```
cffls, line 309: Let's try not to leave large chunk of commented code. Are we supposed to uncomment these lines?

core/parallel_state_processor.go
```diff
@@ -171,9 +182,9 @@ func (task *ExecutionTask) Settle() {
 
 	task.finalStateDB.Prepare(task.tx.Hash(), task.index)
 
-	coinbase, _ := task.blockChain.Engine().Author(task.header)
+	// coinbase, _ := task.blockChain.Engine().Author(task.header)
```
cffls, line 185: Let's remove this line.

core/parallel_state_processor.go
```diff
@@ -372,3 +426,46 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 
 	return receipts, allLogs, *usedGas, nil
 }
+
+func GetDepListForTx(txDependency [][]uint64, txIdx int) []int {
```
cffls, line 430: Let's remove this if not used.

core/parallel_state_processor.go
```diff
@@ -372,3 +426,46 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
...
+	}
+
+	for i := 0; i < len(txDependency[tempIdx]); i++ {
+		tempArr = append(tempArr, int(txDependency[tempIdx][i]))
+	}
+
+	return tempArr
+}
+
+// Jerry's function
```
cffls, line 452: Remove this line :D

pratikpatil024:
> How can others check if the miner provides the wrong dependencies in the block header

@516108736 thank you for asking. We already have some ideas, but it is not part of the scope of this PR.

core/blockstm/executor_test.go
cffls: Is there a way to make them run faster, e.g. reduce the number Txs?

pratikpatil024: Yes, that will work. Also, if it doesn't affect other tests, we can increase the go-test timeout in the CI.

core/blockstm/dag.go
```diff
@@ -69,6 +75,58 @@ func BuildDAG(deps TxnInputOutput) (d DAG) {
...
+
+				for k := range newDependencies[i] {
+					_, boool1 := newDependencies[j][k]
+
+					if boool1 {
+						delete(newDependencies[i], k)
+					}
+				}
+			}
+		}
```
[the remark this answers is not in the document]
pratikpatil024: So I also tried doing something like this:
```
func depsHelper(dependencies map[int]map[int]bool, allOutputs interface{}, txTo TxnInput, i int, interfaceFlag int) map[int]map[int]bool {
	var allOutputs1 interface{}

	if interfaceFlag == 1 {
		allOutputs1 = allOutputs.([][]WriteDescriptor)
	} else if interfaceFlag == 2 {
		allOutputs1 = allOutputs.([]TxnOutput)
	}

	for j := 0; j <= i-1; j++ {
		txFrom := allOutputs1[j]

		if HasReadDep(txFrom, txTo) {
			dependencies[i][j] = true

			for k := range dependencies[i] {
				_, foundDep := dependencies[j][k]

				if foundDep {
					delete(dependencies[i], k)
				}
			}
		}
	}

	return dependencies
}

func UpdateDeps(deps map[int]map[int]bool, t TxDep) map[int]map[int]bool {
	txTo := t.ReadList

	deps[t.Index] = map[int]bool{}

	deps = depsHelper(deps, t.FullWriteList, txTo, t.Index, 1)

	return deps
}

func GetDep(deps TxnInputOutput) map[int]map[int]bool {
	newDependencies := map[int]map[int]bool{}

	for i := 1; i < len(deps.inputs); i++ {
		txTo := deps.inputs[i]

		newDependencies[i] = map[int]bool{}

		newDependencies = depsHelper(newDependencies, deps.allOutputs, txTo, i, 2)
	}

	return newDependencies
}

```

But because of the different type of `deps.allOutputs` and `t.Index`, I cannot do that.

core/blockstm/executor_test.go
pratikpatil024: Reduced the number of txns.

core/parallel_state_processor.go
```diff
@@ -280,6 +290,14 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 
 	coinbase, _ := p.bc.Engine().Author(header)
 
+	deps, delayMap := GetDeps(block.Header().TxDependency)
```
cffls, line 293: Can we have some unit tests for block processing with dependencies? The unit tests will allow us to easily test changes in this file without re-syncing a chain.

miner/worker.go
```diff
@@ -999,6 +1063,54 @@ func (w *worker) commitTransactions(env *environment, txs *types.TransactionsByP
...
 		}
+
+		if EnableMVHashMap {
+			env.state.ClearReadMap()
+			env.state.ClearWriteMap()
+		}
+	}
+
+	// nolint:nestif
+	if EnableMVHashMap {
```
cffls, line 1074: Same as previous comment, can we have some unit test for when mvhashmap is enabled?

pratikpatil024:
> > > How can others check if the miner provides the wrong dependencies in the block header
> > 
> > 
> > @516108736 thank you for asking. We already have some ideas, but it is not part of the scope of this PR.
> 
> Thanks for your answer, I would like to ask two more questions 1: Where can I find your ideas to this problem？ 2: Does the code of the [block-stm](https://github.com/maticnetwork/bor/tree/block-stm) branch solve this problem?

Hi @516108736, there are a few ways to check the dependencies provided by the block producer.
One of them is to check for any circular dependency (for example transaction 0 depending on transaction 1, and transaction 1 depending on transaction 0). Also, the transaction validation will state if the dependencies are correct or not.
Thanks.

cffls (approved):
LGTM. Thank you!

PR #565, opened 2022-10-20
metrics: handle values from config file

manav2401:
# Description

The metrics package uses global variables to enable metrics collection (with the flag `--metrics` and `--metrics.expensive`). As we're moving towards a toml config based approach, we found out that the values of these 2 parameters from toml file weren't being honoured in some cases. On further investigation we found out that due to the call to `init()` (in metrics/metrics.go), OS arguments were being read and metrics were being set based on those arguments. 

Now, there are some metrics which needs to be registered as soon the bor process starts. We were setting these global values but at much later point in the code when we're done setting up the node related processes which caused all the metric (meters, counters, gauges, etc) registrations to fail and hence leading to 0 values in prometheus. If we were to start bor using cli, it would work as it finds the flags in OS args but not in case of toml. 

This PR fixes this by reading the config file itself in the `init()` function of metrics package and enabling/disabling the global variables before any other process starts. 

# Changes

# Checklist

## Testing

- [x] I have tested this code manually on mumbai/mainnet

### DIFF
--- a/internal/cli/server/server.go
+++ b/internal/cli/server/server.go
@@ -253,6 +253,12 @@ func (s *Server) Stop() {
 }
 
 func (s *Server) setupMetrics(config *TelemetryConfig, serviceName string) error {
+	// Check the global metrics if they're matching with the provided config
+	if metrics.Enabled != config.Enabled || metrics.EnabledExpensive != config.Expensive {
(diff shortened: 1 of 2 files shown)


Raneet10 (approved):

cffls (approved):
LGTM, thanks!

PR #576, opened 2022-11-04
fix : TestRemoteMultiNotifyFull

0xsharma:
In this PR, we are fixing the indeterministic behaviour of TestRemoteMultiNotifyFull which caused failure of PRs in the CI. 
The error faced is : 
`sealer.go:394: WARN [11-04|16:45:09.256] Failed to notify remote miner            err="Post \"http://127.0.0.1:50643/\": context canceled"`


manav2401 (approved):
Not sure if 2 second delay is required, but LGTM!

0xKrishna (approved):

PR #597, opened 2022-11-24
Added flags to run heimdall as a child process

0xKrishna:
# Description

In this PR I have added 2 flags
- `bor.runheimdall` to run heimdall as a child process
- `bor.runheimdallargs` to provide arguments to the heimdall child process

These are supported in new cli/server and legacy geth start code as well.
Also heimdall has been imported as a package to start that as a process.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests
- [x] I have tested this code manually on mumbai


JekaMas (approved):

cmd/geth/main.go
```diff
@@ -347,6 +351,15 @@ func geth(ctx *cli.Context) error {
 	stack, backend := makeFullNode(ctx)
 	defer stack.Close()
 
+	if ctx.GlobalBool(utils.RunHeimdallFlag.Name) {
+		shutdownCtx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+		defer stop()
```
JekaMas, line 356: don't we cancel the context and HeimdallService by the end of `geth()` call?

0xKrishna: stop() is same as calling ctx.cancel()
That's why I used `defer stop()`
My understanding is this will call stop at the end of parent function which is `geth()` in this case.

JekaMas: It will stop by the end of the current function. Let's recheck it with a small example in play.go. could you make one?

0xKrishna: Yeah but it will only reach end of this function when the node is supposed to stop as `stack.Wait()` will keep it waiting.

cmd/geth/main.go
```diff
@@ -347,6 +351,15 @@ func geth(ctx *cli.Context) error {
...
+
+		go func() {
+			service.NewHeimdallService(shutdownCtx, getHeimdallArgs(ctx))
+		}()
+	}
+
 	startNode(ctx, stack, backend, false)
 	stack.Wait()
```
0xKrishna, line 364: @JekaMas Here, If I am not wrong this will keep the process running until it's supposed to terminate.

temaniarpit27 (approved):

cmd/geth/main.go
JekaMas: have you tested that it works fine? like with term signal or something?

0xKrishna: Yup, I have added few logs as well in graceful shutdown of all the services and it worked fine while testing.

temaniarpit27 (approved):

PR #600, opened 2022-11-25
Update default flags

manav2401:
This PR updates the default flags in the toml based config files. The defaults present in the code are kept as similar to geth. The config files for mumbai and mainnet are updated. The new values are as below

||miner.gaslimit|miner.gasprice|txpool.gaslimit|gpo.ignoreprice|
|---|---|---|---|---|
|mainnet|30M|30 Gwei|30 Gwei|30 Gwei|
|mumbai|30M|1 Gwei|1 wei|2 wei|

Additionally, this PR also changes the cache override mechanism. Instead it keeps geth's default values in code (which are also same as mumbai - 1024) and the mainnet config files are updated with a mainnet suitable value (which is 4096).

### DIFF
--- a/packaging/templates/testnet-v4/sentry/sentry/bor/config.toml
+++ b/packaging/templates/testnet-v4/sentry/sentry/bor/config.toml
@@ -41,12 +41,12 @@ syncmode = "full"
     # locals = []
     # journal = ""
     # rejournal = "1h0m0s"
-    # pricelimit = 30000000000
+    # pricelimit = 1
     # pricebump = 10
 
 [miner]
-    gaslimit = 20000000
-    # gasprice = "30000000000"
+    gaslimit = 30000000
+    # gasprice = "1000000000"
     # mine = false
     # etherbase = ""
     # extradata = ""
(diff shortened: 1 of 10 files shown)


JekaMas (approved):

cffls (approved):

pratikpatil024 (approved):

packaging/templates/testnet-v4/sentry/sentry/bor/config.toml
```diff
@@ -41,12 +41,12 @@ syncmode = "full"
...
     # rejournal = "1h0m0s"
-    # pricelimit = 30000000000
+    # pricelimit = 1
     # pricebump = 10
 
 [miner]
-    gaslimit = 20000000
-    # gasprice = "30000000000"
+    gaslimit = 30000000
+    # gasprice = "1000000000"
```
ssandeep, line 49: Is this supposed to be commented?

manav2401: Yes, the gas price is defaulted to 1 Gwei only in the code (which is correct for mumbai). 

I have updated the commented values (even though it does not matter technically) because they denote the code defaults. Even though we're not using it, it just represents that if you don't pass, this value is getting used in the code.

ssandeep (approved):

PR #604, opened 2022-11-29
Shivam/txpool tracing

0xsharma:
# Description

# Changes

# Breaking changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


temaniarpit27 (approved):

JekaMas (approved):

kaber2:
Again, what the hell is this? Did anyone review this mess? Its claiming to add tracing to tx_pool, while in reality it introduces an unholy locking mess, plus 1000 unrelated changed.

Looking at txpool:Pending(), you:

- add 6 atomic operations to the inner most loop, which will normally be executed hundreds of times
- add an uneccessary nil check for gasPriceUint
- repeatedly load gasPriceUint and baseFee, using 4 atomic ops inside the loop, even though they shouldn't normally change during execution and if they do, it would mess up the ordering of the transactions, so a single load outside of the loop would be the proper (and way faster) thing to do

On top of all of the implementation problems, you realize that even upstream geth intends to switch to uint256.Int long term, so you could have just done it properly instead of this unholy mess and merge it upstream.

Seriously, every change I look at is just garbage. There is no nicer way to say this.

PR #611, opened 2022-11-30
Cleanup for the packager to use git tag in the package profile naming…

djpolygon:
…. Added conditional check for directory structure, this is in prep for v0.3.1, as this will create a failure on upgrade path in package due to file exist

# Description

# Changes

- [ x ] Bugfix (non-breaking change that solves an issue)

# Breaking changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

During build, the output packages now use the tag properly instead of hardcoded static value as previous. I have also fixed an issue that will occur in the future for directory creation. 

# Additional comments

### DIFF
--- a/packaging/templates/package_scripts/postinst
+++ b/packaging/templates/package_scripts/postinst
@@ -2,6 +2,11 @@
 # This is a postinstallation script so the service can be configured and started when requested
 #
 sudo adduser --disabled-password --disabled-login --shell /usr/sbin/nologin --quiet --system --no-create-home --home /nonexistent bor
-sudo mkdir -p /var/lib/bor
-sudo chown -R bor /var/lib/bor
+if [ -d "/var/lib/bor" ]
+then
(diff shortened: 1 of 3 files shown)


packaging/templates/package_scripts/postinst
```diff
@@ -2,6 +2,11 @@
 # This is a postinstallation script so the service can be configured and started when requested
 #
 sudo adduser --disabled-password --disabled-login --shell /usr/sbin/nologin --quiet --system --no-create-home --home /nonexistent bor
```
cffls, line 4: If the user already exists, will this fail? Maybe it is better to check if user `bor` exists before adding it?

cffls: Looks like this command won't fail even when the user already exists.

PR #640, opened 2022-12-19
Shivam/pos 1041

0xsharma:
In this PR, we add tolerance of multiple keystore files error. Earlier, bor would exit when multiple keystores are found. Now, it can recover from that and will select the correct keystore from multiple files.

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -937,13 +937,13 @@ func unlockAccount(ks *keystore.KeyStore, address string, i int, passwords []str
 }
 
 func ambiguousAddrRecovery(ks *keystore.KeyStore, err *keystore.AmbiguousAddrError, auth string) accounts.Account {
-	fmt.Printf("Multiple key files exist for address %x:\n", err.Addr)
+	log.Warn("Multiple key files exist for", "address", err.Addr)
 
 	for _, a := range err.Matches {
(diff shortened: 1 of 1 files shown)


internal/cli/server/config.go
```diff
@@ -904,6 +902,61 @@ var (
 	gitDate          = "" // Git commit date YYYYMMDD of the release (set via linker flags)
 )
 
+// tries unlocking the specified account a few times.
+func unlockAccount(ks *keystore.KeyStore, address string, i int, passwords []string) (accounts.Account, string) {
```
temaniarpit27, line 906: was this func present earlier?

0xsharma: Yes it is present in main package. Cannot call it directly due to cyclic dependency.

internal/cli/server/config.go
```diff
@@ -904,6 +902,75 @@ var (
...
+			match = &a
+			break
+		}
+	}
+
+	if match == nil {
+		utils.Fatalf("None of the listed files could be unlocked.")
+	}
+
+	fmt.Printf("Your password unlocked %s\n", match.URL)
```
JekaMas, line 962: It might worth it to use standard logging.

JekaMas (approved):

PR #646, opened 2023-01-03
Heimdall App implementation

0xKrishna:
# Description

In this PR the goal is to implement heimdall interface using heimdallApp, So that we can directly communicate b/w bor & heimdall while they are running as a single process.

Some changes in this PR are also done to remove cyclic dependency, Basically some unused code which was creating that cyclic dependency is commented out.

Also the flags to run heimdall is marked as WIP so that people don't use it as of now, We will release this feature with proper doc after milestone HF.
# Changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


consensus/bor/heimdallapp/state_sync.go
```diff
@@ -0,0 +1,51 @@
+package heimdallapp
+
+import (
+	"context"
+	"encoding/json"
```
JekaMas, line 5: Should we use json-iter instead?

0xKrishna: I am not sure, I used what I generally use to do json marshal/unmarshal, Also I checked other implementations of IHeimdallClient, We are using `encoding/json` there as well, Let me know if you want me to change/update.

temaniarpit27 (approved):

temaniarpit27 (approved):

consensus/bor/heimdallapp/checkpoint.go
```diff
@@ -0,0 +1,50 @@
+package heimdallapp
+
+import (
+	"context"
+	"encoding/json"
```
JekaMas, line 5: Could you use go-iterator instead?

0xKrishna: Okay

JekaMas: and in other places is the same

consensus/bor/heimdallapp/state_sync.go
JekaMas: check other places - we started using go-iterator for quite a while

consensus/bor/heimdallapp/checkpoint.go
0xKrishna: Are you talking about this one ? https://pkg.go.dev/google.golang.org/api/iterator

0xKrishna: I think you were talking about this, My bad
https://pkg.go.dev/github.com/json-iterator/go

consensus/bor/heimdallapp/checkpoint.go
```diff
@@ -0,0 +1,50 @@
...
+	log.Info("Fetched checkpoint", "number", number)
+
+	resBytes, err := json.Marshal(res)
+	if err != nil {
+		return nil, err
+	}
+
+	var checkpoint checkpoint.Checkpoint
+
+	err = json.Unmarshal(resBytes, &checkpoint)
```
JekaMas, line 44: why we marshall to unmarshall then?

consensus/bor/heimdallapp/client.go
```diff
@@ -0,0 +1,27 @@
...
+
+func NewHeimdallAppClient() *HeimdallAppClient {
+	return &HeimdallAppClient{
+		hApp: service.GetHeimdallApp(),
+	}
+}
+
+func (h *HeimdallAppClient) Close() {
+	// Nothing to close as of now
+	log.Debug("Shutdown detected, Closing Heimdall App conn")
```
JekaMas, line 26: maybe warn level logging?

consensus/bor/heimdallapp/checkpoint.go
JekaMas: that one

JekaMas (approved):

consensus/bor/heimdallapp/state_sync.go
0xKrishna: Yes ser, On it

consensus/bor/heimdallapp/checkpoint.go
0xKrishna: To convert the response from `types.Checkpoint` to `checkpoint.Checkpoint`
Basically they have similar data but one is from heimdall repo (from where we are fetching checkpoint) and other is bor repo's checkpoint object.

0xKrishna: We can't directly assign as the data and json tags are same, But their internall types are different.

https://github.com/maticnetwork/heimdall/blob/develop/types/header.go#L9

https://github.com/maticnetwork/bor/blob/develop/consensus/bor/heimdall/checkpoint/checkpoint.go#L10

consensus/bor/heimdallapp/span.go
```diff
@@ -38,23 +37,10 @@ func toSpan(hdSpan *hmTypes.Span) *span.HeimdallSpan {
...
-	// set validator map
-	validatorsMap := make(map[common.Address]int, len(set.Validators))
-
-	for i, val := range set.Validators {
-		validatorsMap[val.Address] = i
-	}
-
-	set.SetMap(validatorsMap)
-
-	set.SetTotalVotingPower(vs.TotalVotingPower())
```
JekaMas: Don't we need to fill all fields?

0xKrishna: Nope, That is calculated in validator-set while processing. In other ways of communications (HTTP/gRPC) also we don't set totalVotingPower and the validator map as they are anyway calculated.

PR #651, opened 2023-01-06
consensus/bor: handle unauthorized signer in consensus.Prepare

manav2401:
# Description

This PR checks if the current signer (of a node) is eligible for mining (i.e. existing in a validator set) very early in `consensus.Prepare()`. Earlier it used to perform this check in `consensus.Seal()` which is too late in our case. It seems suitable for a PoW based network where you're not supposed to check eligibility to mine a block and try mining it whatever the case. In our case, a node which is not in the validator set will try to produce a block (prepare header, pick transactions by performing sorting in txpool, execute them, add state-sync by interacting with heimdall) and then in the last step before signing, it will know that it's not authorized to produce this block (i.e. does not exist in the validator set). This seems very unoptimised as we're wasting resources on things we're not supposed to do. 

# Changes

# Breaking changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -676,6 +676,13 @@ func (c *Bor) Prepare(chain consensus.ChainHeaderReader, header *types.Header) e
 
 	currentSigner := *c.authorizedSigner.Load()
 
+	// Bail out early if we're unauthorized to sign a block. This check also takes
+	// place before block is signed in `Seal`.
+	if !snap.ValidatorSet.HasAddress(currentSigner.signer) {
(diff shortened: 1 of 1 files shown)


manav2401:
I'd like to test these changes on a node. More than that, I'd like to know the reason for why this check was not added in prepare in the first place.

JekaMas (approved):

pratikspatil024 (approved):

PR #657, opened 2023-01-11
Added checks to RPC requests and introduced new flags to customise the parameters

pratikspatil024:
This PR adds

- 2 limits related to RPC requests and return messages 
  1. limits the maximum number of requests in a batch (default 100 requests)
  2. limits the maximum size of the result of the request (default 100000 bytes)
- and introduces 2 flags - 
  1. `rpcbatchlimit` - Maximum number of messages in a batch
  3. `rpcreturndatalimit` - Maximum size (in bytes) a result of an rpc request could have

This gives a node more control over the requests that are being sent to it and make sure that they do not consume a big part of the memory.


cffls (approved):

JekaMas (approved):

moraesjeremias:
Hey @pratikspatil024 thanks for this contribution to bor client.
I'd like to know where `rpc.batchlimit` and `rpc.returndatalimit` default values (100 and 100000, respectively) came from? 

Did you run internal perf tests to evaluate best values for both limits? Could you share any info related to these default thresholds?

I appreciate all shared info!

pratikspatil024:
Hi, @moraesjeremias thanks for the question.

We observed that if these two limits are absent, then the  CPU usage will go high, and thus the process running bor will stop.
We tested these values for different types of machines and concluded that it's safe to use lower default values for both parameters. I agree that 200 and 100000 default values (respectively) will also work for a higher-spec machine, but it's up to the user to configure this for his/her client.

Also, Erigon is using these same values, so it was better to use the same in bor for consistency.

PR #666, opened 2023-01-13
Added hardfork related checks and updated existing tests

pratikspatil024:


miner/worker_test.go
```diff
@@ -810,11 +812,13 @@ func BenchmarkBorMiningBlockSTMMetadata(b *testing.B) {
 				b.Fatalf("failed to insert new mined block %d: %v", block.NumberU64(), err)
 			}
 
-			// check for dependencies
-			deps := block.TxDependency()
-			for i := 1; i < block.Transactions().Len(); i++ {
-				if deps[i][0] != uint64(i) || deps[i][1] != uint64(0) || deps[i][2] != uint64(i-1) || len(deps[i]) != 3 {
-					b.Fatalf("wrong dependency")
+			// check for dependencies for block number > 4
+			if block.NumberU64() > 4 {
```
cffls, line 816: Can you add a check that, when block number is less than 4, block.TxDependency is nil?

pratikspatil024: Oh yes, my bad. Thanks.

cffls (approved):
LGTM, thank you!

PR #671, opened 2023-01-15
chg : trieTimeout from 60 to 10 mins

0xsharma:
# Description
In this PR we resolve an issue of huge rewinds. `trieTimeout`/`trieTimeLimit` is responsible for the frequency of storing block state to the persistent db for a full node. It was set to 60 minutes which resulted in huge gap between states of blocks in the persistent memory. We have reduced this gap to 10 minutes which will increase the frequency of storing the block states to the persistent database.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests
- [x] I have tested this code manually on mumbai

### DIFF
--- a/builder/files/config.toml
+++ b/builder/files/config.toml
@@ -127,7 +127,7 @@ syncmode = "full"
   # preimages = false
   # txlookuplimit = 2350000
   # triesinmemory = 128
-  # timeout = "1h0m0s"
+  # timeout = "10m0s"
 
 [accounts]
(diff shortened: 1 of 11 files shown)


marcello33 (approved):

JekaMas (approved):

pratikspatil024 (changes requested):
Please update the default values in `config.toml` files as well.

pratikspatil024 (approved):

PR #695, opened 2023-01-20
fix nil state-sync issue, increase grpc limit

manav2401:
# Description

Due to a nil value in state sync slice, a filter receiving it's subscription panics. This PR fixes it.

Additional context:
The issue was because the slice which was to be returned was initialised with all empty state sync objects before even it was filled with data. Due to some cause (mostly a validation issue), if the execution loop breaks, it will return that half-filled slice which has some nil elements. Others (filter services in this case) assume that there can’t be a nil element and hence they panic on receiving one.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/filters/bor_api.go
+++ b/eth/filters/bor_api.go
@@ -67,8 +67,8 @@ func (api *PublicFilterAPI) NewDeposits(ctx context.Context, crit ethereum.State
 		for {
 			select {
 			case h := <-stateSyncData:
-				if crit.ID == h.ID || bytes.Compare(crit.Contract.Bytes(), h.Contract.Bytes()) == 0 ||
-					(crit.ID == 0 && crit.Contract == common.Address{}) {
+				if h != nil && (crit.ID == h.ID || bytes.Equal(crit.Contract.Bytes(), h.Contract.Bytes()) ||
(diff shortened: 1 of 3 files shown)


JekaMas (approved):

eth/filters/bor_api.go
```diff
@@ -67,8 +67,8 @@ func (api *PublicFilterAPI) NewDeposits(ctx context.Context, crit ethereum.State
 		for {
 			select {
 			case h := <-stateSyncData:
-				if crit.ID == h.ID || bytes.Compare(crit.Contract.Bytes(), h.Contract.Bytes()) == 0 ||
-					(crit.ID == 0 && crit.Contract == common.Address{}) {
+				if h != nil && (crit.ID == h.ID || bytes.Equal(crit.Contract.Bytes(), h.Contract.Bytes()) ||
```
JekaMas, line 70: crit.Contract == h.Contract

temaniarpit27 (approved):

PR #697, opened 2023-01-20
consensus/bor : add : devFakeAuthor flag

0xsharma:
# Description

In this PR, we added `--bor.devfakeauthor` flag. Using this we can run bor as a validator without the requirement of initial validator-set or Heimdall. This flag must be used with `--bor.withoutheimdall`.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -227,7 +227,8 @@ type Bor struct {
 	HeimdallClient         IHeimdallClient
 
 	// The fields below are for testing only
-	fakeDiff bool // Skip difficulty verifications
+	fakeDiff      bool // Skip difficulty verifications
+	devFakeAuthor bool
 
 	closeOnce sync.Once
 }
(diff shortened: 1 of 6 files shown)


pratikspatil024 (changes requested):
1. As we will be using this flag with `bor.withoutheimdall`, I think we should use the same config (`HeimdallConfig`) which `bor.withoutheimdall` uses. Just to keep these flags organised.
2. Please update the docs by running `make docs` once the final place of this flag is decided.

0xsharma:
@pratikspatil024 
1. This change is not just related to heimdall. It's a superset of that. This is related to developer mode where we would want a validator node running on bor consensus which can effectively create blocks without need of genesis contracts and heimdall.

2. Yes, this is in draft stage. Once everything is finalized, we will edit the docs.

JekaMas (approved):

pratikspatil024 (approved):

temaniarpit27 (approved):

PR #700, opened 2023-01-25
Added new api to support conditional transactions (EIP-4337)

pratikspatil024:
Added a new API to support conditional transactions.

Here is an example request:
```
curl localhost:8545 -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0", "id": 1, "method":"eth_sendRawTransactionConditional", "params": ["0x7004c824391ea40b1084e4cad27d3a550e1f784d274b26ec96a09aa9ca36e66e", {"knownAccounts": {"0xadd1add1add1add1add1add1add1add1add1add1": "0x313AaDcA1750CaadC7BCB26FF08175c95DCf8E38", "0xadd2add2add2add2add2add2add2add2add2add2": {"0x0000000000000000000000000000000000000000000000000000000000000aaa": "0x0000000000000000000000000000000000000000000000000000000000000bbb", "0x0000000000000000000000000000000000000000000000000000000000000ccc": "0x0000000000000000000000000000000000000000000000000000000000000ddd"}}}]}'
```


accounts/abi/bind/backends/simulated.go
```diff
@@ -675,6 +675,40 @@ func (b *SimulatedBackend) SendTransaction(ctx context.Context, tx *types.Transa
...
 
+// a replica of above `SendTransaction` function
+func (b *SimulatedBackend) SendTransactionConditional(ctx context.Context, tx *types.Transaction, knownAccounts map[string]map[common.Address]interface{}) error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	// Get the last block
+	block, err := b.blockByHash(ctx, b.pendingBlock.ParentHash())
+	if err != nil {
+		return fmt.Errorf("could not fetch parent")
```
JekaMas, line 686: Could you use errors as predefined values? Also, I believe it's useful to add come context to each error and use `errors.New()` for defining new errors and `fmt.Errorf("%w %v - any placeholders", errMyError, var1, var2...)` providing additional context to an error.

accounts/abi/bind/backends/simulated.go
```diff
@@ -675,6 +675,40 @@ func (b *SimulatedBackend) SendTransaction(ctx context.Context, tx *types.Transa
...
+	if err != nil {
+		return fmt.Errorf("invalid transaction: %v", err)
+	}
+	nonce := b.pendingState.GetNonce(sender)
+	if tx.Nonce() != nonce {
+		return fmt.Errorf("invalid transaction nonce: got %d, want %d", tx.Nonce(), nonce)
+	}
+	// Include tx in chain
+	blocks, _ := core.GenerateChain(b.config, block, ethash.NewFaker(), b.database, 1, func(number int, block *core.BlockGen) {
+		for _, tx := range b.pendingBlock.Transactions() {
```
JekaMas, line 700: could we remove variable shadowing?

JekaMas: could `pending` block be changed during the for loop? is it thread safe to perform such loop?

ethclient/ethclient_test.go
```diff
@@ -569,6 +569,8 @@ func testCallContract(t *testing.T, client *rpc.Client) {
 func testAtFunctions(t *testing.T, client *rpc.Client) {
 	ec := NewClient(client)
 
+	sendTransactionConditional(ec)
```
JekaMas, line 572: should we check the error here?

accounts/abi/bind/backends/simulated.go
pratikspatil024: Right, this can be done.
This function is the replica of the `SendTransaction` function defined just above it. It is not used anywhere, and I have just kept it here because it was included in the interface.

accounts/abi/bind/backends/simulated.go
pratikspatil024: I think this function is used only for testing purposes, so I think this is fine. What do you think?

accounts/abi/bind/backends/simulated.go
```diff
@@ -675,6 +675,40 @@ func (b *SimulatedBackend) SendTransaction(ctx context.Context, tx *types.Transa
...
+			block.AddTxWithChain(b.blockchain, tx)
+		}
+		block.AddTxWithChain(b.blockchain, tx)
+	})
+	stateDB, _ := b.blockchain.State()
+
+	b.pendingBlock = blocks[0]
+	b.pendingState, _ = state.New(b.pendingBlock.Root(), stateDB.Database(), nil)
```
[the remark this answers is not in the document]
pratikspatil024: Yes, I will add an error check here. Thanks!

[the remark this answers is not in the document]
pratikspatil024: Yes, I will add an error check here. Thanks!

internal/ethapi/api.go
```diff
@@ -1928,6 +1929,46 @@ func (s *PublicTransactionPoolAPI) SendRawTransaction(ctx context.Context, input
...
+	tx := new(types.Transaction)
+	if err := tx.UnmarshalBinary(input); err != nil {
+		return common.Hash{}, err
+	}
+
+	if knownAccounts["knownAccounts"] == nil {
+		// should we allow this?
+		// throwing an error, for now
+		return common.Hash{}, errors.New("knownAccounts is empty")
+	}
```
cffls, line 1944: I don't think we should assume this to be always non-empty. Bundlers should be fully responsible for this field.

pratikspatil024: Okay, that makes sense.

cffls (approved):

PR #701, opened 2023-01-25
Arpit/temp bor sync

temaniarpit27:
# Description

Fix issues raised by Alchemy for bor syncing

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/rawdb/bor_receipt.go
+++ b/core/rawdb/bor_receipt.go
@@ -7,6 +7,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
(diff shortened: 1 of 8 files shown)


core/rawdb/bor_receipt.go
```diff
@@ -114,8 +114,8 @@ func ReadBorReceipt(db ethdb.Reader, hash common.Hash, number uint64) *types.Rec
 		return nil
 	}
 
-	body := ReadBody(db, hash, number)
-	if body == nil {
+	body := HasBody(db, hash, number)
```
JekaMas, line 116: I read more code, let's revert this change.

internal/ethapi/api.go
```diff
@@ -1453,15 +1453,23 @@ func newRPCPendingTransaction(tx *types.Transaction, current *types.Header, conf
 func newRPCTransactionFromBlockIndex(b *types.Block, index uint64, config *params.ChainConfig, db ethdb.Database) *RPCTransaction {
 	txs := b.Transactions()
 
-	borReceipt := rawdb.ReadBorReceipt(db, b.Hash(), b.NumberU64())
-	if borReceipt != nil {
-		tx, _, _, _ := rawdb.ReadBorTransaction(db, borReceipt.TxHash)
+	if index >= uint64(len(txs)+1) {
```
JekaMas, line 1456: Won't it suppress working with Bor transactions?

manav2401: Earlier `if index == uint64(len(txs))` check wasn't there and hence we added check on top before even making a call. I think with this check being added, we can remove this `if index >= uint64(len(txs)+1)` totally. But to answer your question, the check if against `len(txs)+1` so it keeps a buffer incase state-sync transactions are present.

cffls: The check also prevents out of range index, which could be malicious, from invoking ReadBorReceipt or ReadBorTransaction.

pratikspatil024 (approved):

PR #705, opened 2023-01-27
core,eth,internal/cli,internal/ethapi: add --rpc.allow-unprotected-txs flag to allow txs to get replayed (for shadow node) 

Raneet10:
# Description

This PR introduces `rpc.allow-unprotected-txs` flag to facilitate replaying of transactions from a different network to a shadow node (i.e a node producing blocks in isolation with the state of the network from which transactions are being relayed). 

# Changes

# Nodes audience

Currently, this flag is only meant to be enabled if we are running a shadow node.

# Checklist

## Testing

- [ ] I have added unit tests

### DIFF
--- a/builder/files/config.toml
+++ b/builder/files/config.toml
@@ -13,7 +13,7 @@ syncmode = "full"
 # snapshot = true
 # "bor.logs" = false
 # ethstats = ""
-
+# devfakeauthor = false
 # ["eth.requiredblocks"]
 
 [p2p]
@@ -65,6 +65,7 @@ syncmode = "full"
 #   ipcpath = ""
 #   gascap = 50000000
 #   txfeecap = 5.0
+#   allow-unprotected-txs = false
 #   [jsonrpc.http]
 #     enabled = false
 #     port = 8545
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -227,7 +227,8 @@ type Bor struct {
 	HeimdallClient         IHeimdallClient
 
 	// The fields below are for testing only
-	fakeDiff bool // Skip difficulty verifications
+	fakeDiff      bool // Skip difficulty verifications
+	devFakeAuthor bool
 
 	closeOnce sync.Once
 }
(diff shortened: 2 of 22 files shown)


pgebheim:
This patch is also useful for testing 4337 so that deterministic deployer contracts can be deployed at the correct address for testing.

cc: @pratikspatil024

Raneet10:
> This patch is also useful for testing 4337 so that deterministic deployer contracts can be deployed at the correct address for testing.
> 
> cc: @pratikspatil024

I'm afraid I don't have much context. But perhaps you could enlighten me on this :)

pratikspatil024 (commented):
Please update all the TOML config files.

Raneet10:
Done: [4917fde](https://github.com/maticnetwork/bor/pull/705/commits/4917fde5be2a8c1eb5f6147b1900d52cefcd7dbd)

temaniarpit27 (approved):

marcello33 (approved):

pratikspatil024 (approved):

kaber2:
Just noticed this gem while merging the latest changes. Does anyone actually review these patches?

`
// Exclude transactions with invalid signatures as soon as
// possible and cache senders in transactions before
// obtaining lock

if pool.config.AllowUnprotectedTxs {
        pool.signer = types.NewFakeSigner(tx.ChainId())
`

Why the hell is this changing global pool configuration *inside the transaction add path*?

Raneet10:
Hey @kaber2 , firstly let's please try to watch our tone in these conversations; we're all trying to collaborate here. Secondly, the signer change happens only when `AllowUnprotectedTxs` is set to true (mostly when you're running a shadow forked node) and we override the signer with the chain ID of the network from which the transaction is coming (instead of the local network's) since we're allowing transactions from a different network to be replayed. We can maybe make this change on an upper level of the call stack, which would be logically sound though my initial assumption was that this might result in some repetitive code. Of course, open to your *mild* thoughts as to where this condition should be placed!

PR #707, opened 2023-01-28
Revert "Reduce txArriveTimeout to 100ms"

ephess:
This reverts commit 243d231fe45bc02f33678bb4f69e941167d7f466.

# Description

I'm part of the team at FastLane Labs who are building out an ecosystem friendly MEV protocol for Polygon. Currently, we are in open beta and have validators representing more than 20% of new block production onboard.

This PR reverts a recent commit that has made it into the current develop branch which significantly reduces the txArriveTimeout variable in the `eth/fetcher/tx_fetcher.go` file.

This change seems to have been committed straight into develop, rather than going through a PR so it's hard for us to understand the rationale behind it, but we'd be keen to discuss that here.

## Potential issues introduced by the change

Reducing the txArriveTimeout value to from 500 ms to 100 ms seems likely to increase the amount of noise in the p2p layer of the network, and will also likely increase bandwidth consumption for full node operators.

This is because the 100ms value that is being introduced here is much less than the worst case round-the-world P95 latency of around 350ms-400ms. This means that if a transaction originates in another part of the world, this change will increase the likelihood that a node which is the recipient of an announcement makes a request for the full body of a transaction that it likely would have received by direct propagation anyway.

This has the effect of both increasing the number of redundant messages being sent through the p2p layer, but also increasing the number of full transactions that are being sent, both of which have an impact on the network resources of node operators.

## Impact on the FastLane System

The current delay of 500ms on validator nodes is a key element that keeps the design of the FastLane MEV protocol simple, and makes the [patch](https://github.com/Polygon-Fast-Lane/sentry-patch/blob/main/announce_only.patch) that needs to be applied to validators sentries easy to understand.

One of the main goals of the FastLane protocol is to be as ecosystem friendly as possible, and to help improve the health of the p2p layer of the network by eliminating one of the primary incentives for network participants to spam: namely trying to be directly propagated from the validators sentry to the validator itself, rather than going through the announce mechanism. For a backrunner, your chance to be directly propagated is directly proportional to the number of transactions you send. As adoption of the FastLane protocol increases and we remove this mechanism, the incentive to spam on the network will drop proportionally which will eventually result in a much healthier network for both users and node operators.

As things are today, this presents a breaking change to the design of the FastLane protocol and we would be interested in working with the Polygon team to understand the intent of this change, figure out if this is the best way to achieve this goal, and finally if it is to coordinate between FastLane and Polygon to ensure a seamless release for all users of the network.

# Additional comments

I have gone and reviewed geth and the bsc node fork to confirm that the original 500 ms setting is still being used on both of these chains as well, and that is still the case:

https://github.com/bnb-chain/bsc/blob/master/eth/fetcher/tx_fetcher.go
https://github.com/ethereum/go-ethereum/blame/master/eth/fetcher/tx_fetcher.go

The FastLane mechanism is outlined in our whitepaper [here](https://www.fastlane.finance/PFL_WHITE_PAPER_1_5.pdf).

### DIFF
--- a/eth/fetcher/tx_fetcher.go
+++ b/eth/fetcher/tx_fetcher.go
@@ -55,11 +55,11 @@ const (
 
 	// txArriveTimeout is the time allowance before an announced transaction is
 	// explicitly requested.
-	txArriveTimeout = 100 * time.Millisecond
+	txArriveTimeout = 500 * time.Millisecond
 
 	// txGatherSlack is the interval used to collate almost-expired announces
 	// with network fetches.
-	txGatherSlack = 20 * time.Millisecond
+	txGatherSlack = 100 * time.Millisecond
 )
 
 var (


pratikspatil024:
Hello, thank you for the information and PR. We will discuss this internally with the team.



> This change seems to have been committed straight into develop, rather than going through a PR so it's hard for us to understand the rationale behind it, but we'd be keen to discuss that here.

Just FYI, here's the [PR](https://github.com/maticnetwork/bor/pull/627) which merges [this commit](https://github.com/maticnetwork/bor/commit/243d231fe45bc02f33678bb4f69e941167d7f466). The PR has ample of information regarding this change.

ephess:
Thanks @pratikspatil024, appreciate you pointing me to the PR for this. For some reason the commit isn't linked against it in GitHub for me, I should have searched for it as well!

Looking forward to hearing back from you.

JekaMas (changes requested):
Hi @ephess !
Really appreciate your help.
I understand your warry, however we'd need more data to proof that this change will effect p2p in a bad way. Do you have any data to analyze?
The previous discussion was in https://github.com/maticnetwork/bor/pull/627

JekaMas:
@temaniarpit27 @cffls

JekaMas (approved):
We are not in a rush with reducing this parameter.
Let's revert the change, research more and make it configurable with a default value, that we will get as a result of the research.

cffls (approved):

pratikspatil024 (approved):

ephess:
That's great @JekaMas, appreciate your input here.

Who's responsible for merging this PR? I don't see a assignee on it yet. I don't think there's anything I can do here from what I can see.

We are working on getting together a PR for making this configurable at the moment, so hopefully we can get that merged shortly after this revert goes through.

JekaMas:
@temaniarpit27 Could you merge the PR, introduce new 'MEV' label, mark all related to MEV PRs and issues, make a new task to discuss proper txpool, txfetcher params?

temaniarpit27:
> @temaniarpit27 Could you merge the PR, introduce new 'MEV' label, mark all related to MEV PRs and issues, make a new task to discuss proper txpool, txfetcher params?

waiting for tests to be completed

ajb:
This is completely unacceptable. Fast Lane Labs (a for-profit company) should not be allowed to bully Polygon into reverting an accepted change to bor simply because it threatens their attempted monopoly, which is built on a fundamentally flawed design.

> This change seems to have been committed straight into develop, rather than going through a PR so it's hard for us to understand the rationale behind it, but we'd be keen to discuss that here.

What do you mean? There is a PR with tons of discussion here: https://github.com/maticnetwork/bor/pull/627

> I have gone and reviewed geth and the bsc node fork to confirm that the original 500 ms setting is still being used on both of these chains as well, and that is still the case:

This is silly, the block time on Ethereum is 5x longer than Polygon.  BSC has a centralized validator set. Not to mention numerous other differences between the networks.

PR #708, opened 2023-01-30
Added flag in Status to wait for backend, and fixed panic issue.

pratikspatil024:
Added an optional flag (`-w`) to the `bor status` command, which waits for at max 10 seconds (5 iterations of 2 seconds each) if the backend is not ready.

`bor status -w` and `bor status` both of these commands are valid.


internal/cli/server/service.go
```diff
@@ -165,6 +165,10 @@ func (s *Server) ChainSetHead(ctx context.Context, req *proto.ChainSetHeadReques
 }
 
 func (s *Server) Status(ctx context.Context, _ *empty.Empty) (*proto.StatusResponse, error) {
+	if s.backend == nil {
+		return nil, fmt.Errorf("the service is currently unavailable, try again later")
```
JekaMas, line 169: why fmt.Errorf ?

pratikspatil024: Changed to `errors.New()`, thanks.

JekaMas: Also possible to define a package level error variable and reuse it here.

pratikspatil024: Done, thanks!

internal/cli/server/service.go
```diff
@@ -23,6 +24,8 @@ import (
 
 const chunkSize = 1024 * 1024 * 1024
 
+var ErrUnavailable = errors.New("the service is currently unavailable, try again later")
```
cffls, line 27: Let's be more specific on this error, e.g.

```
var ErrBorServiceUnavailable = errors.New("Bor service is currently unavailable, try again later")
```

pratikspatil024: Done, thanks!

manav2401 (approved):
LGTM. Do you think we should add a test for the status command?

pratikspatil024:
Added a test to check if the command is successful or not. But this does not cover the whole code, as the backend gets available before we can get the grpc port. A solution for this is to break the `CreateMockServer` into 2 parts, the first one will allocate the ports, and the second one will use those ports, and in between, we can make a call to the Status command.

cffls (approved):

PR #712, opened 2023-02-01
Ratelimit RPC concurrent requests

0xsharma:
# Description

In this PR, we introduce to new flags `--http.maxconcreq` and `--ws.maxconcreq`. These flags can be used to ratelimit the number concurrent RPC requests on the client. `0` is the default value which `disables` the rate-limiting feature.

# Changes

# Nodes audience

This test is to facilitate the RPC providers to have more control on the RPC client in Bor.

# Checklist

## Testing

- [ ] I have added unit tests
- [x] I have tested this code manually on mainnet

### DIFF
--- a/node/api.go
+++ b/node/api.go
@@ -78,6 +78,24 @@ func (api *privateAdminAPI) AddPeer(url string) (bool, error) {
 	return true, nil
 }
 
+type MaxConcReq struct {
+	HttpLimit uint64 // max concurrent requests for http requests. 0(Default) means no limit
+	WSLimit   uint64 // max concurrent requests for ws requests. 0(Default) means no limit
+}
+
+// get max concurrent requests parameter for rpc requests
+func (api *privateAdminAPI) GetMaxConcReq() *MaxConcReq {
+	httpLimit := api.node.http.httpConfig.maxConcReq
+	wsLimit := api.node.ws.wsConfig.maxConcReq
+
+	maxConcReq := &MaxConcReq{
+		HttpLimit: httpLimit,
+		WSLimit:   wsLimit,
+	}
+
+	return maxConcReq
+}
+
 // RemovePeer disconnects from a remote node if the connection exists
 func (api *privateAdminAPI) RemovePeer(url string) (bool, error) {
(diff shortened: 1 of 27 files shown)


rpc/server.go
```diff
@@ -93,6 +100,21 @@ func (s *Server) ServeCodec(codec ServerCodec, options CodecOption) {
 // is used to serve HTTP connections. Subscriptions and reverse calls are not allowed in
 // this mode.
 func (s *Server) serveSingleRequest(ctx context.Context, codec ServerCodec) {
+
+	if s.maxConReq > 0 {
+		if s.reqCount > s.maxConReq { //if maxConReq is 0, then no limit
+			maxConcReqDiscardedTxs.Inc(1)
+			log.Warn("RPC server ratelimiting", "Concurrent Reqs", s.reqCount, "MaxConcurrentReqs", s.maxConReq)
```
JekaMas, line 107: I believe we should use atomic here as well. To prevent such things, could you use https://pkg.go.dev/sync/atomic#Int64 instead of plain int64?

JekaMas (approved):
LGTM with a small comment

rpc/server.go
0xsharma: Done !

node/api.go
```diff
@@ -78,6 +78,24 @@ func (api *privateAdminAPI) AddPeer(url string) (bool, error) {
...
 }
 
+type MaxConcReq struct {
+	HttpLimit uint64 // max concurrent requests for http requests. 0(Default) means no limit
+	WSLimit   uint64 // max concurrent requests for ws requests. 0(Default) means no limit
+}
+
+// get max concurrent requests parameter for rpc requests
+func (api *privateAdminAPI) GetMaxConcReq() *MaxConcReq {
+	httpLimit := api.node.http.httpConfig.maxConcReq
```
JekaMas, line 88: For changing these values online via RPC, I believe, we'd need to have them atomic. How do you thing?

0xsharma: Done ✅

temaniarpit27 (approved):

PR #714, opened 2023-02-02
Make worker size and timeout configurable for execution pool

temaniarpit27:
# Description

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


rpc/client.go
```diff
@@ -112,7 +112,8 @@ func (c *Client) newClientConn(conn ServerCodec) *clientConn {
 	ctx := context.Background()
 	ctx = context.WithValue(ctx, clientContextKey{}, c)
 	ctx = context.WithValue(ctx, peerInfoContextKey{}, conn.peerInfo())
-	handler := newHandler(ctx, conn, c.idgen, c.services, NewExecutionPool(0))
+	// TODO - Check this
+	handler := newHandler(ctx, conn, c.idgen, c.services, NewExecutionPool(0), 10*time.Second)
```
JekaMas, line 116: why 10*time.Second? Maybe treat 0 as 'switched off'?

rpc/endpoints.go
```diff
@@ -27,7 +28,7 @@ import (
 func StartIPCEndpoint(ipcEndpoint string, apis []API) (net.Listener, *Server, error) {
 	// Register all the APIs exposed by the services.
 	var (
-		handler    = NewServer()
+		handler    = NewServer(100, 10*time.Second)
```
JekaMas, line 31: could we get rid of hard coded values?

rpc/handler.go
```diff
@@ -64,28 +64,30 @@ type handler struct {
 	subLock    sync.Mutex
 	serverSubs map[ID]*Subscription
 
-	executionPool *SafePool
+	executionPoolRequestTimeout time.Duration
+	executionPool               *SafePool
```
JekaMas, line 68: If we don't have request specific timeouts, then `executionPoolRequestTimeout` can be part of executionPool.

JekaMas (approved):

rpc/handler.go
JekaMas: also it should be atomic to make it changeable in runtime

internal/jsre/deps/web3.js
```diff
@@ -1192,7 +1192,7 @@ module.exports = SolidityTypeInt;
     You should have received a copy of the GNU Lesser General Public License
     along with web3.js.  If not, see <http://www.gnu.org/licenses/>.
 */
```
JekaMas, line 1194: do we need this file changed?

node/api.go
```diff
@@ -342,3 +343,57 @@ func (s *publicWeb3API) ClientVersion() string {
...
+	WSLimit   uint64
+}
+
+type ExecutionPoolRequestTimeout struct {
+	HttpLimit time.Duration
+	WSLimit   time.Duration
+}
+
+func (api *privateAdminAPI) GetExecutionPoolThreads() *ExecutionPoolThreads {
+	httpLimit := api.node.http.httpConfig.threads
```
JekaMas, line 358: if these values can be changed in runtime, then we should have either atomics or a mutex for them

node/api.go
```diff
@@ -342,3 +343,57 @@ func (s *publicWeb3API) ClientVersion() string {
...
+	executionPoolThreads := &ExecutionPoolThreads{
+		HttpLimit: httpLimit,
+		WSLimit:   wsLimit,
+	}
+
+	return executionPoolThreads
+}
+
+func (api *privateAdminAPI) GetExecutionPoolRequestTimeout() *ExecutionPoolRequestTimeout {
+	httpLimit := api.node.http.httpConfig.requesttimeout
```
JekaMas, line 370: will `ChangeSize, ChangeTimeout` change this config values as well? If not we need either to start changing them or read actual values from SafePool.

rpc/server.go
```diff
@@ -49,26 +50,34 @@ type Server struct {
...
 
 	// Register the default service providing meta information about the RPC service such
 	// as the services and methods it offers.
 	rpcService := &RPCService{server}
 	server.RegisterName(MetadataApi, rpcService)
-
 	return server
 }
 
+func (s *Server) SetExecutionPoolSize(n int) {
```
JekaMas, line 73: Could we also log that changes as in https://github.com/maticnetwork/bor/pull/712/files# (please ask @0xsharma for the latest changes)? It'd be great to have these logs similar.

internal/jsre/deps/web3.js
temaniarpit27: no

node/api.go
```diff
@@ -342,3 +343,61 @@ func (s *publicWeb3API) ClientVersion() string {
...
+		HttpLimit: httpLimit,
+		WSLimit:   wsLimit,
+	}
+
+	return executionPoolRequestTimeout
+}
+
+func (api *privateAdminAPI) SetWSExecutionPoolRequestTimeout(n int) *ExecutionPoolRequestTimeout {
+	api.node.ws.wsHandler.Load().(*rpcHandler).server.SetExecutionPoolRequestTimeout(time.Duration(n) * time.Second)
+	log.Info("modifying ws execution pool request timeout", "timeout", n)
```
JekaMas, line 383: let's make it warning level

rpc/endpoints.go
```diff
@@ -27,7 +28,7 @@ import (
 func StartIPCEndpoint(ipcEndpoint string, apis []API) (net.Listener, *Server, error) {
 	// Register all the APIs exposed by the services.
 	var (
-		handler    = NewServer()
+		handler    = NewServer(10, time.Duration(10)*time.Second)
```
JekaMas, line 31: do we need it hard-coded?

temaniarpit27: no i am checking whether this is causing ipc issues

rpc/execution_pool.go
```diff
@@ -8,27 +8,27 @@ import (
...
 
-const (
-	threads        = 100
-	requestTimeout = 0 //10 * time.Second
-)
-
 type SafePool struct {
 	executionPool *atomic.Pointer[workerpool.WorkerPool]
 	fastPath      bool
+	timeout       *atomic.Pointer[time.Duration]
```
JekaMas, line 14: if timeout was 0 then fastPath was true, if we then change timeout to non-zero value, fastPath incorrectly will be kept true.
I believe we need `fastPath` atomic and change `ChangeTimeout` `Change*`. Maybe have 1 RWMutex instead 2 atomic values?

temaniarpit27: fastpath is dependent upon size and not timeout I think

docs/cli/example_config.toml
```diff
@@ -72,25 +72,30 @@ ethstats = ""                # Reporting URL of a ethstats service (nodename:sec
     api = ["eth", "net", "web3", "txpool", "bor"]  # API's offered over the HTTP-RPC interface
     vhosts = ["localhost"]                         # Comma separated list of virtual hostnames from which to accept requests (server enforced). Accepts '*' wildcard.
     corsdomain = ["localhost"]                     # Comma separated list of domains from which to accept cross origin requests (browser enforced)
+    ep-size = 100                                  # Maximum size of workers to run in rpc execution pool for HTTP requests (default: 100)
```
JekaMas, line 75: let's make it 40

internal/jsre/deps/web3.js
JekaMas: Could we revert formatting this file?

node/api.go
```diff
@@ -342,3 +343,69 @@ func (s *publicWeb3API) ClientVersion() string {
...
+func (api *privateAdminAPI) SetWSExecutionPoolSize(n int) *ExecutionPoolSize {
+	api.node.ws.wsConfig.executionPoolSize = uint64(n)
+	api.node.ws.wsHandler.Load().(*rpcHandler).server.SetExecutionPoolSize(n)
+	log.Warn("Updating ws execution pool size", "threads", n)
+
+	return api.GetExecutionPoolSize()
+}
+
+func (api *privateAdminAPI) SetHttpExecutionPoolSize(n int) *ExecutionPoolSize {
+	api.node.http.httpConfig.executionPoolSize = uint64(n)
```
JekaMas, line 406: is it atomic? why we need both? direct `=` and `SetExecutionPoolSize` ?

JekaMas (approved):

node/api.go
manav2401: One of them is updating the http config which does not need to be atomic. Another one is actually calling the RPC function to set the new size which will actually update the number of workers used by the execution pool. There's a mutex for that in the library.

JekaMas: could we have a few SetHttpExecutionPoolSize at the same time?

internal/jsre/deps/web3.js
manav2401: Yes, will do

PR #722, opened 2023-02-06
Added verbosity flag, supports log-level as well, but will remove that in future.

pratikspatil024:
`--verbosity` supports these (`TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, and `CRIT`) 6 logging levels.

### DIFF
--- a/docs/cli/bootnode.md
+++ b/docs/cli/bootnode.md
@@ -6,7 +6,9 @@
 
 - ```v5```: Enable UDP v5 (default: false)
 
-- ```verbosity```: Logging verbosity (trace|debug|info|warn|error|crit) (default: info)
+- ```verbosity```: Logging verbosity (5=trace|4=debug|3=info|2=warn|1=error|0=crit) (default: 3)
+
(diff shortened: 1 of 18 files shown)


Raneet10 (approved):

anshalshukla (approved):

docs/cli/bootnode.md
```diff
@@ -6,7 +6,7 @@
 
 - ```v5```: Enable UDP v5 (default: false)
 
-- ```log-level```: Log level (trace|debug|info|warn|error|crit) (default: info)
+- ```verbosity```: Logging verbosity (trace|debug|info|warn|error|crit) (default: info)
```
cffls, line 9: Verbosity was an int flag: https://github.com/maticnetwork/bor/blob/2321e64ed7f14a82776c854607148810f8389d17/cmd/evm/main.go#L88-L91 I think for consistency, we should make it an int flag in new CLI too.

pratikspatil024: Done, thanks.

cffls (approved):

PR #726, opened 2023-02-07
Feat : SetMaxPeers

0xsharma:
# Description

In this PR, we are adding two new RPC methods to setMaxPeers and getMaxPeers of a node respectively.
With this change, maxPeers can be adjusted in runtime. This can be used to perform several chaotic tests.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### DIFF
--- a/internal/web3ext/web3ext.go
+++ b/internal/web3ext/web3ext.go
@@ -192,6 +192,15 @@ web3._extend({
 			name: 'stopWS',
 			call: 'admin_stopWS'
 		}),
+		new web3._extend.Method({
+			name: 'getMaxPeers',
+			call: 'admin_getMaxPeers'
+		}),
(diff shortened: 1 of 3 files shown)


manav2401 (commented):
Is it possible to have this under `debug` namespace. I believe these functions suit that more.

0xsharma:
> Is it possible to have this under `debug` namespace. I believe these functions suit that more.

@manav2401 , yes this can be added under `debug` namespace. However, node related apis that can control the client should be kept under privateAdminAPI. Admin namespace already consists of node related calls like peers, addPeer, addTrustedPeer and setMaxPeers should also be added there as well.

temaniarpit27 (approved):

pratikspatil024 (approved):

PR #727, opened 2023-02-08
Added 2 flags to enable parallel EVM and set the number of speculative processes

pratikspatil024:


builder/files/config.toml
```diff
@@ -136,3 +136,7 @@ syncmode = "full"
 # [developer]
   # dev = false
   # period = 0
+
+# [parallelevm]
+  # enable = false
+  # specprocs = 8
```
cffls, line 142: Let's use `procs` or `cores`, which is more straightforward.

cffls (approved):

builder/files/config.toml
pratikspatil024: Done. Changed to `procs`.

PR #731, opened 2023-02-08
Add : mutex pprof profile

0xsharma:
# Description

In this PR, we added mutex profile generation using pprof.

### DIFF
--- a/internal/cli/debug_pprof.go
+++ b/internal/cli/debug_pprof.go
@@ -129,8 +129,10 @@ func (d *DebugPprofCommand) Run(args []string) int {
 
 	// Only take cpu and heap profiles by default
 	profiles := map[string]string{
-		"heap": "heap",
-		"cpu":  "cpu",
+		"heap":  "heap",
+		"cpu":   "cpu",
+		"trace": "trace",
+		"mutex": "mutex",
 	}
 
 	if !d.skiptrace {
--- a/internal/cli/debug_pprof.go
+++ b/internal/cli/debug_pprof.go
@@ -131,7 +131,6 @@ func (d *DebugPprofCommand) Run(args []string) int {
 	profiles := map[string]string{
 		"heap":  "heap",
 		"cpu":   "cpu",
-		"trace": "trace",
 		"mutex": "mutex",
 	}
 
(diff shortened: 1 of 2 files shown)


internal/cli/debug_pprof.go
```diff
@@ -129,8 +129,10 @@ func (d *DebugPprofCommand) Run(args []string) int {
 
 	// Only take cpu and heap profiles by default
 	profiles := map[string]string{
-		"heap": "heap",
-		"cpu":  "cpu",
+		"heap":  "heap",
+		"cpu":   "cpu",
+		"trace": "trace",
```
manav2401, line 134: "trace" was removed from the default list of profiles because it takes too much space in the final output. That case is handled below at line 138.

0xsharma: Will be taken care when merging the PR.

0xsharma:
> Won't there be any implementation for this profiling in the switch case above? Sorry, I am not much aware of how it's done so asked.

@manav2401 , no `mutex` will go through the default case which is lookup. In, lookup it finds the profile with the same name and returns the result. Here is code snippet from pprof package. Also, it has been tested and I was able to get a mutex profile. ✅ 

```
// Lookup returns the profile with the given name, or nil if no such profile exists.
func Lookup(name string) *Profile {
	lockProfiles()
	defer unlockProfiles()
	return profiles.m[name]
}


func lockProfiles() {
	profiles.mu.Lock()
	if profiles.m == nil {
		// Initial built-in profiles.
		profiles.m = map[string]*Profile{
			"goroutine":    goroutineProfile,
			"threadcreate": threadcreateProfile,
			"heap":         heapProfile,
			"allocs":       allocsProfile,
			"block":        blockProfile,
			"mutex":        mutexProfile,
		}
	}
}

func unlockProfiles() {
	profiles.mu.Unlock()
}
```

internal/cli/debug_pprof.go
0xsharma: Resolved ✅

cffls (approved):

temaniarpit27 (approved):

JekaMas (approved):

PR #736, opened 2023-02-09
Check if block is nil to prevent panic

dkeysil:
# Description

Block from `BlockByNumberOrHash` function can be nil, so we need to check.

Code is the same as here: https://github.com/maticnetwork/bor/blob/c46aae23daf0a1a172e08f97dcf2d7159e33eae5/internal/ethapi/api.go

# Changes

# Nodes audience

# Checklist

## Testing

- [x] I have tested this code on production mainnet node

# Additional comments

Issue: https://github.com/maticnetwork/bor/issues/725

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -628,6 +628,10 @@ func (s *PublicBlockChainAPI) GetTransactionReceiptsByBlock(ctx context.Context,
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


manav2401 (approved):
LGTM. Thanks for the PR.

pratikspatil024 (approved):

PR #754, opened 2023-02-23
core: add : impossible reorg block dump

0xsharma:
# Description

In this PR, we add the functionality to dump the blocks data in `~/impossibleReorg-<TIME>.txt`. This will allow better observability to resolve impossible reorg issues in the future.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -18,6 +18,7 @@
 package core
 
 import (
+	"compress/gzip"
 	"errors"
 	"fmt"
 	"io"
@@ -25,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


Raneet10 (approved):

pratikspatil024 (approved):

core/blockchain.go
```diff
@@ -2191,6 +2193,20 @@ func (bc *BlockChain) reorg(oldBlock, newBlock *types.Block) error {
 	} else {
 		// len(newChain) == 0 && len(oldChain) > 0
 		// rewind the canonical chain to a lower point.
+		out := fmt.Sprintf("oldChain: %+v,\n\noldBlock : %+v,\n\nnewBlock : %+v", oldChain, oldBlock, newBlock)
```
JekaMas, line 2196: Maybe 2 separate dumps ready to be imperted without changes?

JekaMas (approved):

core/blockchain.go
0xsharma: Working on it.

core/blockchain.go
```diff
@@ -2191,6 +2193,35 @@ func (bc *BlockChain) reorg(oldBlock, newBlock *types.Block) error {
...
+			fmt.Println("Impossible reorg : Unable to get user home dir", "Error", err)
+		}
+		outPath := filepath.Join(home, "impossible-reorgs", fmt.Sprintf("%v-impossibleReorg", time.Now().Format(time.RFC3339)))
+
+		if _, err := os.Stat(outPath); errors.Is(err, os.ErrNotExist) {
+			err := os.MkdirAll(outPath, os.ModePerm)
+			if err != nil {
+				log.Error("Impossible reorg : Unable to create Dir", "Error", err)
+			}
+		}
```
JekaMas, line 2208: I believe we need `else if err != nil` here and log any other errors.
Also are we ok with moving forward in case of errors? Maybe we need not to write to files.

core/blockchain.go
```diff
@@ -2191,6 +2193,35 @@ func (bc *BlockChain) reorg(oldBlock, newBlock *types.Block) error {
...
+
+		err = os.WriteFile(filepath.Join(outPath, "oldchain.txt"), []byte(fmt.Sprintf("%+v", oldChain)), 0600)
```
JekaMas, line 2210: stop. we won't be able to import from that format

JekaMas: look for `func ExportChain(blockchain *core.BlockChain, fn string) error` and `func ImportChain(chain *core.BlockChain, fn string) error {` and similar functions in `cmd/utils/cmd.go`

core/blockchain.go
0xsharma: Sure.

core/blockchain.go
0xsharma: Sure, will move on this path going forward.

PR #787, opened 2023-03-21
v0.3.6 fix

cffls:
# Description

This PR includes two hot fixes for v0.3.6:
 1. Fix getting validator set in verifyCascadingFields. Getting validator set and checking it against a header was initially introduced in https://github.com/maticnetwork/bor/pull/768. However, the implementation had difficulty in getting correct validator set. This PR fixes the issue by providing the current header to get validator set calls, and reduced the runtime complexity.
 2. Cherry-picking https://github.com/maticnetwork/bor/pull/673 to reduce unnecessary log info regarding tx commits.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


miner/worker.go
```diff
@@ -1117,7 +1141,12 @@ func (w *worker) prepareWork(genParams *generateParams) (*environment, error) {
 	}
 	// Run the consensus preparation with the default or customized consensus engine.
 	if err := w.engine.Prepare(w.chain, header); err != nil {
-		log.Error("Failed to prepare header for sealing", "err", err)
+		switch err.(type) {
```
JekaMas, line 1144: Could we use errors.Is(As) here?

JekaMas (approved):

consensus/bor/heimdall/span/spanner.go
```diff
@@ -108,13 +108,13 @@ func (c *ChainSpanner) GetCurrentValidators(ctx context.Context, headerHash comm
 	gas := (hexutil.Uint64)(uint64(math.MaxUint64 / 2))
 
 	// block
-	blockNr := rpc.BlockNumberOrHashWithHash(headerHash, false)
+	// blockNr := rpc.BlockNumberOrHashWithHash(headerHash, false)
```
marcello33, line 111: Why is this commented out?

consensus/bor/bor.go
```diff
@@ -464,18 +464,11 @@ func (c *Bor) verifyCascadingFields(chain consensus.ChainHeaderReader, header *t
...
 	// Verify the validator list match the local contract
 	if IsSprintStart(number+1, c.config.CalculateSprint(number)) {
-		parentBlockNumber := number - 1
 
-		var newValidators []*valset.Validator
-
-		var err error
-
-		for newValidators, err = c.spanner.GetCurrentValidators(context.Background(), chain.GetHeaderByNumber(parentBlockNumber).Hash(), number+1); err != nil && parentBlockNumber > 0; {
-			parentBlockNumber--
```
marcello33, line 474: Just out of curiosity @cffls 
I understand the bug about potentially going backwards indefinitely because of `parentBlockNumber--`
But why were we getting `nil` headers with `chain.GetHeaderByNumber` and we have to use `rpc` now?

consensus/bor/bor.go
```diff
@@ -464,18 +464,11 @@ func (c *Bor) verifyCascadingFields(chain consensus.ChainHeaderReader, header *t
...
-		}
+		newValidators, err := c.spanner.GetCurrentValidatorsByBlockNrOrHash(context.Background(), rpc.BlockNumberOrHashWithNumber(rpc.LatestBlockNumber), number+1)
 
 		if err != nil {
-			return errUnknownValidators
+			return err
 		}
 
 		sort.Sort(valset.ValidatorsByAddress(newValidators))
```
manav2401, line 474: @0xsharma and me were discussing if "sorting" the validators available locally and those available in header's extra data makes sense or not. Not sure if that was already discussed or not but just posting in here for visibility. 

I think we should compare the validator set as it is without sorting. Wdyt?

miner/worker.go
cffls: This comes from cherry-picking the commit, but we can try `errors.Is` here.

consensus/bor/heimdall/span/spanner.go
cffls: Good catch, this should be deleted.

consensus/bor/bor.go
cffls: It is still not clear to me why passing explicit block number won't work here. The header was looked up by this function: https://github.com/maticnetwork/bor/blob/3ed22568fd39b4e690ed45a90d9cd44370d01e26/eth/api_backend.go#L65-L76. For some some reason `b.eth.blockchain.GetHeaderByNumber(uint64(number))` will sometimes return nil, but `b.eth.blockchain.CurrentBlock().Header()` won't.

marcello33: Yeah, definitely strange

temaniarpit27:
Package version upgrade needed

marcello33 (approved):

hzj-edu-nju:
After this change, fast sync cannot be used。
why not use a better way。
Example: pass heimdall check

marcello33:
Hi @hzj-edu-nju 
The fix should actually solve the sync problem detected on v0.3.6

hzj-edu-nju:
You are right. One PR for one issue.
 Maybe I should create another issue for fast-sync problem

PR #793, opened 2023-03-28
Refactored the code and updated the miner to check for the validity of options

pratikspatil024:
- Updated the spec for the conditional transaction (EIP-4337), added more checks.
- Added checks in the miner module to validate the `options` while including the (bundled) transaction in the block.
- Added unit tests

References:
- https://github.com/OffchainLabs/nitro/tree/send-raw-transaction-conditional
- https://github.com/OffchainLabs/go-ethereum/tree/send-raw-transaction-conditional
- https://github.com/tynes/go-ethereum/tree/eip4337


JekaMas (commented):
What was the goal if the task?

pratikspatil024:
@JekaMas the spec here (https://notes.ethereum.org/@yoav/SkaX2lS9j) was updated, so just updating the code accordingly.
Also, made some changes in the miner module to validate the knownAccounts.

core/state/statedb.go
```diff
@@ -1046,3 +1047,30 @@ func (s *StateDB) AddressInAccessList(addr common.Address) bool {
 func (s *StateDB) SlotInAccessList(addr common.Address, slot common.Hash) (addressPresent bool, slotPresent bool) {
 	return s.accessList.Contains(addr, slot)
 }
+
+func (s *StateDB) ValidateKnownAccounts(knownAccounts types.KnownAccounts) error {
+	if knownAccounts == nil {
+		return fmt.Errorf("knownAccounts cannot be nil")
```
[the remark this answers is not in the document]
pratikspatil024: Agree. Will do that. Thanks.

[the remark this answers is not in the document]
pratikspatil024: [Done](https://github.com/maticnetwork/bor/pull/793/commits/aa7780bc9f81aef315491c8b40ae087bb38c59d3)

cffls (commented):
The implementation looks good. Could you please add some tests?

pratikspatil024:
@cffls yes, I'll add them today. Thank you.

miner/worker.go
```diff
@@ -1001,6 +1001,35 @@ func (w *worker) commitTransactions(env *environment, txs *types.TransactionsByP
 		//
 		// We use the eip155 signer regardless of the current hf.
 		from, _ := types.Sender(env.signer, tx)
+
+		// not prioritising conditional transaction, yet.
+		if options := tx.GetOptions(); options != nil {
+			if options.BlockNumberMax != big.NewInt(0) {
```
JekaMas, line 1007: Could we use https://pkg.go.dev/math/big#Int.Cmp rather then equality with ref type?

core/types/transactionConditional.go
```diff
@@ -0,0 +1,63 @@
...
+		return ErrEmptyKnownAccounts
+	}
+
+	length := 0
+
+	for _, v := range ka {
+		// check if the value is hex string or an object
+		if _, ok := v.(string); ok {
+			length += 1
+		} else if object, ok := v.(map[string]interface{}); ok {
```
JekaMas, line 51: how many levels it might be? only 2 or more?

pratikspatil024: Yes, just 2

miner/worker.go
pratikspatil024: Done, thanks.

PR #794, opened 2023-03-28
core/vm, miner : Make commitInterrupt OPCODE level

0xsharma:
# Description

In this PR, we make `CommitInterruptExperiment` (#556)  to OPCODE level.


core/state_processor.go
```diff
@@ -92,13 +92,14 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	return receipts, allLogs, *usedGas, nil
 }
 
-func applyTransaction(msg types.Message, config *params.ChainConfig, bc ChainContext, author *common.Address, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM) (*types.Receipt, error) {
+// nolint : unparam
+func applyTransaction(msg types.Message, config *params.ChainConfig, bc ChainContext, author *common.Address, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM, interruptCh *chan struct{}) (*types.Receipt, error) {
```
JekaMas, line 96: Do we need interrupt channel or context and check ctx.Done() then?

core/state_prefetcher.go
```diff
@@ -89,6 +89,6 @@ func precacheTransaction(msg types.Message, config *params.ChainConfig, gaspool
 	// Update the evm with the new transaction context.
 	evm.Reset(NewEVMTxContext(msg), statedb)
 	// Add addresses to access list if applicable
-	_, err := ApplyMessage(evm, msg, gaspool)
+	_, err := ApplyMessage(evm, msg, gaspool, nil)
```
JekaMas, line 92: Do we need to pass nil everywhere? '...' (ellipsis) doesn't require that.

core/state_processor.go
```diff
@@ -92,17 +93,22 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	return receipts, allLogs, *usedGas, nil
 }
 
-func applyTransaction(msg types.Message, config *params.ChainConfig, bc ChainContext, author *common.Address, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM) (*types.Receipt, error) {
+// nolint : unparam
+func applyTransaction(msg types.Message, config *params.ChainConfig, bc ChainContext, author *common.Address, gp *GasPool, statedb *state.StateDB, blockNumber *big.Int, blockHash common.Hash, tx *types.Transaction, usedGas *uint64, evm *vm.EVM, interruptCtx context.Context) (*types.Receipt, error) {
```
JekaMas, line 97: Why it with context rather that ...func() error?

accounts/abi/bind/backends/simulated.go
```diff
@@ -638,7 +638,7 @@ func (b *SimulatedBackend) callContract(ctx context.Context, call ethereum.CallM
 	vmEnv := vm.NewEVM(evmContext, txContext, stateDB, b.config, vm.Config{NoBaseFee: true})
 	gasPool := new(core.GasPool).AddGas(math.MaxUint64)
 
-	return core.NewStateTransition(vmEnv, msg, gasPool).TransitionDb()
+	return core.NewStateTransition(vmEnv, msg, gasPool).TransitionDb(context.TODO())
```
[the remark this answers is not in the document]
0xsharma: Yes sure !

core/state_prefetcher.go
0xsharma: Resolved ✅

core/state_processor.go
0xsharma: Trying this way with the context usage. If required will convert to ...func().

JekaMas (approved):

core/vm/interpreter.go
```diff
@@ -107,14 +121,182 @@ func NewEVMInterpreter(evm *EVM, cfg Config) *EVMInterpreter {
...
+	// explicit STOP, RETURN or SELFDESTRUCT is executed, an error occurred during
+	// the execution of one of the operations or until the done flag is set by the
+	// parent context.
+	for {
+		if interruptCtx != nil {
+			// case of interrupting by timeout
+			select {
+			case <-interruptCtx.Done():
+				opcodeCommitInterruptCounter.Inc(1)
+				log.Warn("OPCODE Level interrupt")
```
cffls, line 219: Since this is an expected behavior, should we change the log level to debug or info?

0xsharma: This is a rare log and something that would let us know that blocks are being partially processed, could be due to heavy txs. If a user sees this frequently, this can indicate that they might need to upgrade their hardware. This is the reason I have increased the verbosity of this log.

cffls (approved):

PR #804, opened 2023-04-03
Optimized the dependency metadata structure

pratikspatil024:
In this PR, I removed the first 2 elements in the TxDependency structure: the Transaction index and flag informing about delaying the fee transfer.

### DIFF
--- a/core/parallel_state_processor.go
+++ b/core/parallel_state_processor.go
@@ -319,7 +319,7 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 			shouldDelayFeeCal = false
 		}
 
-		if len(header.TxDependency) > 0 {
+		if len(header.TxDependency) != len(block.Transactions()) {
 			task := &ExecutionTask{
 				msg:               msg,
 				config:            p.config,
--- a/core/parallel_state_processor.go
+++ b/core/parallel_state_processor.go
@@ -297,7 +297,7 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 
 	coinbase, _ := p.bc.Engine().Author(header)
 
-	deps, delayMap := GetDeps(block.Header().TxDependency)
+	deps := GetDeps(block.Header().TxDependency)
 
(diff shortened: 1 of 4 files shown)


core/parallel_state_processor.go
```diff
@@ -315,9 +315,11 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 
 		cleansdb := statedb.Copy()
 
-		if len(header.TxDependency) > 0 {
-			shouldDelayFeeCal = delayMap[i]
+		if msg.From() == coinbase {
+			shouldDelayFeeCal = false
+		}
 
+		if len(header.TxDependency) > 0 {
```
cffls, line 322: Should we check if `len(header.TxDependency) == len(tasks)`? If lengths mismatch, we shouldn't use any dependency data.

miner/worker.go
```diff
@@ -1133,30 +1133,11 @@ func (w *worker) commitTransactions(env *environment, txs *types.TransactionsByP
...
-				tempDeps[i] = []uint64{uint64(i)}
-
-				reads := mvReadMapList[i-1]
-
-				_, ok1 := reads[blockstm.NewSubpathKey(env.coinbase, state.BalancePath)]
-				_, ok2 := reads[blockstm.NewSubpathKey(common.HexToAddress(w.chainConfig.Bor.CalculateBurntContract(env.header.Number.Uint64())), state.BalancePath)]
-
-				if ok1 || ok2 {
-					// 0 -> delay is not allowed
-					tempDeps[i] = append(tempDeps[i], 0)
```
cffls, line 1154: I think we should still keep this check. When delay is not allowed, we simply set dependency to empty, so validators will directly run serial execution.

core/parallel_state_processor.go
pratikspatil024: Nice idea. Thanks!

miner/worker.go
pratikspatil024: Yes, that will be nice.

cffls (approved):

PR #805, opened 2023-04-04
added a write lock to the txs.filter method and a read lock to the tx…

thogard785:
The filter method of txSortedMap is typically called by caller that holds the "m" RWLock object.

During reorgs, however, a Filter method of txList calls the filter method of txSortedMap but does not engage the lock. 

This Filter method is called only by the txPool methods demoteUnexecutables and promoteExecutables, which are only called during reorgs (which explains the low frequency with which this crash occurs). 

# Description

Added in a read lock and a write lock

# Changes

# Nodes audience

This PR is in response to a validator crash

# Checklist

# Additional comments

Have not had time to test this yet but it seemed urgent.

### DIFF
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -506,8 +506,12 @@ func (l *txList) Filter(costLimit *uint256.Int, gasLimit uint64) (types.Transact
 				lowest = nonce
 			}
 		}
+		l.txs.m.Lock()
 		invalids = l.txs.filter(func(tx *types.Transaction) bool { return tx.Nonce() > lowest })
+		l.txs.m.Unlock()
 	}
+	l.txs.m.RLock()
+	defer l.txs.m.RUnlock()
 	l.txs.reheap()
 	return removed, invalids
 }


core/tx_list.go
```diff
@@ -506,8 +506,12 @@ func (l *txList) Filter(costLimit *uint256.Int, gasLimit uint64) (types.Transact
 				lowest = nonce
 			}
 		}
+		l.txs.m.Lock()
```
JekaMas, line 509: Could you use Trylock? in case that the lock already happened.

thogard785: Line 492 calls the Filter() method which locks (and defers unlock) the lock in question so it should be safe to assume the lock isn't already held by this go routine (if it were, it'd deadlock on 492).  I'm not particularly familiar w/ TryLock though - should I still add it?

JekaMas (approved):

core/tx_list.go
```diff
@@ -506,8 +506,12 @@ func (l *txList) Filter(costLimit *uint256.Int, gasLimit uint64) (types.Transact
...
 		invalids = l.txs.filter(func(tx *types.Transaction) bool { return tx.Nonce() > lowest })
+		l.txs.m.Unlock()
 	}
+	l.txs.m.RLock()
+	defer l.txs.m.RUnlock()
 	l.txs.reheap()
```
JekaMas, line 515: Could we remove this Rlock and use that reheap code https://github.com/maticnetwork/bor/pull/806/files#diff-4aae684b15e6df1ff1c22f6845c43bd308a67dc43fc4fdec6f620979ac6941a9R157

manav2401:
Thanks @thogard785 for the PR. Have picked your commit against another branch to make a release. Please follow https://github.com/maticnetwork/bor/pull/807 for more details. Closing this for now. Thanks again!

PR #807, opened 2023-04-04
core: improve locks in txpool

manav2401:
# Description

This PR cherry-picks https://github.com/maticnetwork/bor/pull/805 and https://github.com/maticnetwork/bor/pull/806 which improves locks on a map used by transaction pool. 

Thanks and shout out to @thogard785 for helping in identifying and fixing this issue. 

Quoting the description from #805 by @thogard785
> The filter method of txSortedMap is typically called by caller that holds the "m" RWLock object.
> 
> During reorgs, however, a Filter method of txList calls the filter method of txSortedMap but does not engage the lock.
> 
> This Filter method is called only by the txPool methods demoteUnexecutables and promoteExecutables, which are only called during reorgs (which explains the low frequency with which this crash occurs).

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -173,9 +173,13 @@ func (m *txSortedMap) reheap(withRlock bool) {
 	heap.Init(&index)
 
 	if withRlock {
-		m.m.RLock()
-		m.index = &index
-		m.m.RUnlock()
+		m.m.Lock()
+	}
+
+	m.index = &index
+
+	if withRlock {
+		m.m.Unlock()
 	}
 
 	m.cacheMu.Lock()
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -150,19 +150,25 @@ func (m *txSortedMap) Filter(filter func(*types.Transaction) bool) types.Transac
 	removed := m.filter(filter)
 	// If transactions were removed, the heap and cache are ruined
 	if len(removed) > 0 {
-		m.reheap()
+		m.reheap(false)
 	}
 	return removed
 }
 
-func (m *txSortedMap) reheap() {
-	*m.index = make([]uint64, 0, len(m.items))
+func (m *txSortedMap) reheap(withRlock bool) {
+	index := make(nonceHeap, 0, len(m.items))
 
 	for nonce := range m.items {
-		*m.index = append(*m.index, nonce)
+		index = append(index, nonce)
 	}
 
-	heap.Init(m.index)
+	heap.Init(&index)
+
+	if withRlock {
+		m.m.RLock()
+		m.index = &index
+		m.m.RUnlock()
+	}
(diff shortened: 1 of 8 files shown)


pratikspatil024 (approved):

Raneet10 (approved):

marcello33 (approved):

kaber2:
Wow, you guys are truly the worst programmers I've seen in more than a decade. Conditional locking? Seriously? The proper solution is right before you eyes, but no, you merge garbage like this. OMG.

JekaMas:
> Wow, you guys are truly the worst programmers I've seen in more than a decade. Conditional locking? Seriously? The proper solution is right before you eyes, but no, you merge garbage like this. OMG.

I don't think this is helpful. Would love to see your solution with proper benchmarks and profiling.
By far I suggest you not using offensive language. I don't think that such things are welcome in open source.

core/tx_list.go
```diff
@@ -150,19 +151,38 @@ func (m *txSortedMap) Filter(filter func(*types.Transaction) bool) types.Transac
...
 	if len(removed) > 0 {
-		m.reheap()
+		m.reheap(false)
 	}
 	return removed
 }
 
-func (m *txSortedMap) reheap() {
-	*m.index = make([]uint64, 0, len(m.items))
+func (m *txSortedMap) reheap(withRlock bool) {
```
wjmelements, line 159: This is not good practice. Instead you should have two methods, `Reheap` and `reheap`, where `Reheap` wraps `reheap` and takes the lock.

core/tx_list.go
```diff
@@ -150,19 +151,38 @@ func (m *txSortedMap) Filter(filter func(*types.Transaction) bool) types.Transac
...
+	index := make(nonceHeap, 0, len(m.items))
+
+	if withRlock {
+		m.m.RLock()
+		log.Info("[DEBUG] Acquired lock over txpool map while performing reheap")
```
wjmelements, line 164: Use log.Debug

core/tx_list.go
```diff
@@ -521,12 +541,17 @@ func (l *txList) Filter(costLimit *uint256.Int, gasLimit uint64) (types.Transact
 				lowest = nonce
 			}
 		}
+
+		l.txs.m.Lock()
 		invalids = l.txs.filter(func(tx *types.Transaction) bool { return tx.Nonce() > lowest })
+		l.txs.m.Unlock()
```
wjmelements, line 547: This should be `l.txs.Filter` instead of locking outside

Issue #582, opened 2022-11-16
Wrong author data from bor_getAuthor during chain reorg


chaoyandy:
Hi,

We're experiencing the issue that author data returned from orphaned block during chain reorg. Though eventually the author data corrects by itself, but it is affecting the data correctness for live ingestion. What's worse, currently only `bor_getAuthor` API can return back author data,  but it's no where to tell which specific block(no block hash in both request/reply) it's associated to based on request param(`BlockNumber`) and return result(`Author` value). 

So in order to solve this, I found out a previous [commit](https://github.com/maticnetwork/bor/pull/435) that is adding correct miner address on API at RPC level but it's not been released yet. Is that possible to make this change into the next official build? Or will there be any other alternate solution? Thank you!

0xKrishna:
Yes this is a known issue, And we have already fixed that with the above mentioned commit.
It will be shipped with the next release for sure, Thanks for reporting.

Let me know if we can close this.

chaoyandy:
Thank you! We can close this then, btw, when probably will be the next release?

0xKrishna:
Most probably in Dec OR Jan (Too many holidays in Dec).

manav2401:
Hi @chaoyandy, so there has been some change with the PR you mentioned in your first comment (which adds miner field at rpc level). 

The change actually was supposed to go with an upcoming version but because the change is only on the RPC level, it breaks some downstream tooling applications which are relying on the equivalence of COINBASE opcode and response of the `eth_getBlock` endpoints. 

So, at the moment, this change has been reverted and won't be seen in upcoming releases. We'd be exploring other ways to achieve the same (as it would require consensus level changes) but I can't commit on any timeline for the same. 

We're also be exploring on the lines of what you suggested: adding some metadata in the `bor_getAuthor` endpoint like the block hash. This looks a doable addition and we'll surely think on this. Thanks for the suggestion. Are there any such more params (apart form block hash) which you think might be a good addition?

chaoyandy:
Thanks @manav2401 ! This is understandable and appreciate the follow-up. 

I think `block_hash` would be sufficient if you could add to the response of `bor_getAuthor`. Another way to resolve this issue is adding an API `bor_getAuthorByBlockHash` so that we can call this new API instead. 

Could you provide some ticket/item that I can track for fixing this issue? Or we can simply use this one?

chaoyandy:
Hello team/ @manav2401 , any updates on this?

[...]

PR #822, opened 2023-04-14
consensus/bor: modify bor RPC methods

Raneet10:
# Description

The PR aims to address #582 . The `bor` RPC methods `bor_getAuthor`, `bor_getSnapshotProposer` and `bor_getSnapshotProposerSequence` will now be invokable via block hash along with block number.

**Minor Note to the release conductor**:  Let's please add the feature changes to the API in the notes for the next release, albeit non-breaking.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### DIFF
--- a/consensus/bor/api.go
+++ b/consensus/bor/api.go
@@ -75,12 +75,34 @@ func rankMapDifficulties(values map[common.Address]uint64) []difficultiesKV {
 }
 
 // GetSnapshotProposerSequence retrieves the in-turn signers of all sprints in a span
-func (api *API) GetSnapshotProposerSequence(number *rpc.BlockNumber) (BlockSigners, error) {
-	snapNumber := *number - 1
+func (api *API) GetSnapshotProposerSequence(blockNrOrHash *rpc.BlockNumberOrHash) (BlockSigners, error) {
+	var (
+		snap       *Snapshot
+		snapNumber rpc.BlockNumber
+		err        error
+	)
+
+	if blockNrOrHash == nil {
+		header := api.chain.CurrentHeader()
+		snapNumber = rpc.BlockNumber(header.Number.Int64() - 1)
+		snap, err = api.GetSnapshot(&snapNumber)
+	} else {
+		if blockNr, ok := blockNrOrHash.Number(); ok {
+			snapNumber = blockNr - 1
+			if blockNr == rpc.LatestBlockNumber {
+				header := api.chain.CurrentHeader()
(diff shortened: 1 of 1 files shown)


consensus/bor/api.go
```diff
@@ -75,13 +75,37 @@ func rankMapDifficulties(values map[common.Address]uint64) []difficultiesKV {
...
+
+	//nolint:nestif
+	if blockNrOrHash == nil {
+		header = api.chain.CurrentHeader()
+		snapNumber = rpc.BlockNumber(header.Number.Int64() - 1)
+		snap, err = api.GetSnapshot(&snapNumber)
+	} else {
+		if blockNr, ok := blockNrOrHash.Number(); ok {
+			snapNumber = blockNr - 1
+			if blockNr == rpc.LatestBlockNumber {
```
manav2401, line 94: I believe this is done because rpc.LatestBlock is stored as (-1). Post milestone we might provide the "finalised" tag as well. Should we update this check to something like `if blockNr <= 0`?

consensus/bor/api.go
```diff
@@ -75,13 +75,37 @@ func rankMapDifficulties(values map[common.Address]uint64) []difficultiesKV {
...
+	var (
+		snap       *Snapshot
+		snapNumber rpc.BlockNumber
+		header     *types.Header
+		err        error
+	)
+
+	//nolint:nestif
+	if blockNrOrHash == nil {
+		header = api.chain.CurrentHeader()
```
manav2401, line 88: Do you think we should perform a `nil` header check here and at other few places below?

manav2401 (approved):
LGTM, just few minor comments.

consensus/bor/api.go
Raneet10: We could sure , though if someone passes `PendingBlockNumber` (-2) for instance we might end up giving an inaccurate response.

consensus/bor/api.go
Raneet10: Ahh I think I see what you mean : 
`snapNumber = rpc.BlockNumber(header.Number.Int64() - 1)`

marcello33:
Are the comments addressed? In such case, LGTM and I'd approve

marcello33 (approved):

temaniarpit27 (approved):

manav2401 (approved):

consensus/bor/api.go
manav2401: Alright, for now just checking for Latest sounds good.

PR #832, opened 2023-04-21
[WIP] core, miner, internal/cli : added interruptCommitFlag and cache

0xsharma:
# Description

Work in progress.

### DIFF
--- a/core/vm/interpreter.go
+++ b/core/vm/interpreter.go
@@ -26,13 +26,13 @@ import (
 	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/metrics"
-
 	lru "github.com/hashicorp/golang-lru"
 )
 
 var (
 	opcodeCommitInterruptCounter = metrics.NewRegisteredCounter("worker/opcodeCommitInterrupt", nil)
 	ErrInterrupt                 = errors.New("EVM execution interrupted")
+	NoCache                      = errors.New("no tx cache found")
 )
 
 const (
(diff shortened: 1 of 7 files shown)


core/vm/interpreter.go
```diff
@@ -34,8 +36,18 @@ var (
...
 	InterruptCtxOpcodeDelayKey = "opcodeDelay"
+
+	// InterruptedTxCacheSize is size of lru cache for interrupted txs
+	InterruptedTxCacheSize = 90000
+
+	// InterruptedTxCacheKey gets you the pointer to lru cache storing the interrupted txs
+	InterruptedTxCacheKey = "interruptedTxCache"
+
+	// InterruptedTxContext_currenttxKey gets you the hash of the current tx being executed
+	InterruptedTxContext_currenttxKey = "currenttx"
```
JekaMas, line 50: Could you make context key private and rewrire the code in a way like https://go.dev/play/p/2YNOSVTioB8

core/vm/interpreter.go
```diff
@@ -215,10 +227,20 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool, i
...
 			case <-interruptCtx.Done():
-				opcodeCommitInterruptCounter.Inc(1)
-				log.Warn("OPCODE Level interrupt")
+				txHash := interruptCtx.Value(InterruptedTxContext_currenttxKey).(common.Hash)
+				interruptedTxCache := interruptCtx.Value(InterruptedTxCacheKey).(*lru.Cache)
+
+				// if the tx is already in the cache, it means that it has been interrupted before and we will not interrupt it again
+				found, _ := interruptedTxCache.ContainsOrAdd(txHash, true)
+				if found {
+					interruptedTxCache.Remove(txHash)
```
cffls, line 236: I might be wrong, but if we allow a previously interrupted transaction to continue its execution when the second time we see it, the node might end up with producing empty block again, and the transaction will still not be included in the block. Wouldn't this defeat the original purpose of interrupt commit?

0xsharma: Moving this conversation to internal slack chat.

cffls (approved):

PR #841, opened 2023-05-01
Added check for circular and out-of-range dependency problem

pratikspatil024:
POS-1391 and POS-1392


core/parallel_state_processor.go
```diff
@@ -439,3 +444,28 @@ func GetDeps(txDependency [][]uint64) map[int][]int {
...
+	// number of transactions in the block
+	n := len(deps)
+
+	// Handle out-of-range and circular dependency problem
+	for tx, val := range deps {
+		for depTx := range val {
+			// out-of-range
+			if depTx >= n {
+				return false
+			}
```
cffls, line 459: If we also enforce `depTx < tx` here, the circular dependency could be avoided, and we won't need the loop below. wdyt?

pratikspatil024: Good point, Jerry. I agree.

pratikspatil024: [Done](https://github.com/maticnetwork/bor/pull/841/commits/c7334d3f0afc99c71daa9815a31cfa8898d73d9d).

core/parallel_state_processor.go
```diff
@@ -439,3 +444,20 @@ func GetDeps(txDependency [][]uint64) map[int][]int {
...
+
+// returns true if dependencies are correct
+func VerifyDeps(deps map[int][]int) bool {
+	// number of transactions in the block
+	n := len(deps)
+
+	// Handle out-of-range and circular dependency problem
+	for tx, val := range deps {
+		for depTx := range val {
+			if depTx >= n && depTx < tx {
```
cffls, line 456: Should this be `||` instead of `&&`?

core/parallel_state_processor.go
```diff
@@ -304,6 +304,11 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 		metadata = true
 	}
 
+	if !VerifyDeps(deps) {
```
cffls, line 307: If deps is nil or empty, we can skip the check.

core/parallel_state_processor.go
pratikspatil024: My bad, [done](https://github.com/maticnetwork/bor/pull/841/commits/8fe290164e80634a5c9776464f0e99aa2e86ba5e) thanks!

core/parallel_state_processor.go
pratikspatil024: [Done](https://github.com/maticnetwork/bor/pull/841/commits/8fe290164e80634a5c9776464f0e99aa2e86ba5e), thanks.

cffls (approved):

PR #851, opened 2023-05-05
fix : go-routine leak in commitInterrupt channel

0xsharma:
# Description

In this PR, we fix the go-routine leak identified in the code.


marcello33 (approved):

manav2401 (approved):

cffls (commented):
Could we add a goleak test for commit interrupt, similar to what we did [here](https://github.com/maticnetwork/bor/blob/e36a54d60021a3845ce23d58580194f1cd0d9285/accounts/abi/bind/backends/simulated_test.go#L44)?

PR #857, opened 2023-05-10
mardizzone/backport-master-to-develop

marcello33:
# Description

Backport changes from `master` to `develop` after `v0.3.9` release

# Changes

# Checklist

### DIFF
--- a/docs/cli/server.md
+++ b/docs/cli/server.md
@@ -208,11 +208,11 @@ The ```bor server``` command runs the Bor client.
 
 - ```netrestrict```: Restricts network communication to the given IP networks (CIDR masks)
 
-- ```nodekey```:  P2P node key file
+- ```maxpendpeers```: Maximum number of pending connection attempts (defaults used if set to 0)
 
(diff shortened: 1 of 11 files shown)


docs/cli/server.md
```diff
@@ -208,11 +208,11 @@ The ```bor server``` command runs the Bor client.
 
 - ```netrestrict```: Restricts network communication to the given IP networks (CIDR masks)
 
-- ```nodekey```:  P2P node key file
+- ```maxpendpeers```: Maximum number of pending connection attempts (defaults used if set to 0)
 
-- ```nodekeyhex```: P2P node key as hex
+- ```nat```: NAT port mapping mechanism (any|none|upnp|pmp|extip:<IP>)
```
pratikspatil024, line 213: I think this is some error, because `maxpendpeers` and `nat` are already defined above, and not sure why `nodekey` and `nodekeyhex` is removed.

marcello33:
Thanks @pratikspatil024 
Will fix this and the failing tests, then re-raise it

PR #871, opened 2023-05-19
Change legacy rando geth fork choice to be deterministic

paulgoleary:
Opening for discussion. `ReorgNeeded` is part of the legacy geth-based fork choice rules but I assume it's still what's used. If so, it has this strange coin-flip last tie breaker rule. Not sure how often this is fired but if it is it would contribute to unnecessary forks and re-orgs. Trivial and safe to change and does not require a HF.
Change was safe / successfully at previous company FWIW.

# Description

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


cffls:
Will this change enable validators to win a fork by mining blocks with lower hash value?

paulgoleary:
The history of this implementation is related to concerns about selfish mining / hash-grinding attacks. But it's important to note that it has always been the final 'tie breaker' after other fork choice rules have been applied, so difficulty still dominates the choice.
So with Bor (Clique-derived) consensus it's hard to see how a validator could gain an advantage by trying to game this. Since difficulty is always won by the primary, the only possible attack would be for a non-primary producer to somehow engineer a tie in difficulty and then grind out a lower hash to get their fork chosen.

JekaMas:
It might be that yhis change willbe safe and useful after introducing finality.

cffls (approved):

pratikspatil024 (approved):

0xsharma (approved):

manav2401 (approved):

marcello33 (approved):

ssandeep (approved):

PR #896, opened 2023-06-12
Use `big.Int#Set` instead of `big.Int#SetBytes` when possible

maoueh:
There is no reason to make the bytes transformation and back to big int to assign a big int to another one.

The `Set` method should be used instead.

# Description

# Changes

- [X] Improvements non-breaking


cffls:
Thanks for the improvement!

cffls (approved):

marcello33 (approved):

PR #899, opened 2023-06-13
Fix worker `nil` pointer exception because of `baseFee == nil`

maoueh:
# Description

The `baseFee` can be `nil` in certain situations, the worker must not convert `baseFee` to `*big.Int` unless it was checked otherwise if such case happen, Go is going to panic:

```
goroutine 115 [running]:
math/big.(*Int).Bits(...)
	/opt/homebrew/Cellar/go/1.20.4/libexec/src/math/big/int.go:105
github.com/holiman/uint256.(*Int).SetFromBig(0x14001297410?, 0x107714520?)
	/Users/maoueh/go/pkg/mod/github.com/holiman/uint256@v1.2.0/conversion.go:102 +0x24
github.com/holiman/uint256.FromBig(...)
	/Users/maoueh/go/pkg/mod/github.com/holiman/uint256@v1.2.0/conversion.go:50
github.com/ethereum/go-ethereum/common/math.FromBig(0x1073923e0?)
	/Users/maoueh/work/sf/ethereum.chain/common/math/uint.go:20 +0x34
github.com/ethereum/go-ethereum/miner.(*worker).fillTransactions.func4({0x1400015c080?, 0x107827f08?}, {0x107836a40, 0x14000730be0})
	/Users/maoueh/work/sf/ethereum.chain/miner/worker.go:1521 +0x4c
github.com/ethereum/go-ethereum/common/tracing.Exec({0x107827f08, 0x140011b7020}, {0x0, 0x0}, {0x1066bde48, 0x27}, {0x140012975c0, 0x1, 0x0?})
	/Users/maoueh/work/sf/ethereum.chain/common/tracing/context.go:70 +0x1e4
github.com/ethereum/go-ethereum/miner.(*worker).fillTransactions(0x14000192780, {0x107827f08?, 0x140011b68d0?}, 0x1400060914c, 0x140002d4000, {0x107827e98, 0x1400005c0e8})
	/Users/maoueh/work/sf/ethereum.chain/miner/worker.go:1520 +0x1a4
github.com/ethereum/go-ethereum/miner.(*worker).commitWork(0x14000192780, {0x107827f08, 0x140011b6240}, 0x0?, 0x0, 0x6488d5f2)
	/Users/maoueh/work/sf/ethereum.chain/miner/worker.go:1663 +0x3e8
github.com/ethereum/go-ethereum/miner.(*worker).mainLoop(0x14000192780, {0x107827f08, 0x14000137290})
	/Users/maoueh/work/sf/ethereum.chain/miner/worker.go:603 +0x7e4
created by github.com/ethereum/go-ethereum/miner.newWorker
	/Users/maoueh/work/sf/ethereum.chain/miner/worker.go:344 +0x7e0
```

# Changes

## Testing

- [ ] I have added unit tests
- [X] I have tested this code manually on local environment

### Manual tests

Ran a clique consensus engine mining node using Polygon codebase and send transactions to it, looking for the transactionReceipt of everything that has been pushed to chain.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -32,6 +32,7 @@ import (
 
 	mapset "github.com/deckarep/golang-set"
 	lru "github.com/hashicorp/golang-lru"
+	"github.com/holiman/uint256"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
@@ -672,7 +673,12 @@ func (w *worker) mainLoop(ctx context.Context) {
 					txs[acc] = append(txs[acc], tx)
 				}
 
-				txset := types.NewTransactionsByPriceAndNonce(w.current.signer, txs, cmath.FromBig(w.current.header.BaseFee))
+				var baseFee *uint256.Int
+				if w.current.header.BaseFee != nil {
+					baseFee = cmath.FromBig(w.current.header.BaseFee)
+				}
+
+				txset := types.NewTransactionsByPriceAndNonce(w.current.signer, txs, baseFee)
 				tcount := w.current.tcount
 
 				//nolint:contextcheck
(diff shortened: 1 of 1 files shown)


cffls:
Thanks for the fix! Could you provide some insights about when the basefee will be nil? Geth doesn't have this check: https://github.com/ethereum/go-ethereum/blob/master/miner/worker.go#L541. I am wondering if this issue is specific to polygon.

maoueh:
I maintain Firehose patch for Polygon and we uses `clique` engine to populate a chain pre-London. For me it happened in this case.

> txset := types.NewTransactionsByPriceAndNonce(w.current.signer, txs, w.current.header.BaseFee)

This code is different, the bug happen here because of the use of `cmath.FromBig`, in upstream `geth` as you can see their `NewTransactionsByPriceAndNonce` receives a `*big.Int` (https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go#L569) while in Polygon case, it receives a `*uint256.Int` (https://github.com/maticnetwork/bor/blob/891ec7fef619cac0a796e3e29d5b2d4c095bdd9b/core/types/transaction.go#L643) so the problem happen at the conversion level.

It seems `geth` is different, not sure when the differences happened between you and `geth`.

cffls (approved):
Thanks @maoueh for the explanation. LGTM.

PR #903, opened 2023-06-21
eth/filters, ethclient/gethclient: add fullTx option to pending tx filter

lmittmann:
This PR is a cherry-pick of [go-ethereum#25186](https://github.com/ethereum/go-ethereum/pull/25186). It adds an option to subscribe to full pending transactions.

> This PR adds a way to subscribe to the _full_ pending transactions, as opposed to just being notified about hashes.
>
> In use cases where client subscribes to newPendingTransactions and gets txhashes only to then request the actual transaction, the caller can now shortcut that flow and obtain the transactions directly.

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -40,6 +40,7 @@ type filter struct {
 	typ      Type
 	deadline *time.Timer // filter is inactiv when deadline triggers
 	hashes   []common.Hash
+	txs      []*types.Transaction
 	crit     FilterCriteria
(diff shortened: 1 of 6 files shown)


manav2401:
Hey @lmittmann, thanks for the PR. We're already in process of merging the upstream geth changes (https://github.com/maticnetwork/bor/pull/901) and this change will be a part of it. We'd like to have this change as a part of that PR. Closing this for now, thanks!

PR #910, opened 2023-06-26
fix : runtime testcases

0xsharma:


core/state_processor.go
```diff
@@ -143,24 +141,8 @@ func applyTransaction(msg *Message, config *params.ChainConfig, gp *GasPool, sta
...
-
-		msg.From,
-		evm.Context.Coinbase,
-
-		result.FeeTipped,
-		result.SenderInitBalance,
-		coinbaseBalance,
-		output1.Sub(output1, result.FeeTipped),
-		output2.Add(output2, result.FeeTipped),
-	)
```
cffls, line 163: Need to double check the correctness of this. Tipping is already skipped in `ApplyMessageNoFeeBurnOrTip(evm, *msg, gp, interruptCtx)` (line 132) above. If we don't tip miner at this point, the miner won't get tipped at all. cc @pratikspatil024

Raneet10: Yeah I don't remember exactly why the logging was removed but I can restore it. Will it have an effect on miners not getting tip fee ?

cffls: Fee is tipped to miner in line 145. If lines from 147 to 163 are removed, the tipping log will be missing, which results in mismatched receipt hash.

Raneet10: Ahh yes makes sense!

PR #911, opened 2023-06-26
Fix Indore hardfork block in genesis-testnet-v4.json

cshintov:
Causes `bad block` errors after upgrade to 0.4.5-beta-5.

# Description

set `indoreBlock` to `37075456` in bor testnet genesis.

# Changes

- [X] Hotfix (change that solves an urgent issue, and requires immediate attention)

### Manual tests

The upgraded nodes work as expected without `bad block` errors.

# Additional comments

Fixes https://github.com/maticnetwork/bor/issues/906

### DIFF
--- a/builder/files/genesis-testnet-v4.json
+++ b/builder/files/genesis-testnet-v4.json
@@ -17,9 +17,9 @@
       "jaipurBlock": 22770000,
       "delhiBlock": 29638656,
       "parallelUniverseBlock": 0,
-      "indoreBlock": 36877056,
+      "indoreBlock": 37075456,
       "stateSyncConfirmationDelay": {
-        "36877056": 128
+        "37075456": 128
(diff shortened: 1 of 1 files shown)


pratikspatil024:
Hi, @cshintov thanks for the PR.

If you are using `v0.4.0-beta-5`, then the Indore block for Mumbai is already `37075456`. See [here](https://github.com/maticnetwork/bor/blob/v0.4.0-beta-5/builder/files/genesis-testnet-v4.json#L20).

If you are running a node, we recommend you look at the code corresponding to the particular release you are using. The code on the `develop` branch is not necessarily the same as that of the latest release because of the various processes we follow for the release.

cshintov:
Got it. Thanks for the info. Will help me in future upgrades.

PR #964, opened 2023-08-16
internal/cli: add db.engine flag

anshalshukla:
# Description

With recent changes in geth a new flag was introduced to select upon the database. We already have the required changes after upstream merge, I've just added the flag to select the database.

# Changes

# Breaking changes

N.A

# Nodes audience

N.A

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

I've tested it locally and it works well.

# Additional comments

[POS-1763](https://polygon.atlassian.net/browse/POS-1723)

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -67,6 +67,9 @@ type Config struct {
 	// Ancient is the directory to store the state in
 	Ancient string `hcl:"ancient,optional" toml:"ancient,optional"`
 
+	// DBEngine is used to select leveldb or pebble as database
+	DBEngine string `hcl:"dbengine,optional" toml:"dbengine,optional"`
+
 	// KeyStoreDir is the directory to store keystores
 	KeyStoreDir string `hcl:"keystore,optional" toml:"keystore,optional"`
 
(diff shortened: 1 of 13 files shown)


internal/cli/server/config.go
```diff
@@ -67,6 +67,9 @@ type Config struct {
 	// Ancient is the directory to store the state in
 	Ancient string `hcl:"ancient,optional" toml:"ancient,optional"`
 
+	// DBEngine is used to select leveldb or pebble as database
+	DBEngine string `hcl:"dbengine,optional" toml:"dbengine,optional"`
```
manav2401, line 71: You'll need to add `DBEngine` in the `DefaultConfig` function below. Use `leveldb` as default for now.

internal/cli/server/flags.go
```diff
@@ -53,6 +53,12 @@ func (c *Command) Flags() *flagset.Flagset {
 		Value:   &c.cliConfig.Ancient,
 		Default: c.cliConfig.Ancient,
 	})
+	f.StringFlag(&flagset.StringFlag{
+		Name:    "db.engine",
+		Usage:   "Backing database implementation to use ('leveldb' or 'pebble')",
+		Value:   &c.cliConfig.DBEngine,
+		Default: "leveldb",
```
manav2401, line 60: Use `Default: c.cliConfig.DBEngine` instead of hardcoded value.

manav2401 (changes requested):
Few nit picks before merging. 
1. Run `make docs` to update the docs at the end. 
2. Update the 9 `config.toml` files with this value (commented) (in `builder` and `packaging` modules). 
3. Update `example_config.toml` in `docs/cli`. 
4. Test it locally to confirm few things like - if pebble db is enabled for fresh node or not, if you start node with leveldb and restart it with pebble are the errors handled correctly or not, etc.

marcello33 (approved):

pratikspatil024 (approved):

0xsharma (approved):

PR #965, opened 2023-08-16
metrics/prometheus: fix count field of metric

manav2401:
# Description

With changes done in https://github.com/maticnetwork/bor/pull/881/ and https://github.com/maticnetwork/bor/pull/893/, the `count` field of a metric was added into summary and was calculated in wrong way. The tests didn't complain as they were updated in few places to cater to the new logic or the values according to the new logic vs old logic turned out to be same for the test case 

E.g. histogram count should be ideally 0 as no values are reported but it was changed to 6 which actually denotes the total percentile values reported. In case of timers, the values reported were 6 and the percentiles were also 6 so the test didn't catch this check. 

This PR updates (or actually reverts) the reporting of count metric. Instead of reporting count as `len(values)` where values are the total percentiles reported, we fetch it directly from `metric.Count()` function as earlier (and also similar to how geth does). 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [x] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/metrics/prometheus/collector.go
+++ b/metrics/prometheus/collector.go
@@ -76,7 +76,7 @@ func (c *collector) addHistogram(name string, m metrics.Histogram) {
 	}
 
 	c.writeSummarySum(name, fmt.Sprintf("%f", sum))
-	c.writeSummaryCounter(name, len(ps))
+	c.writeSummaryCounter(name, m.Count())
 	c.buff.WriteRune('\n')
 }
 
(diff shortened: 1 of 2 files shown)


marcello33 (approved):

pratikspatil024 (approved):

rekibnikufesin (approved):
:shipit:

PR #968, opened 2023-08-18
Geth Merge v1.12.2

temaniarpit27:


core/blockstm/mvhashmap.go
```diff
@@ -115,7 +115,6 @@ func (mv *MVHashMap) Write(k Key, v Version, data interface{}) {
 			rw: sync.RWMutex{},
 			tm: treemap.NewWithIntComparator(),
 		}
-		cells = n
```
pratikspatil024, line 118: Any reason why this was removed?

internal/cli/server/config.go
```diff
@@ -1106,8 +1105,6 @@ func (c *Config) buildEth(stack *node.Node, accountManager *accounts.Manager) (*
 		log.Debug("Sanitizing Go's GC trigger", "percent", int(gogc))
 		godebug.SetGCPercent(int(gogc))
 
-		n.TrieCleanCacheJournal = c.Cache.Journal
-		n.TrieCleanCacheRejournal = c.Cache.Rejournal
```
pratikspatil024, line 1110: I can see that these 2 flags have been removed by geth (in `cmd/utils/flag.go`), I think we should remove these flags from bor as well.

Also, in `cmd/utils/flag.go` I can see that some new flags have also been introduced, should we add them to bor now, or we can do that later?

pratikspatil024 (approved):

cffls (approved):

PR #977, opened 2023-08-29
miner/worker :: add : start commit work only after connecting to peers

0xsharma:
In this PR, we wait for a validator node to connect to atleast 1 peer before starting to produce blocks. This helps in stopping the validator to creating its own fork which might later result in the entire networking getting reorged to that fork. This PR tries to improve the stability of the network. 

This is only implemented to work on `mainnet` and `mumbai`. It will not work on devnets/testnets.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -351,6 +351,10 @@ func makeExtraData(extra []byte) []byte {
 	return extra
 }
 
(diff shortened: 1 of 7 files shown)


VAIBHAVJINDAL3012 (approved):

cffls (approved):
LGTM

PR #992, opened 2023-09-07
core/types: fix receipt EffectiveGasPrice unmarshaling

marcello33:
# Description

During some tests, I found the following error while getting main chain receipts: `math/big: cannot unmarshal "\"0x0\"" into a *big.Int`
The changes in this PR fixed the error. Without the fixes, `heimdall`, which uses `bor` as a dependency, would not be able to properly unmarshal the receipts while fetching all the events from L1 (stateSynced, topup, unstakeInit events etc...).
[This is present already in geth](https://github.com/ethereum/go-ethereum/blob/master/core/types/receipt.go#L82).

Also, added a `nil`  check for `EffectiveGasPrice` and implemented its unit tests. Missing that would break the `eth_getTransactionReceipt` correct functioning.

The current version undergoing tests on Mumbai (`v1.0.0-beta`) is also affected, but the error does not appear because the version of `bor` used in `heimdall` as dependency in `go.mod` is `v0.3.x`, where the field `EffectiveGasPrice` was not introduced yet.

As an advice, as `bor` is used as a dependency in `heimdall`, we should always keep the dependencies aligned, every time a new `bor` version is released.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### DIFF
--- a/core/types/gen_receipt_json.go
+++ b/core/types/gen_receipt_json.go
@@ -25,7 +25,7 @@ func (r Receipt) MarshalJSON() ([]byte, error) {
 		TxHash            common.Hash    `json:"transactionHash" gencodec:"required"`
 		ContractAddress   common.Address `json:"contractAddress"`
 		GasUsed           hexutil.Uint64 `json:"gasUsed" gencodec:"required"`
-		EffectiveGasPrice *big.Int       `json:"effectiveGasPrice"`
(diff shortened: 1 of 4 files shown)


marcello33:
Closing as the changes got included in v1.0.0-beta-candidate branch

PR #1008, opened 2023-09-16
internal/cli: add support to overwrite config.toml values via cli flags

manav2401:
# Description

With the new-cli introduced in v0.3.0, there were 2 ways to set the configuration required to run bor. One can either do it via normal cli flags (args) or via a toml based config. Due to some restrictions, if the toml based config is provided, bor overwrites all the given cli flags (explicitly by the user) with those provided in toml config. 

An ideal behaviour should in any cli application should be that explicitly provided flags should take control no matter what value is provided in the toml based config. This PR implements that. Sufficient tests have been added and existing tests have been improved to cover all cases for all data types. 

The order of flag will now be as follows:
1. If set in cli args, use that value
2. If not set in cli args and set in config, use that value
3. If not set in any of them, use default value

Example:
If the toml config looks something like this: 
```toml
chain = "mainnet"
identity = "Anon"
datadir = "config_datadir"
### More values
```
and the user runs the following command
```
bor server --config=config.toml --datadir=cli_datadir
```

Currently bor would set the `datadir` the value mentioned in config i.e. `config_datadir`. 

With this PR, the value set for `datadir` would be `cli_datadir`. 

This is a breaking change as it changes the way people run their bor nodes. It mostly affects those who run a combination of both. Until now, their cli flags weren't set if config was set but now onwards, their additional flags will also be set. This might require an explicit communication to node operators. 

# Changes

- [x] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

As mentioned above, it changes the way flags are parsed and handled. Cli args will overwrite config flags. 

# Checklist

- [x] Created a task in Jira and informed the team for implementation in Erigon client (if applicable)

## Testing

- [x] I have added unit tests

### Manual tests

Manual testing of flags with different datatypes have been carried to make sure it works for all types of values.

### DIFF
--- a/internal/cli/dumpconfig.go
+++ b/internal/cli/dumpconfig.go
@@ -42,7 +42,7 @@ func (c *DumpconfigCommand) Synopsis() string {
 func (c *DumpconfigCommand) Run(args []string) int {
 	// Initialize an empty command instance to get flags
 	command := server.Command{}
-	flags := command.Flags()
+	flags := command.Flags(nil)
 
 	if err := flags.Parse(args); err != nil {
 		c.UI.Error(err.Error())
--- a/internal/cli/flagset/flagset.go
+++ b/internal/cli/flagset/flagset.go
@@ -4,13 +4,10 @@ import (
 	"flag"
 	"fmt"
 	"math/big"
-	"reflect"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
-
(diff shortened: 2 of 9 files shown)


cffls (approved):

temaniarpit27 (approved):

pratikspatil024 (approved):
Noice!

PR #1025, opened 2023-10-05
Shanghai/Agra HF

0xsharma:
# Description

Shanghai HF on Ethereum was done using timestamp rather than blockNumber. On POS, we would continue with blockNumber based hardforks. 

In this PR, we convert the timestamp based HF to blockNumber based. Also, certain beacon chain related changes specefic to withdrawals( not applicable to Polygon-POS ) will also be removed.


marcello33:
Thanks @0xsharma 
As discussed, once PR is ready, please make sure to delete the `TODO marcello double check` comments that I left there at time of upstream merge to make such changes related to the Shanghai HF.

temaniarpit27 (approved):

PR #1109, opened 2023-12-18
Removed the check for milestoneID in the GetVoteOnHash()

VAIBHAVJINDAL3012:
# Description

Removed the check for milestoneID in the GetVoteOnHash() function, in present system there is an extra check which query the authenticity of the milestone being voted by querying the milestoneID from the Heimdall.
It seemed to be an extra check which extra call overhead and sometime also causing delay.

# Changes

- [.] Bugfix (non-breaking change that solves an issue)


Raneet10 (approved):

pratikspatil024 (approved):

marcello33 (approved):

marcello33:
If we need to merge, please provide a bit more context and remove the [WIP] from the PR title, thanks.

CitizenSearcher:
I reported an issue related to this section that is causing block import delays on clients, and may also delay block production on validators.

https://github.com/maticnetwork/bor/issues/1118

PR #1120, opened 2024-01-05
fix bug: should use Lock when mutating the flag

zhiqiangxu:
1. hold the write lock while mutating
2. hold the read lock while reading, otherwise the result may be inconsistent(partially updated)


cffls (approved):

marcello33 (approved):

pratikspatil024 (approved):

pratikspatil024:
Cherry-picked the 3 commits from @zhiqiangxu into https://github.com/maticnetwork/bor/pull/1141.

PR #1129, opened 2024-01-17
Chg: Changed the maximum automatic rewind length from 256 to 126

VAIBHAVJINDAL3012:
# Description
Have changed the automatic rewind length in the situation of checkpoint or milestone mismatch 
from 256 to 126 as we have TriesInMemory size of 128. Keeping it 256 could have unexpected output.


temaniarpit27:
Closing as this change has been picked in v1.2.4-beta-candidate and will be back merged to develop

PR #1149, opened 2024-01-29
fix: bootnode

anshalshukla:
# Description

This fixes the bootnodes issue due to which the port mapping was getting done properly. Added a helper function to export the logic from bootnodes binary, this won't require any update on bootnodes flag function incase there is a change from upstream.

# Changes

# Checklist

- [X] I have added at least 2 reviewer or the whole pos-v1 team
- [X] I have added sufficient documentation in code
- [X] I will be resolving comments - if any - by pushing each fix in a separate commit and linking the commit hash in the comment reply
- [X] Created a task in Jira and informed the team for implementation in Erigon client (if applicable)

## Testing

- [ ] I have added unit tests
- [X] I have tested this code manually on local environment

### Manual tests

I tried connecting two nodes using the bootnode.

### DIFF
--- a/cmd/bootnode/main.go
+++ b/cmd/bootnode/main.go
@@ -15,7 +15,8 @@
 // along with go-ethereum. If not, see <http://www.gnu.org/licenses/>.
 
 // bootnode runs a bootstrap node for the Ethereum Discovery Protocol.
-package main
+// Keep package as bootnode during upstram merge.
+package bootnode
 
 import (
 	"crypto/ecdsa"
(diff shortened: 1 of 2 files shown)


cmd/bootnode/main.go
```diff
@@ -15,7 +15,8 @@
 // along with go-ethereum. If not, see <http://www.gnu.org/licenses/>.
 
 // bootnode runs a bootstrap node for the Ethereum Discovery Protocol.
-package main
+// Keep package as bootnode during upstram merge.
+package bootnode
```
marcello33, line 19: Why is this enforced?

marcello33:
Can you please spend a few words about the changes in the PR description?
Thanks

0xsharma (approved):
LGTM !

cmd/bootnode/main.go
anshalshukla: I am trying to import functions from this module, we don't use a different bootnode binary like in case of geth.

marcello33 (approved):

PR #1243, opened 2024-05-07
eth: explicitly commented the code were bor could get into snap-sync mode

pratikspatil024:
# Description

In the [recent upstream](https://github.com/maticnetwork/bor/releases/tag/v1.3.0-beta-2), [this PR](https://github.com/ethereum/go-ethereum/pull/28171) was included. As a result of that we observed that node went into snap sync mode. We do not support snap sync at the moment, so in this PR, I have explicitly commented the code were bor could get into snap-sync mode.

***Please note that that geth PR is important one when running a PBSS node, so in the future, we will have to find a workaround for the scenarios where the associated head state is missing in PBSS.***

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/sync.go
+++ b/eth/sync.go
@@ -200,22 +200,29 @@ func peerToSyncOp(mode downloader.SyncMode, p *eth.Peer) *chainSyncOp {
 }
 
 func (cs *chainSyncer) modeAndLocalHead() (downloader.SyncMode, *big.Int) {
-	// If we're in snap sync mode, return that directly
-	if cs.handler.snapSync.Load() {
-		block := cs.handler.chain.CurrentSnapBlock()
-		td := cs.handler.chain.GetTd(block.Hash(), block.Number.Uint64())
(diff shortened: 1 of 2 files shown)


cffls (approved):

eth/sync.go
```diff
@@ -216,15 +216,20 @@ func (cs *chainSyncer) modeAndLocalHead() (downloader.SyncMode, *big.Int) {
 			return downloader.SnapSync, td
```
manav2401, line 216: We still can go to snap sync in this case, right? I think this was the reason due to which we commented it earlier. Can we just return `downloader.FullSync, td` right away in the beginning?

manav2401 (approved):

PR #1258, opened 2024-06-05
eth: broadcast block to static and trusted peers as well

pratikspatil024:
# Description

The block is broadcasted to only Sqrt(peers) in geth. We observed that in case of validator-sentry architecture, the validator, sometime, might not receive a block which will force the validator to mine on his own, resulting in a reorg.

In this PR, we are making sure that the block will be broadcasted to all the static and trusted peers of a particular node.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -606,6 +606,17 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
 			log.Error("Propagating dangling block", "number", block.Number(), "hash", hash)
 			return
 		}
+
+		// These are the static and trusted peers which are not
+		// in `transfer := peers[:int(math.Sqrt(float64(len(peers))))]`
+		staticAndTrustedPeers := []*ethPeer{}
+
+		for _, peer := range peers[int(math.Sqrt(float64(len(peers)))):] {
+			if peer.IsTrusted() || peer.IsStatic() {
+				staticAndTrustedPeers = append(staticAndTrustedPeers, peer)
+			}
+		}
+
 		// Send the block to a subset of our peers
 		transfer := peers[:int(math.Sqrt(float64(len(peers))))]
 		for _, peer := range transfer {
@@ -613,6 +624,14 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
 		}
 
 		log.Trace("Propagated block", "hash", hash, "recipients", len(transfer), "duration", common.PrettyDuration(time.Since(block.ReceivedAt)))
+
+		// Send the block to the trusted and static peers
+		for _, peer := range staticAndTrustedPeers {
+			log.Trace("Propagating block to static and trusted peer", "hash", hash, "peerID", peer.ID())
+			peer.AsyncSendNewBlock(block, td)
+		}
+
+		log.Trace("Propagated same block to additional static and trusted peers", "hash", hash, "recipients", len(staticAndTrustedPeers), "duration", common.PrettyDuration(time.Since(block.ReceivedAt)))
 
 		return
 	}
--- a/eth/protocols/eth/peer.go
+++ b/eth/protocols/eth/peer.go
@@ -470,6 +470,16 @@ func (p *Peer) RequestTxs(hashes []common.Hash) error {
 	})
 }
 
+// IsTrusted returns whether the peer is a trusted peer or not.
+func (p *Peer) IsTrusted() bool {
+	return p.Info().Network.Trusted
+}
+
+// IsStatic returns whether the peer is a static peer or not.
+func (p *Peer) IsStatic() bool {
+	return p.Info().Network.Static
+}
+
 // knownCache is a cache for known hashes.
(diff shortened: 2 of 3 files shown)


eth/handler.go
```diff
@@ -606,13 +606,30 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
 			log.Error("Propagating dangling block", "number", block.Number(), "hash", hash)
 			return
 		}
+
+		// These are the static and trusted peers which are not
+		// in `transfer := peers[:int(math.Sqrt(float64(len(peers))))]`
+		staticAndTrustedPeers := []*ethPeer{}
+
+		for _, peer := range peers[int(math.Sqrt(float64(len(peers)))):] {
```
marcello33, line 614: Not sure about this, just asking: shouldn't this cycle go over the whole range of peers?

eth/handler.go
```diff
@@ -606,13 +606,30 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
...
 		}
 
-		log.Trace("Propagated block", "hash", hash, "recipients", len(transfer), "duration", common.PrettyDuration(time.Since(block.ReceivedAt)))
+		// Send the block to the trusted and static peers
+		for _, peer := range staticAndTrustedPeers {
+			log.Trace("Propagating block to static and trusted peer", "hash", hash, "peer", peer.ID())
+			peer.AsyncSendNewBlock(block, td)
+		}
+
+		log.Trace("Propagated block", "hash", hash, "recipients", len(transfer)+len(staticAndTrustedPeers), "duration", common.PrettyDuration(time.Since(block.ReceivedAt)))
```
marcello33, line 632: WDYT about logging the number of `transfer` recipient AND the number of `staticAndTrustedPeers` recipients?

eth/handler.go
```diff
@@ -606,13 +606,30 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
...
 		// Send the block to a subset of our peers
 		transfer := peers[:int(math.Sqrt(float64(len(peers))))]
 		for _, peer := range transfer {
 			peer.AsyncSendNewBlock(block, td)
 		}
 
-		log.Trace("Propagated block", "hash", hash, "recipients", len(transfer), "duration", common.PrettyDuration(time.Since(block.ReceivedAt)))
+		// Send the block to the trusted and static peers
+		for _, peer := range staticAndTrustedPeers {
+			log.Trace("Propagating block to static and trusted peer", "hash", hash, "peer", peer.ID())
```
marcello33, line 628: Shall we rename `peer` to `peerID`, for clarity?

eth/protocols/eth/peer.go
```diff
@@ -470,6 +470,16 @@ func (p *Peer) RequestTxs(hashes []common.Hash) error {
 	})
 }
 
+// IsTrusted returns whether the peer is a trusted peer or not.
+func (p *Peer) IsTrusted() bool {
+	return p.Info().Network.Trusted
```
marcello33, line 475: here and below, for both functions, no risks of `nil` pointers exceptions, right?

eth/handler.go
pratikspatil024: So this loop is to only get those static and trusted peers which are not eligible for the legacy broadcast. The peers we are skipping here are already included below
```
transfer := peers[:int(math.Sqrt(float64(len(peers))))]
```

I have mentioned this in the comment
```
// These are the static and trusted peers which are not
// in `transfer := peers[:int(math.Sqrt(float64(len(peers))))]`
staticAndTrustedPeers := []*ethPeer{}
```

eth/handler.go
pratikspatil024: Actually `peer` here is the `*ethPeer` structure. `peerID` would be `peer.ID()`

eth/handler.go
pratikspatil024: Makes sense. I refactored the log into 2 parts to keep the geth code unchanged. Does it makes sense?

eth/protocols/eth/peer.go
pratikspatil024: Yes, I think we should be fine here. Checked the `Info()` function and other occurances of it being called.

ssandeep (approved):

eth/handler.go
marcello33: Yeah, I meant in the trace log, where you print `peer` as `peer.ID()`
Anyway, minor one

marcello33 (approved):

cffls (approved):
LGTM

manav2401 (approved):

eth/handler.go
pratikspatil024: Oops, my bad. Updated, thanks!

PR #1382, opened 2024-12-06
Fix incorrect balance when CreateContract is used in block-stm

cffls:
# Description

`CreateContract` was a new function introduced from upstream merge after v1.5.0. When `CreateContract` is called, it should also call `mvRecordWritten` to ensure the state object is in the live set of statedb. Otherwise, some information, e.g. balance of the account, might get lost in the rest of opcodes in that txn.

# Changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [x] I have added unit tests
- [x] I have tested this code manually on mumbai/amoy

### Manual tests

# Additional comments

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1107,6 +1107,9 @@ func (s *StateDB) CreateAccount(addr common.Address) {
 // correctly handle EIP-6780 'delete-in-same-transaction' logic.
 func (s *StateDB) CreateContract(addr common.Address) {
 	obj := s.getStateObject(addr)
+	if obj != nil {
+		obj = s.mvRecordWritten(obj)
+	}
 	if !obj.newContract {
 		obj.newContract = true
 		s.journal.append(createContractChange{account: addr})
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -418,7 +418,6 @@ func (s *StateDB) ApplyMVWriteSet(writes []blockstm.WriteDescriptor) {
 
 			switch path.GetSubpath() {
 			case BalancePath:
-				// todo: @anshalshukla || @cffls - check balance change reason
 				s.SetBalance(addr, sr.GetBalance(addr), tracing.BalanceChangeUnspecified)
 			case NoncePath:
 				s.SetNonce(addr, sr.GetNonce(addr))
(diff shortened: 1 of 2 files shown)


manav2401:
The fix LGTM. I think the upstream merge introduced a few more changes and we should address them (not necessarily in this PR).

1. In the ApplyMVWriteSet function, this change was made during the upstream merge. 
```diff
case SuicidePath:
-		stateObject := sr.getDeletedStateObject(addr)
-		if stateObject != nil && stateObject.deleted {
+		stateObject := sr.getStateObject(addr)
+		if stateObject != nil {
			s.SelfDestruct(addr)
		}
```
This was done because `getDeletedStateObject` function was removed and `getStateObject` was used instead. As the code suggests above, we checked the `deleted` field of state object and then did self destruct. The new definition of `getStateObject` says that it will return nil if object was not found or deleted in this execution context. But, the newly added logic during upstream merge will proceed to self destruct if `stateObject != nil` which seems incorrect as it means that object wasn't deleted actually. The check should be `stateObject == nil` instead. 

Just wanted to confirm this because I wonder why any block until now has not failed. 

2. A minor nit: `CreateAccount` function calls `createObject` which internally runs the below code.
```
MVWrite(s, blockstm.NewAddressKey(addr))
```
The same thing is also executed in `CreateAccount` again which seems redundant. Nothing major, but it can be removed and the TODO can also be resolved. 

3. Rest of the TODOs are about adding reason for balance change. I think they can also be resolved.

manav2401 (approved):

cffls:
Thanks for the review!

1. This is a nice catch, but I think the current code is correct. Let's analyze `SelfDestruct` in two scenarios: a) pre-[EIP-6780](https://eips.ethereum.org/EIPS/eip-6780), and b) post-EIP-6780.
  a). In pre EIP-6780 era, a previously created account could be self-destructed by a new txn. If an address has a SelfDestruct  in the write list, we know the state object can exist and it should be self-destructed. The code seems to align with this.
  b). After EIP-6780, a previously created account could not be self-destructed in a new txn after, and we only need to consider when the object is created and self-destructed in the same txn. In this case, the object returned by `sr.getStateObject` will be nil, see [this code block](https://github.com/maticnetwork/bor/blob/71cbe3954c0b7110ea9322adf2676dc03409843d/core/state/statedb.go#L984-L987). It is important to understand that `ApplyMVWriteSet` applies the writes to the final statedb, named `s` in the code block, that hasn't been written by the current txn beforehand. Therefore, we won't need to do anything to the final statedb in this case.
2. Makes sense, will remove.
3. Agreed, will remove those todos.

pratikspatil024 (approved):

manav2401:
Thanks for the explanation. I wasn't aware of the full context.

PR #1392, opened 2024-12-23
Fallback to serial execution if parallel execution fails

cffls:
# Description

Enables the node to bypass unexpected errors from parallel execution and use the result from serial execution instead.

# Changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [x] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -39,7 +39,6 @@ import (
 	"github.com/ethereum/go-ethereum/common/mclock"
 	"github.com/ethereum/go-ethereum/common/prque"
 	"github.com/ethereum/go-ethereum/consensus"
-	"github.com/ethereum/go-ethereum/core/blockstm"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/state/snapshot"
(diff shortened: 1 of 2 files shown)


core/blockchain.go
```diff
@@ -606,44 +609,58 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header) (_
...
+				vtime = time.Since(vstart)
+			}
+			resultChan <- Result{receipts, logs, usedGas, err, statedb, blockExecutionSerialCounter, false}
 		}()
 	}
 
 	result := <-resultChan
 
-	if _, ok := result.err.(blockstm.ParallelExecFailedError); ok {
+	if result.parallel && result.err != nil {
```
Raneet10, line 657: We're almost always guaranteed to have the parallel execution result first in `resultChan` right ? By common sense it should be that way but just ensuring.

core/blockchain.go
```diff
@@ -606,44 +609,58 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header) (_
...
 		log.Warn("Parallel state processor failed", "err", result.err)
-
+		blockExecutionParallelErrorCounter.Inc(1)
 		// If the parallel processor failed, we will fallback to the serial processor if enabled
 		if processorCount == 2 {
-			result.statedb.StopPrefetcher()
 			result = <-resultChan
+			result.statedb.StopPrefetcher()
```
Raneet10, line 663: So if parallel execution doesn't fail we don't stop the prefetcher ? There's a clause below that seems to ensure we shut down both serial and parallel's prefetchers but my assumption could be wrong.

Raneet10 (approved):
LGTM! Just some minor doubts.

core/blockchain.go
cffls: Yes, we will mostly get parallel result. However, in the case where the result is returned from serial, this check `if result.parallel` will skip the code block.

core/blockchain.go
cffls: Yes, the one below will shutdown the remaining prefetcher, regardless what's left.

Raneet10: I don't think I understood perfectly. If the parallel execution succeeds, we only end up stopping the serial's prefetcher. Is that expected ?

cffls: Ahh let me try to clarify.. We will stop the prefetcher of the first finished executor at line 663 (the one you highlighted), and then stop the second prefetcher at [line 674](https://github.com/maticnetwork/bor/pull/1392/files#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215R674).

ssandeep (approved):

PR #1400, opened 2025-01-07
internal/cli: trim trailing / from ancient path

simonecid:
# Description

Fix a bug that leads to node corruption when `--datadir.ancient` has a trailing `/`.

To reproduce:

```
/usr/bin/bor snapshot prune-block --datadir=/var/lib/polygon/bor/ --datadir.ancient=/var/lib/polygon/bor/bor/chaindata/ancient/
```

Behaviour:
* `bor` saves `ancient_back` to `/var/lib/polygon/bor/bor/chaindata/ancient/ancient_back` because of `filepath.Split` at line 485
* folder `ancient` is deleted after pruning, removing `ancient_back`
* -> node is corrupted

Fix: use `TrimSuffix` to remove the trailing `/` is present.

# Changes

### DIFF
--- a/internal/cli/snapshot.go
+++ b/internal/cli/snapshot.go
@@ -473,7 +473,7 @@ func checkDeletePermissions(path string) (bool, error) {
 func (c *PruneBlockCommand) pruneBlock(stack *node.Node, fdHandles int) error {
 	name := "chaindata"
 
-	oldAncientPath := c.datadirAncient
+	oldAncientPath := strings.TrimSuffix(c.datadirAncient, "/")
 
 	switch {
 	case oldAncientPath == "":


avalkov (approved):

pratikspatil024 (approved):

go.mod
```diff
@@ -93,11 +93,11 @@ require (
 	go.opentelemetry.io/otel/sdk v1.27.0
 	go.uber.org/automaxprocs v1.5.3
 	go.uber.org/goleak v1.3.0
-	golang.org/x/crypto v0.24.0
+	golang.org/x/crypto v0.31.0
```
manav2401, line 96: I would refrain from updating any core libraries in this fix. Maybe that can be done in a separate PR. Can you revert them if possible?

manav2401: Okay my bad. It seems like they were introduced because of a change in base branch.

simonecid:
Thank you. I can not merge this PR myself. Who can do it?

PR #1412, opened 2025-01-23
Use PBSS and pebble by default

cffls:
# Description

Use PBSS and pebble by default

# Changes

- [x] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

This change will use path based storage scheme (PBSS) by default. Users running with this version would need to resync bor from genesis or use a snapshot created with PBSS.

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


packaging/templates/mainnet-v1/archive/config.toml
```diff
@@ -4,7 +4,8 @@ chain = "mainnet"
 # vmdebug = false
 datadir = "/var/lib/bor/data"
 # ancient = ""
-# "db.engine" = "leveldb"
+# "db.engine" = "pebble"
+state.scheme = "hash"
```
cffls, line 8: PBSS isn't supported in archive nodes, so we will continue using HBSS.

docs/cli/example_config.toml
```diff
@@ -8,7 +8,7 @@ verbosity = 3                   # Logging verbosity for the server (5=trace|4=de
 vmdebug = false                 # Record information useful for VM and contract debugging
 datadir = "var/lib/bor"         # Path of the data directory to store information
 ancient = ""                    # Data directory for ancient chain segments (default = inside chaindata)
-"db.engine" = "leveldb"           # Used to select leveldb or pebble as database (default = leveldb)
+"db.engine" = "pebble"           # Used to select leveldb or pebble as database (default = leveldb)
```
pratikspatil024, line 11: can we also change  the comment?

pratikspatil024: `state.scheme` is missing here. Maybe worth adding it here for consistency. Also, not here, but somewhere we can mention that `PBSS` it is not available for archive nodes?

pratikspatil024 (approved):

docs/cli/example_config.toml
manav2401: Or maybe fail starting archive node in PBSS with an error?

manav2401 (approved):

docs/cli/example_config.toml
cffls: Thanks for the suggestions! Added `state.scheme` here. Also add a check that will fail when PBSS and archive is enabled at the same time.

PR #1433, opened 2025-02-04
Upstream merge v1.14.10

lucca30:
# Description

Usptream merge from Geth.

**Releases Included:** [Aegis (v1.14.9)](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.9) and [Kopis (v1.14.10)](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.10)

### Points of attention:

- `core/blockchain.go` is a critical file in this upmerge. There was a lot of changes on it. So reviewer, please **double check on this**
- [PR 29179](https://github.com/ethereum/go-ethereum/pull/29179) 
  - This one I reverted since it removes `forkChoicer` which is an essential part of bor
- [PR 30125](https://github.com/ethereum/go-ethereum/pull/30125)
  - This one I reverted since it fixes a scenario for blob transactions (not supported by us) but also break some of our tests
- Some `eth/catalyst` tests were also skipped following the behaviour we already did on past upstream merges
- InterruptCtx got removed when doing conflict solving but then I re-inserted the feature back

# Changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests


manav2401 (commented):
Few minor nits. I am yet to review a few files.

core/blockchain.go
```diff
@@ -2308,29 +2321,49 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
...
 		}
-
 		// Retrieve the parent block and it's state to execute on top
 		start := time.Now()
 
 		parent := it.previous()
 		if parent == nil {
 			parent = bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
 		}
+		statedb, err := state.New(parent.Root, bc.statedb)
```
cffls, line 2337: As discussed, I think the code block from this line to 2358 could be deleted. `statedb` created here isn't used. Instead, it is created as part of `ProcessBlock`.

core/state_processor.go
```diff
@@ -76,51 +76,51 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
-	if withdrawals != nil {
-		withdrawals = nil
+	// Read requests if Prague is enabled.
+	var requests types.Requests
+	if p.config.IsPrague(block.Number()) {
+		requests, err = ParseDepositLogs(allLogs, p.config)
+		if err != nil {
+			return nil, err
+		}
 	}
```
cffls, line 113: Let's disable this. Unlike Ethereum, bor doesn't have deposit contract.

core/state_processor.go
```diff
@@ -76,51 +76,51 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
+		err     error
 	)
 	context = NewEVMBlockContext(header, p.hc, nil)
 	vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, p.config, cfg)
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
 		ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb)
 	}
+	if p.config.IsPrague(block.Number()) {
+		ProcessParentBlockHash(block.ParentHash(), vmenv, statedb)
+	}
```
cffls, line 88: We might want to add the same logic to `parallel_state_processor.go` as well.

core/vm/contracts.go
```diff
@@ -114,7 +118,7 @@ var PrecompiledContractsCancun = map[common.Address]PrecompiledContract{
 
 // PrecompiledContractsPrague contains the set of pre-compiled Ethereum
 // contracts used in the Prague release.
-var PrecompiledContractsPrague = map[common.Address]PrecompiledContract{
+var PrecompiledContractsPrague = PrecompiledContracts{
```
cffls, line 121: We want to add `common.BytesToAddress([]byte{0x01, 0x00}): &p256Verify{},` (bor-specific pre-compile) to Prague as well. If possible, add a unit test for it and make sure this precompile is always available in future HFs.

core/blockchain.go
```diff
@@ -627,19 +631,22 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header) (_
...
 			if err == nil {
 				vstart := time.Now()
-				err = bc.validator.ValidateState(block, parallelStatedb, receipts, usedGas, false)
+				err = bc.validator.ValidateState(block, parallelStatedb, res, false)
 				vtime = time.Since(vstart)
 			}
-			resultChan <- Result{receipts, logs, usedGas, err, parallelStatedb, blockExecutionParallelCounter, true}
+			if res == nil {
+				res = &ProcessResult{}
+			}
```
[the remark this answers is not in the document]
lucca30: Exactly. Looking at the return of this routine:
https://github.com/maticnetwork/bor/blob/0873515bcfb12950f9ac32d0198b1aadcd032da8/core/blockchain.go#L644

And the current develop code:
https://github.com/maticnetwork/bor/blob/01f9cd701fd856b0155dfe8783f53e5b44e93404/core/blockchain.go#L650-L660

We can see that setting `res` to an **empty object** only ensures we avoid a `nil reference` but does not change the behavior of the returned values. The execution flow remains the same, and all `receipts, logs, usedGas` will still return `nil`, just as before.

Let me know if you agree with this approach!

core/state_processor.go
```diff
@@ -76,51 +76,51 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
 	// Iterate over and process the individual transactions
 	for i, tx := range block.Transactions() {
-		if interruptCtx != nil {
-			select {
-			case <-interruptCtx.Done():
-				return nil, nil, 0, interruptCtx.Err()
-			default:
-			}
-		}
```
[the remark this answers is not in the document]
lucca30: Yeah, completely agreed. I'll include it back an validate against tests. Any issue found I'll let you know

core/types/block.go
```diff
@@ -103,6 +111,9 @@ type Header struct {
 
 	// ParentBeaconRoot was added by EIP-4788 and is ignored in legacy headers.
 	ParentBeaconRoot *common.Hash `json:"parentBeaconBlockRoot" rlp:"optional"`
+
+	// RequestsHash was added by EIP-7685 and is ignored in legacy headers.
+	RequestsHash *common.Hash `json:"requestsRoot" rlp:"optional"`
```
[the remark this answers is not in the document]
lucca30: That's a nice concern. I'll accept this help checking this.

core/blockchain.go
lucca30: 100% agreed. After removing these lines below:

https://github.com/maticnetwork/bor/blob/0873515bcfb12950f9ac32d0198b1aadcd032da8/core/blockchain.go#L2343-L2358

We basically solver the routine/memory leak while syncing the solution. But I'll now also remove these ones, as suggested:

https://github.com/maticnetwork/bor/blob/0873515bcfb12950f9ac32d0198b1aadcd032da8/core/blockchain.go#L2337-L2341

core/state_processor.go
lucca30: Ok, thanks. Disabling this.

core/state_processor.go
lucca30: Yes, it makes sense. Just double checking: 

Do you think I should apply this on all changes related to this [Pull Request](https://github.com/ethereum/go-ethereum/pull/29465/files)?

core/vm/contracts.go
lucca30: Hum, that's really interest. I included a test on it. My approach to protect us in new HFs was looking to an hypothetical "too far" block over `BorMainnetChainConfig` config. Would you suggest something else?

https://github.com/maticnetwork/bor/blob/9b0ca3650775548172dfc1af53f666188f3109b2/core/vm/contracts_test.go#L444-L460

core/state_processor.go
lucca30: After disabling I also needed to remove a test case. See commits below:

- https://github.com/maticnetwork/bor/pull/1433/commits/1193fc40e094ca0ba3e6f04e07eaf9ece040e94c
- https://github.com/maticnetwork/bor/pull/1433/commits/f19812f32cc1ad0f98cbbbeb43e87cde3a51913e

core/blockchain.go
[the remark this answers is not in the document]
manav2401: You're right actually. Makes sense, thanks for the explanation. As long as `err` is correct, it shouldn't be a problem.

core/state_processor.go
cffls: Yes, I think you should. Just applying the changes to `parallel_state_processor.go` is sufficient (and you can try to reuse the function if possible).

core/vm/contracts.go
cffls: I like this approach, thanks!

core/state_processor.go
lucca30: Nice. Thanks for the ideas. I applied here, but I just want to share some concerns while doing it:

1. `state_processor.go` separates the `HeaderChain` from the `BlockChain`. They do this because when running a stateless execution, the `Blockchain` is `nil` while `HeaderChain` is not.

https://github.com/maticnetwork/bor/blob/39968d1a98e4cd030432d05bcaf3343d55a5c45e/core/state_processor.go#L46-L50

https://github.com/maticnetwork/bor/blob/39968d1a98e4cd030432d05bcaf3343d55a5c45e/core/stateless.go#L67

2. So I'm assuming that we are **not going to support stateless execution for `parallel_state_processor.go`**. In this way I can safely call `p.bc.hc`, since we already call `p.bc.parallelSpeculativeProcesses` when executing in parallel.

For reference, here it is the applied change:
https://github.com/maticnetwork/bor/blob/f30ef31c25f82c75542cb13a4dfc42dcf71e2c2d/core/parallel_state_processor.go#L304-L309

cffls: Looks good. You are right, we don't need to worry about parallel execution in stateless mode for now.

core/types/block.go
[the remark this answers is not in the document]
lucca30: The checks were included. 
The commit is this one, for reference: https://github.com/maticnetwork/bor/pull/1433/commits/133b5bd7bb751cf84730527d4293a24daba4eafd

core/parallel_state_processor.go
```diff
@@ -386,13 +391,27 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
...
+	}
+
+	// Read requests if Prague is enabled.
+	var requests types.Requests
+	if p.config.IsPrague(block.Number()) {
+		requests, err = ParseDepositLogs(allLogs, p.config)
+		if err != nil {
+			return nil, err
+		}
 	}
```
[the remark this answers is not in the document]
lucca30: Yeah, I agree. And what about these ones? Should I also disable them?

### 1
https://github.com/maticnetwork/bor/blob/133b5bd7bb751cf84730527d4293a24daba4eafd/miner/worker.go#L1410-L1417
---
### 2
https://github.com/maticnetwork/bor/blob/133b5bd7bb751cf84730527d4293a24daba4eafd/core/chain_makers.go#L362-L371
---
### 3
https://github.com/maticnetwork/bor/blob/133b5bd7bb751cf84730527d4293a24daba4eafd/cmd/evm/internal/t8ntool/execution.go#L397-L418
---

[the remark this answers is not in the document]
lucca30: Disabled all 4 scenarios in the commit below:
- https://github.com/maticnetwork/bor/commit/ccede75223717d4e61e3512f609d0fe8127f4fa9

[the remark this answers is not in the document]
cffls: Another solution I had in mind, which * might * be more forward compatible, is to disable the code block if it is running in bor consensus, e.g.

```go
if config.IsPrague(b.header.Number) && config.bor == nil {
    ...
}
``` 

It keeps the code and prevents future conflicts if [more things](https://github.com/ethereum/go-ethereum/blob/master/miner/worker.go#L123-L133) are added to the code block from upstream.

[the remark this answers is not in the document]
lucca30: Hummm, fair enough. I do also agree. 
The most similar we can keep the code, the better for future upstreams. So I'll revert the last commit and disable via this idea.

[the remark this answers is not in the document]
lucca30: For reference, here it is the applied change:

- https://github.com/maticnetwork/bor/commit/2df763df6071261a1ed73ac1bcd86db1a50431ed

cffls (approved):

PR #1434, opened 2025-02-04
miner: reset ctx timeout before commit tx on new tx notif

manav2401:
# Description

A recent refactor in miner module caused an internal sentry node to continuously print the following logs:
```
Tx Level Interrupt                       hash=000000..000000
```

On further investigation, this channel in main loop always used to trigger a call to commit transactions. 
https://github.com/maticnetwork/bor/blob/d1219ff6d4debd0487f667f43a7a7486c5b1a091/miner/worker.go#L606

Because we use a shared context throughout the worker, it used to set an already timed out context which used to trigger this. 
https://github.com/maticnetwork/bor/blob/d1219ff6d4debd0487f667f43a7a7486c5b1a091/miner/worker.go#L940

Moreover, this case is to build blocks after the main `commitWork` is completed and there are some new transactions which can be filled. Hence, we individually process these transactions as we receive them. Currently, we don't honour commit interrupt which means that a heavy transaction can consume lot of time. 

Hence, before calling commit, we first check if we still have time to process using header's timestamp. If we do, we set the context with a timeout so that if it takes time to execute, we interrupt the process on time.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


cffls (commented):
Would it be possible to add a test for it? Otherwise we might run into the same issue again in the future after refactoring.

manav2401:
@cffls it's a bit difficult to test this via a unit test because every test is from a validator's perspective and one never receives the trigger to mine via the new transaction channel (because the `w.running` is set to true and the trigger is guarded by `!w.IsRunning()` check. Validator will always mine via normal `newHead` channel and produce blocks normally and then immediately commit block (without waiting for any more transactions even if there's time). 

The fix above only applies to sentry and it's not possible to test it (at leasts from the worker module where we can access the low level functions).

avalkov (approved):

cffls (approved):

PR #1444, opened 2025-02-12
consensus/bor: eliminate contract calls for current span

manav2401:
# Description

This PR replaces the logic to check and commit new spans.

Currently, in every sprint start, we make a contract call to fetch the current span. We check against that span if we need to commit a new one and if yes, we commit it to the contract. 

We can easily avoid the contract calls every sprint and make the check very simple and intuitive. Because the span length hasn't changed since genesis, we can simply use a formula to find out which span we're currently in and whether we need to commit a new one. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


lucca30:
Hi @manav2401.
This is a really good proposal in order to save performance. My only curiosity is if we can always safely says that the math you do here:

https://github.com/maticnetwork/bor/blob/64d0da37560991931786711ea3bfa7d45e9f4a3b/consensus/bor/bor.go#L1102-L1108

Is always equivalent on what happens on the [contract](https://github.com/maticnetwork/genesis-contracts/blob/96a19dd75502ee75d59469e5d40257aa5e33371f/contracts/BorValidatorSet.template#L119-L134):

```
  function getSpanByBlock(uint256 number) public view returns (uint256) {
    for (uint256 i = spanNumbers.length; i > 0; i--) {
      Span memory span = spans[spanNumbers[i - 1]];
      if (span.startBlock <= number && span.endBlock != 0 && number <= span.endBlock) {
        return span.number;
      }
    }

    // if cannot find matching span, return latest span
    if (spanNumbers.length > 0) {
      return spanNumbers[spanNumbers.length - 1];
    }

    // return default if not found any thing
    return 0;
  }
```

So is there any risks on lack of consistency because of consensus or any other scenario?

consensus/bor/bor.go
```diff
@@ -49,6 +49,11 @@ const (
 	inmemorySignatures = 4096 // Number of recent block signatures to keep in memory
 )
 
+const (
+	spanLength    = 6400 // Number of blocks in a span
```
lucca30, line 53: Regarding this constant, is it a common/acceptable scenario we change the span duration in devnet setups? If so, we should maybe look on how to query this data from the source.

Looking into heimdall in this [genesis file](https://github.com/maticnetwork/heimdall/blob/a59d2106a7e4cbadaf139d0ca67efc6e609ae700/builder/files/genesis-mainnet-v1.json#L131), it seems the source of truth comes from there.

manav2401: Actually, it's assumed to be constant in bor IMO. We haven't changed it since the beginning and even devnets defaults to 6400 blocks as 1 span. If we're changing it in future, we'll anyways have to address it in multiple places and we'll take care of this function at that time.

tests/bor/helper.go
```diff
@@ -364,7 +364,7 @@ func getMockedSpanner(t *testing.T, validators []*valset.Validator) *bor.MockSpa
 	spanner := bor.NewMockSpanner(gomock.NewController(t))
 	spanner.EXPECT().GetCurrentValidatorsByHash(gomock.Any(), gomock.Any(), gomock.Any()).Return(validators, nil).AnyTimes()
 	spanner.EXPECT().GetCurrentValidatorsByBlockNrOrHash(gomock.Any(), gomock.Any(), gomock.Any()).Return(validators, nil).AnyTimes()
-	spanner.EXPECT().GetCurrentSpan(gomock.Any(), gomock.Any()).Return(&span.Span{0, 0, 0}, nil).AnyTimes()
+	spanner.EXPECT().GetCurrentSpan(gomock.Any(), gomock.Any()).Return(&span.Span{0, 0, 0}, nil).MaxTimes(0)
```
manav2401, line 367: fyi, this change makes sure that `GetCurrentSpan` isn't called even once and the tests still works suggesting that the new implementation is replacable.

consensus/bor/bor.go
manav2401: This is a valid point and needs more though. My bad.

consensus/bor/bor.go
```diff
@@ -1094,39 +1099,44 @@ func (c *Bor) Close() error {
...
-) error {
-	var ctx = context.Background()
-	headerNumber := header.Number.Uint64()
+// SpanIdAt returns the corresponding span id for the given block number.
+func SpanIdAt(blockNum uint64) uint64 {
+	if blockNum > zerothSpanEnd {
+		return 1 + (blockNum-zerothSpanEnd-1)/spanLength
+	}
+	return 0
+}
```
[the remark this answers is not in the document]
lucca30: Okay. I assumed could happen sometimes to a span closes with a few more or less block. But if it always follows the pattern then it's okay,

consensus/bor/bor.go
manav2401: This is handled as we now fetch spans from heimdall (via the `/bor/params`) instead of using default 6400 value.

manav2401:
Closing this as the time taken by contract calls is ~1-2ms (sometimes even <1ms). The complexity and refactors this PR adds is not worth the effort. Should've measured the time earlier. Closing this for now.

Issue #1447, opened 2025-02-16
Observed Inconsistent Test Results in 3 CommitInterruptExperimentBor Cases


lucca30:
While running the `make test` on the project I realized there were 3 test cases showing some inconsistent results. Which means sometimes they work and sometimes they don't. Because of the difficulty to find it, I'm reporting here a little investigation I did **running each test 10 times**.

## Testcases

- `TestCommitInterruptExperimentBor_NewTxFlow` (~6s each test | failed in 8 out of 10)
- `TestCommitInterruptExperimentBorContract` (~35s each test | failed in 3 out of 10)
- `TestCommitInterruptExperimentBor_NormalFlow` (~13s each test | failed in 1 out of 10)

## Steps to Reproduce

- Checkout to develop branch:

```
git checkout develop
```

- Run the tests **multiple times**, I would suggest starting by `TestCommitInterruptExperimentBor_NewTxFlow` which is faster. You can use the commands below to run the tests and output the result in a file

```
for i in {1..10}; do go test -count=1 -timeout 0 -run ^TestCommitInterruptExperimentBor_NormalFlow$ github.com/ethereum/go-ethereum/miner > test_case_normal_flow_$i.txt 2>&1; done
for i in {1..10}; do go test -count=1 -timeout 0 -run ^TestCommitInterruptExperimentBor_NewTxFlow$ github.com/ethereum/go-ethereum/miner > test_case_new_tx_flow_$i.txt 2>&1; done
for i in {1..10}; do go test -count=1 -timeout 0 -run ^TestCommitInterruptExperimentBorContract$ github.com/ethereum/go-ethereum/miner > test_case_bor_contract_$i.txt 2>&1; done
```

PR #1450, opened 2025-02-17
miner: wait for next block before running commit interrupt tests

manav2401:
# Description

Fixes https://github.com/maticnetwork/bor/issues/1447

https://github.com/maticnetwork/bor/pull/1441 added new test for commit interrupt via new transactions channel. 

The results were non-deterministic and used to fail sometimes. This was because we use `w.current` to calculate the `delay` which needs to be correct for the test to work as expected. It takes a while before `w.current` is updated to the next block (block 3 in this case). This PR adds a small delay so that we have the new block and delays are accurately calculated for the test to work.

# Changes

# Checklist

- [x] Created a task in Jira and informed the team for implementation in Erigon client (if applicable)
- [x] Includes RPC methods changes, and the Notion documentation has been updated

### DIFF
--- a/miner/worker_test.go
+++ b/miner/worker_test.go
@@ -876,6 +876,9 @@ func TestCommitInterruptExperimentBor_NewTxFlow(t *testing.T) {
 			head := <-chainHeadCh
 			// We skip the initial 2 blocks as the mining timings are a bit skewed up
 			if head.Block.NumberU64() == 2 {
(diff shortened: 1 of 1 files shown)


marcello33 (approved):

lucca30:
10 out of 10 tests now running correct. Thanks!

#1447 now is closed

lucca30 (approved):

Issue #1455, opened 2025-02-19
Make heimdall timeout configurable


manav2401:
# Rationale

There's a hard coded timeout of 5s in bor when we fetch data from heimdall via api. John suggested to make it configurable as public endpoints might take more time (although it's not recommended to use public endpoints). 

https://github.com/maticnetwork/bor/blob/186deec830bd30db054fe0272687b39350c3f40b/consensus/bor/heimdall/client.go#L36-L37

# Implementation

Add a new flag and pass it across bor to configure it.

marcello33:
PR to be reviewed and possibly merged.

lucca30:
PR merged

PR #1457, opened 2025-02-19
Include Heimdall Timeout

lucca30:
# Description

Implementation for #1455 . Including configurable timeout for Heimdall. You can pass it via flag or via `config.toml`, like in the example below:

```
[heimdall]
  url = "http://localhost:1317"
  timeout = "30s" # <----------here
  "bor.without" = false
```

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


marcello33 (approved):

internal/cli/server/flags.go
```diff
@@ -167,6 +169,12 @@ func (c *Command) Flags(config *Config) *flagset.Flagset {
 		Value:   &c.cliConfig.Heimdall.URL,
 		Default: c.cliConfig.Heimdall.URL,
 	})
+	f.DurationFlag(&flagset.DurationFlag{
+		Name:    "bor.heimdalltimeout",
+		Usage:   "Timeout of Heimdall service",
```
manav2401, line 174: Let's make it more descriptive as it may mean something else as well. 

Maybe something like "Timeout for outgoing requests to heimdall" (just a suggestion, if there's anything better, feel free to use it).

eth/ethconfig/gen_config.go
```diff
@@ -26,6 +26,9 @@ func (c Config) MarshalTOML() (interface{}, error) {
 		NoPruning                            bool
 		NoPrefetch                           bool
 		TxLookupLimit                        uint64                 `toml:",omitempty"`
+		TransactionHistory                   uint64                 `toml:",omitempty"`
```
manav2401, line 29: Were these missing earlier?

internal/cli/server/flags.go
```diff
@@ -167,6 +169,12 @@ func (c *Command) Flags(config *Config) *flagset.Flagset {
...
+		Value:   &c.cliConfig.Heimdall.Timeout,
+		Default: 5 * time.Second,
```
manav2401, line 176: Actually what you did earlier was correct. 

You can add the default value here:
https://github.com/maticnetwork/bor/blob/3154fc4e0e1a1db56f95eac08a8929a3552940be/internal/cli/server/config.go#L647

Now you can use the variable directly like done in other flags.

manav2401 (commented):
Few minor nits.

lucca30:
> Might be worth calling `ticker.Reset()` here because if heimdall api timeout is >5s, the ticker will immediately get triggered when we enter loop again defeating it's purpose.
> 
> https://github.com/maticnetwork/bor/blob/3154fc4e0e1a1db56f95eac08a8929a3552940be/consensus/bor/heimdall/client.go#L325-L331
> 
> Or maybe another idea is to set ticker retry timeout to same as heimdall timeout?

@manav2401 , I like the idea of setting the ticker retry to the same as heimdall. Because it holds the behavior the user previously configured.

lucca30:
@manav2401 
For reference all comments were solved in this commit: https://github.com/maticnetwork/bor/pull/1457/commits/844603ea4eead83c38dc91c5a0322ba04ec9af2c

internal/cli/server/flags.go
manav2401: With the default being set in `config.go`, you won't need a hardcoded value anymore. You can directly use `c.cliConfig.Heimdall.Timeout` here in `Default` field.

manav2401 (approved):
One minor nit, except that, good to go!

lucca30:
> One minor nit. Except that, good to go!

Thanks for the clarification. Fixed on https://github.com/maticnetwork/bor/pull/1457/commits/5f6e2d334cbed39b0e8487094ad5e6d6ac298bbc

PR #1467, opened 2025-02-28
param: transactions announced only

jj1980a:
# Description

Introducing `txannouncementonly` option.
When enabled, the node will gossip transactions to its peers via announcements only.
Default value is `false` (standard behavior).

This is a convenience option for validators who enable the Fastlane MEV protocol.

Validators participating in the Fastlane protocol have been patching their sentry nodes to achieve the same for more than 2 years. Allowing it as a standard parameter would greatly improve validators experience on each new Bor release.

# Changes

# Checklist

## Testing

- [x] I have added unit tests
- [x] I have added tests to CI

### DIFF
--- a/docs/cli/example_config.toml
+++ b/docs/cli/example_config.toml
@@ -32,16 +32,17 @@ devfakeauthor = false           # Run miner without validator set authorization 
   enable-block-tracking = false   # Enables additional logging of information collected while tracking block lifecycle
 
 [p2p]
-  maxpeers = 50           # Maximum number of network peers (network disabled if set to 0)
-  maxpendpeers = 50       # Maximum number of pending connection attempts
-  bind = "0.0.0.0"        # Network binding address
-  port = 30303            # Network listening port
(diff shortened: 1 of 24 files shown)


cffls (approved):

pratikspatil024 (approved):

manav2401:
Hi @jj1980a sorry a bit late to the party but could you explain why would validators want to only announce transactions (and not broadcast it) when running with fastlane? Just curious to know the use case here.

jj1980a:
> Hi @jj1980a sorry a bit late to the party but could you explain why would validators want to only announce transactions (and not broadcast it) when running with fastlane? Just curious to know the use case here.

Hi @manav2401, the "announce only" setup is meant to be ran on validators sentry nodes (not on the validator itself), to streamline transactions propagation time from sentries to validator node. This is a requirement for the Fastlane auction system.

In-depth details about this are available in the Fastlane whitepaper: https://fastlane.xyz/PFL_WHITE_PAPER_1_5.pdf

PR #1472, opened 2025-03-03
consensus/bor: add span store abstraction for stateless verification

manav2401:
# Description

This PR adds a new abstraction called span store in bor consensus.

When bor is syncing, it verifies headers first before the block is processed. This means verifying headers in advance before for range of blocks in bor consensus's `verifyHeaders` function. During the header verification, currently bor needs access to `state` to do 2 things:

1. Maintain the snapshot structure (which has the info related to validator set). We call the child chain contract to populate the validator set every time we build a snapshot.
2. Verify the validator set from header's extradata field. 

There are 2 main issues with using state for header verification (which should be ideally stateless). 
1. If we're importing a side chain, we have to make sure we use the correct `state` to make the contract call else it can lead to bad blocks. 
2. If we're running snap sync in bor where the complete state is not available as it's downloaded via p2p. This can lead to failed verification leading to bor getting stuck as it can't really proceed. 

It's clear that we write the span info (which has validator set info inside) into the contract by fetching it from heimdall which means that we're considering heimdall as the source of truth. Instead of coupling the reading of validator set from state with header verification, we can actually decouple them and rely on heimdall instead (as we're anyways relying on heimdall for the writes). 

Span store is a minimal cache which maintains last 10 spans. It provide simple read methods via span id and block number so that we can use it in consensus. Ideally we would only fetch span from heimdall once and use it from cache for subsequent calls. This makes the verification pretty simple without adding any additional trust assumptions (e.g. availability of heimdall).

Note that post discussions, the logic is updated to have similar behaviour with [`getSpanByBlock`](https://github.com/maticnetwork/genesis-contracts/blob/master/contracts/BorValidatorSet.template#L118-L134) function in child chain contract. This is done to remove the assumption of a specific span length and 0th span end block. Also note that in some of the integration tests which run without heimdall client (which is a crucial requirement for span store), we fallback on spanner to fetch validator set from genesis state and construct a mock 0th span.

The header's extra data is populated via producer set and not validator set. Ideally, in heimdall, the power in both these sets should be same but is isn't for some blocks. We use producer set from span to avoid any mismatch issues. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


consensus/bor/span_store.go
```diff
@@ -0,0 +1,63 @@
...
+	cache, _ := lru.NewARC(10)
+	return SpanStore{
+		store:          cache,
+		heimdallClient: heimdallClient,
+	}
+}
+
+const (
+	defaultSpanLength = 6400 // Default span length i.e. number of bor blocks in a span
+	zerothSpanEnd     = 255  // End block of 0th span
```
cffls, line 28: Although the default span size of the first span is 256, it is possible to configure it through [contract genesis script](https://github.com/maticnetwork/genesis-contracts/blob/master/generate-borvalidatorset.js#L15-L19). Can this value be read from genesis contract itself?

consensus/bor/span_store.go
```diff
@@ -0,0 +1,63 @@
...
+func NewSpanStore(heimdallClient IHeimdallClient) SpanStore {
+	cache, _ := lru.NewARC(10)
+	return SpanStore{
+		store:          cache,
+		heimdallClient: heimdallClient,
+	}
+}
+
+const (
+	defaultSpanLength = 6400 // Default span length i.e. number of bor blocks in a span
```
cffls, line 27: Similar question for `defaultSpanLength`, could this be read from contract?

consensus/bor/span_store.go
manav2401: Good point, will fetch it from contract on the first run. Thanks!

consensus/bor/span_store.go
manav2401: Actually yes. We can fetch it from heimdall params. Will update the code to do so.

consensus/bor/span_store.go
manav2401: Updated the logic in https://github.com/maticnetwork/bor/pull/1472/commits/fc2462d0ee52c0bbd1bd5a686c5c3873f24584c9 as discussed today to use existing span's start and end block (as done in contracts) instead of relying on this formula with hardcoded values.

consensus/bor/span_store.go
manav2401: Updated the logic in https://github.com/maticnetwork/bor/pull/1472/commits/fc2462d0ee52c0bbd1bd5a686c5c3873f24584c9 as discussed today to use existing span's start and end block (as done in contracts) instead of relying on this formula with hardcoded values.

consensus/bor/span_store.go
```diff
@@ -0,0 +1,75 @@
...
+		s.store.Add(spanId, currentSpan)
+		if currentSpan.Span.ID > s.latestKnownSpanId {
+			s.latestKnownSpanId = currentSpan.ID
+		}
+	}
+
+	return currentSpan, nil
+}
+
+// spanByBlockNumber returns a span given a block number. It fetches span from heimdall if not found in cache. It
```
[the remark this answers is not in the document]
manav2401: @lucca30 agreed that if we assume span length to be constant we can fetch it easily without iterating at the risk of modifying the code when we change the span length in future. While I personally was in favour of such assumption, we discussed in the call that we should either have a generic logic or rely on contracts for that value to avoid any risks. On further looking, the child chain contracts don't really have any info on span length. They iteratively check for a block between a span's start and end block. I've tried to use this logic in the new `spanByBlockNumber` function. Also, I don't think it'll be a lot of overhead as in most of the cases, we only need the last span (and not any old one for which we'll have to iterate back) which is available in cache.

lucca30 (approved):

marcello33 (approved):

consensus/bor/span_store.go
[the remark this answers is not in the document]
cffls: Agreed with @manav2401 that it is safer to not assume the length of span is fixed.

cffls (approved):

manav2401:
Ignore the lint errors (due to `copyloopvar`) as they're addressed in upstream merge branch.

consensus/bor/span_store.go
```diff
@@ -0,0 +1,174 @@
...
+	if currentSpan == nil {
+		var err error
+		if s.heimdallClient == nil {
+			if spanId == 0 {
+				currentSpan, err = getMockSpan0(ctx, s.spanner, s.chainId)
+				if err != nil {
+					log.Warn("Unable to fetch span from heimdall", "id", spanId, "err", err)
+					return nil, err
+				}
+			} else {
```
lucca30, line 58: To verify the `Span 0` scenario, what do you think about mocking `IHeimdallClient` (since it’s an interface) and moving the `getMockSpan0()` helper into that mock? This would keep test-specific logic out of our production code. Thoughts?

manav2401: @lucca30 thanks for the suggestion. That's what I did earlier but there are some restrictions due to which I had to fallback to this approach. 

Basically, few e2e tests didn't set heimdall client at all. In those cases, bor consensus needs a mock span 0 and then constructs the next spans for test (see [this](https://github.com/maticnetwork/bor/blob/master/consensus/bor/bor.go#L1147)). It became very tricky for me to update the tests and consensus to always rely on heimdall client. Hence, for the sake of simplicity, I went ahead with passing the spanner to span store so that when heimdall client isn't available, we use spanner instead for those tests.

lucca30: Hum, I see. So LGTM

PR #1481, opened 2025-03-06
Add optional HeimdallWSAddress configuration and refactor handler_bor methods

lucca30:
# Description

This PR introduces the optional configuration parameter `HeimdallWSAddress`, which enables the application to establish a WebSocket connection with Heimdall for event listening rather than relying on polling. This enhancement aims to provide more efficient and real-time event handling.

Changes:

- **New Config**:
Added HeimdallWSAddress as an optional configuration setting to allow connecting via WebSocket to Heimdall.

- **WebSocket Integration**:
Updated the subscription logic to support event listening over a WebSocket connection. This change has been tested on a deployed devnet node, including a scenario where the connection is recovered after stopping bor.

- **Code Refactor**:
In handler_bor.go, refactored the handle and fetch methods to better segregate their roles, improving code clarity and maintainability.

# Changes

# Checklist

# Cross repository changes

- [x] This PR requires changes to heimdall
    - In case link the PR here: https://github.com/0xPolygon/heimdall-v2/pull/184
- [x] This PR requires changes to matic-cli
    - In case link the PR here: https://github.com/maticnetwork/matic-cli/pull/323

## Testing

- [ ] I have added unit tests


eth/backend.go
```diff
@@ -661,18 +660,25 @@ func (s *Ethereum) startCheckpointWhitelistService() {
...
-	const (
-		tickerDuration = 50 * time.Millisecond
-		fnName         = "whitelist milestone"
-	)
+	ethHandler, bor, _ := s.getHandler()
 
-	s.retryHeimdallHandler(s.handleMilestone, tickerDuration, whitelistTimeout, fnName)
+	// If heimdall ws is available use WS subscription to new milestone events instead of polling
+	if bor.HeimdallWSClient != nil {
+		s.subscribeAndHandleMilestone(context.Background(), ethHandler, bor)
```
avalkov, line 673: If we return from here, we should keep retrying to subscribe with some delay in my opinion.

consensus/bor/heimdallws/client.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, fmt.Errorf("failed to send subscription request: %w", err)
+	}
+
+	// Start the goroutine to read messages.
+	go c.readMessages(ctx)
+	return c.events, nil
+}
+
+// readMessages continuously reads messages from the websocket, handling reconnections if necessary.
+func (c *HeimdallWSClient) readMessages(ctx context.Context) {
```
avalkov, line 60: If we haven't received milestone in X seconds (like 30 sec maybe), we should log error for visibility by operator and reconnect I think.

consensus/bor/heimdallws/client.go
```diff
@@ -0,0 +1,192 @@
...
+			continue
+		}
+
+		// Replace the connection.
+		c.mu.Lock()
+		c.conn = newConn
+		c.mu.Unlock()
+		log.Info("Successfully reconnected on Heimdall WS subscription")
+
+		// Re-send the subscription request.
```
avalkov, line 179: Little bit of code repetition with the initial subscribe request above.

avalkov (approved):

eth/backend.go
lucca30: There's already a retry mechanism that avoid returning unless `ctx.Done()` is triggered or we close the events channel (which is protected by the subscribe retry mechanism). I tried to reproduce what we already have in `retryHeimdallHandler` and its `closeCh` param.

So I think we already safe from it, since I already tested dropping heimdall and connecting again.

consensus/bor/heimdallws/client.go
lucca30: Nice one. At first glance I put a timeout mechanism but for finishing the progress. I'm just not sure if we should Warn or log error. I'll include error, but if we agree on Warn I can after change it.

eth/backend.go
avalkov: `	milestoneEvents, err := bor.HeimdallWSClient.SubscribeMilestoneEvents(ctx)
	if err != nil {
		return err
	}`
	
Can return error.

lucca30: Good point, thanks

consensus/bor/heimdallws/client.go
lucca30: Done at https://github.com/maticnetwork/bor/pull/1481/commits/e1f7325d433adcbb2596acf6d2c99c2767643cdd

https://github.com/maticnetwork/bor/blob/e1f7325d433adcbb2596acf6d2c99c2767643cdd/consensus/bor/heimdallws/client.go#L104-L111

Now we always corrupt the connection after 30s which forces reconnection and a error log

eth/backend.go
lucca30: No more errors being raised by this function:
https://github.com/maticnetwork/bor/blob/e1f7325d433adcbb2596acf6d2c99c2767643cdd/eth/backend.go#L759-L765

With `tryUntilSubscribeMilestoneEvents` function we protect against errors.

pratikspatil024:
Can you also update the config files?
Also, I don't think we need to update the `cmd/` module, as we don't use it. Our command line is in `internal/server` which you have already updated

lucca30:
> Can you also update the config files? Also, I don't think we need to update the `cmd/` module, as we don't use it. Our command line is in `internal/server` which you have already updated

Since we currently not apply this for any other Heimdall configuration, even for heimdall url. I decided not to include this one and lefting as an optional flag.

https://github.com/maticnetwork/bor/blob/e1f7325d433adcbb2596acf6d2c99c2767643cdd/packaging/templates/mainnet-v1/archive/config.toml#L48-L52

pratikspatil024:
Makes sense. Thanks!

pratikspatil024 (approved):

PR #1492, opened 2025-03-21
eth: dial nodes from discv5

kamuikatsurgi:
# Description

Combination of https://github.com/ethereum/go-ethereum/pull/30302 and https://github.com/ethereum/go-ethereum/pull/31185 to fix the `p2p` issues we were facing.


cffls (approved):

manav2401 (approved):

manav2401:
Only the `TestCommitInterruptExperimentBor_NormalFlow` is failing which I'll address separately (as it's non deterministic and hence the failure).

PR #1493, opened 2025-03-21
Revert "eth: dial nodes from discv5"

manav2401:
Reverts maticnetwork/bor#1492

Accidentally squashed this instead of merging normally.


kamuikatsurgi:
Close this in favor of https://github.com/maticnetwork/bor/pull/1494?

PR #1495, opened 2025-03-21
cli,p2p: enable discv5 by default

kamuikatsurgi:
# Description

Now that https://github.com/maticnetwork/bor/pull/1494 is in, we should enable `discv5` by default. Geth has also done the same thing here: https://github.com/ethereum/go-ethereum/pull/30327


manav2401 (approved):
Can you run `make docs` to update the cli docs as well?

avalkov (approved):

PR #1499, opened 2025-03-26
Upstream merge v1.14.13

manav2401:
# Description

This PR contains upstream merge from geth including releases [Gei Hinnom (v1.14.12)](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.12) and [Schwarzschild (v1.14.13)](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.13). 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


core/state/statedb.go
```diff
@@ -773,42 +765,49 @@ func (s *StateDB) HasSelfDestructed(addr common.Address) bool {
...
 
-	if stateObject != nil {
-		stateObject = s.mvRecordWritten(stateObject)
-		stateObject.SubBalance(amount, reason)
-		MVWrite(s, blockstm.NewSubpathKey(addr, BalancePath))
+	// TODO(manav): Confirm if we need to record if amount is zero
+	stateObject = s.mvRecordWritten(stateObject)
+	MVWrite(s, blockstm.NewSubpathKey(addr, BalancePath))
+
+	if amount.IsZero() {
```
manav2401, line 800: @cffls a new check for zero amount is added. Do we need to capture using `MVWrite` in that case or not? I've kept that call anyways (irrespective of value) as I don't see any harm but just confirming.

core/state/statedb.go
```diff
@@ -831,13 +830,14 @@ func (s *StateDB) SetCode(addr common.Address, code []byte) {
...
 }
 
-func (s *StateDB) SetState(addr common.Address, key, value common.Hash) {
+func (s *StateDB) SetState(addr common.Address, key, value common.Hash) common.Hash {
 	stateObject := s.getOrNewStateObject(addr)
 	if stateObject != nil {
 		stateObject = s.mvRecordWritten(stateObject)
-		stateObject.SetState(key, value)
 		MVWrite(s, blockstm.NewStateKey(addr, key))
+		return stateObject.SetState(key, value)
```
manav2401, line 838: @cffls earlier the order was to set state and then do MVWrite. Because we want to return the new value, I've changed the order. I hope this doesn't affect anything.

core/blockchain_test.go
```diff
@@ -1512,6 +1513,7 @@ func checkLogEvents(t *testing.T, logsCh <-chan []*types.Log, rmLogsCh <-chan Re
 }
 
 func TestReorgSideEvent(t *testing.T) {
+	t.Skip("needs discussion")
```
manav2401, line 1516: Need some opinion on this. Geth has removed everything related to side chains from their code (as it's handled differently in ethereum now). This test fails for us after the refactor due to the following reasons. 

This test tries to insert a chain and then insert a replacement chain reorging the old one. Initially the replacement chain is side chain and later the original chain becomes side chain. See [this code](https://github.com/maticnetwork/bor/pull/1499/files#diff-d78b498ba524953c810b53b2f1e866245d052a063863ed820c9b4852f8762d69R1562-R1568). 

The `reorg` function in `blockchain.go` is also modified and it now no longer emits certain events after reorg is completed. One such case is the side chain event which was earlier emitted from the old chain during a reorg. That whole section is deleted (see [this](https://github.com/maticnetwork/bor/pull/1499/files#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215L3072)). 

Now, we can either bring that deleted code back and make this test work or update this test to not expected the events for an old chain. I am in favour of updating the tests keeping the reorg code aligned with upstream but I don't know if removal of those events can cause any other dependents of those events to fail or not.

core/state/statedb.go
cffls: Yea I think we can keep it as it is.

core/parallel_state_processor.go
```diff
@@ -395,12 +402,20 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 	}
 
 	// Read requests if Prague is enabled.
-	var requests types.Requests
-	if p.config.IsPrague(block.Number()) && p.config.Bor == nil {
-		requests, err = ParseDepositLogs(allLogs, p.config)
+	var requests [][]byte
+	if p.config.IsPrague(block.Number()) {
```
cffls, line 406: Should keep the check of bor here `p.config.Bor == nil`. Otherwise the following code will be executed after the HF.

core/parallel_state_processor.go
```diff
@@ -302,8 +302,15 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 
 	blockContext := NewEVMBlockContext(header, p.bc, nil)
 	context := NewEVMBlockContext(header, p.bc.hc, nil)
-	vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, p.config, cfg)
 
+	vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, p.config, cfg)
+	var tracingStateDB = vm.StateDB(statedb)
```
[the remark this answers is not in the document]
cffls: Sounds good

cffls (commented):
> @cffls with the state db interface refactor, logger has been removed from `state.StateDB` and hence the setter method of logger as well. I have removed it here but just wanted to confirm if it affects the block-stm logic in any way or not.  

This is okay. This logger is used by tracer and won't impact block-stm if removed.

core/blockchain_test.go
cffls: I can't open the link of deleted code because the PR webpage is too big. Could you link the code directly in repo?

core/state_processor.go
```diff
@@ -310,17 +330,60 @@
...
+		To:        &addr,
+	}
+	vmenv.Reset(NewEVMTxContext(msg), statedb)
+	statedb.AddAddressToAccessList(addr)
+	// Note(bor): It's okay to pass nil interrupt context as this is only for eth related things.
+	ret, _, _ := vmenv.Call(vm.AccountRef(msg.From), *msg.To, msg.Data, 30_000_000, common.U2560, nil)
+	statedb.Finalise(true)
+
+	// Create withdrawals requestsData with prefix 0x01
+	requestsData := make([]byte, len(ret)+1)
```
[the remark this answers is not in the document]
manav2401: Not applicable as we don't have withdrawals or requests.

core/parallel_state_processor.go
manav2401: Good catch, will add the check back. Will also need it in state processor.

internal/ethapi/api.go
```diff
@@ -1353,7 +989,8 @@ func doCall(ctx context.Context, b Backend, args TransactionArgs, state *state.S
 	// Make sure the context is cancelled when the call has completed
 	// this makes sure resources are cleaned up.
 	defer cancel()
-	gp := new(core.GasPool).AddGas(math.MaxUint64)
+	// Note(bor): don't set this to global max cap (as done in geth) as it leads to issues in devnet.
+	gp := new(core.GasPool).AddGas(gomath.MaxUint64)
```
manav2401, line 993: Need some opinions on this. There's a flag named `gascap` in json rpc config which denotes the max gas to use when executing `eth_call`. This is mostly for DoS Protection.

Seems like we're not correctly honouring it and deviate a bit from geth. 

Geth does something like this:
```go
gp := new(core.GasPool)
if globalGasCap == 0 {
       gp.AddGas(gomath.MaxUint64)
} else {
       gp.AddGas(globalGasCap)
}
```
`gp` is the gas pool used to track gas when running a tx / call. We add `globalGasCap` units of gas which is defaulted to `50M`. 

Upon doing this change, bor is unable to mine new blocks (at least on a devnet) with the errors below. 
```
Failed to prepare header for sealing     err="err: gas limit reached (supplied gas 9223372036854775807)"
```
While this call is made in the `eth_call` flow, I am not sure why it leads to errors in sealing new block. Irrespective of that, it clearly originates from the gas pool and suggests that we ran out of gas while running this call. This is expected if the `globalGasCap` is set to `50M` which can be low (for `eth_call`s). For now, I followed what @lucca30 did in last upstream which is setting the gas pool to max value possible of uint64. This does the work and makes the devnet work but seems like it's not honouring the gas cap value and opens up possibility of DoS attacks for rpc providers. 

We can either keep it as it is (max value of uint64) or honour it here and fix it for devnets by setting the value very high via flag/config.

core/blockchain_test.go
manav2401: Sure, apart from few changes in the `reorg` function, this section is completely deleted (if you want to check in ide, it's the last section of the function). There are few more changes (which might not be relevant) but for full picture you can compare the `reorg` function in `blockchain.go` of my branch to that of develop branch. 

```go
// Send out events for logs from the old canon chain, and 'reborn'
	// logs from the new canon chain. The number of logs can be very
	// high, so the events are sent in batches of size around 512.

	// Deleted logs + blocks:
	var deletedLogs []*types.Log

	for i := len(oldChain) - 1; i >= 0; i-- {
		// Also send event for blocks removed from the canon chain.
		bc.chainSideFeed.Send(ChainSideEvent{Block: oldChain[i]})

		// Collect deleted logs for notification
		if logs := bc.collectLogs(oldChain[i], true); len(logs) > 0 {
			deletedLogs = append(deletedLogs, logs...)
		}

		if len(deletedLogs) > 512 {
			bc.rmLogsFeed.Send(RemovedLogsEvent{deletedLogs})
			deletedLogs = nil
		}
	}

	if len(deletedLogs) > 0 {
		bc.rmLogsFeed.Send(RemovedLogsEvent{deletedLogs})
	}

	// New logs:
	var rebirthLogs []*types.Log

	for i := len(newChain) - 1; i >= 1; i-- {
		if logs := bc.collectLogs(newChain[i], false); len(logs) > 0 {
			rebirthLogs = append(rebirthLogs, logs...)
		}

		if len(rebirthLogs) > 512 {
			bc.logsFeed.Send(rebirthLogs)
			rebirthLogs = nil
		}
	}

	if len(rebirthLogs) > 0 {
		bc.logsFeed.Send(rebirthLogs)
	}

	return nil
```

internal/ethapi/api.go
lucca30: Nicely pointed @manav2401 . The first time I looked for this problem I was in the same situation: there was some previous changes that were forcing us to always need gp as `MaxUint64`, because instead it will overflow.

I think for this upstream merge it's enough keeping this way, but its definitely worth creating an issue to look specifically on that, since DoS prevention is a really important feature.

core/blockchain_test.go
cffls: Thanks for providing the details. Although it might be inconvenient for future upstream merge, I prefer keeping the reorg events since we don't know the downstream impact of it. Some dapps might heavily depends on these events.

manav2401: Geth pushed this refactor pretty normally in a PR without any separate announcement but yeah there's no harm in keeping this code. Will add it back (hopefully there are no errors).

marcello33 (approved):

cffls (approved):

core/chain_makers.go
```diff
@@ -461,16 +477,15 @@ func GenerateVerkleChain(config *params.ChainConfig, parent *types.Block, engine
...
-		// if !config.IsPrague(b.parent.Number(), b.parent.Time()) {
-		// Transition case: insert all 256 ancestors
-		// 		InsertBlockHashHistoryAtEip2935Fork(statedb, b.header.Number.Uint64()-1, b.header.ParentHash, chainreader)
-		// 	} else {
-		// 		ProcessParentBlockHash(statedb, b.header.Number.Uint64()-1, b.header.ParentHash)
-		// 	}
-		// }
-		// Execute any user modifications to the block
+		// Pre-execution system calls.
+		if config.IsPrague(b.header.Number) {
```
manav2401, line 481: TODO: add `config.Bor == nil ` check here.

core/genesis.go
```diff
@@ -513,11 +512,12 @@ func (g *Genesis) toBlockWithRoot(root common.Hash) *types.Block {
 			}
 		}
 		if conf.IsPrague(num) {
```
manav2401, line 514: TODO: add `config.Bor == nil` check here.

core/state_processor.go
```diff
@@ -77,14 +77,22 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
 	vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, p.config, cfg)
+	var tracingStateDB = vm.StateDB(statedb)
+	if hooks := cfg.Tracer; hooks != nil {
+		tracingStateDB = state.NewHookedState(statedb, hooks)
+	}
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
-		ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb)
+		ProcessBeaconBlockRoot(*beaconRoot, vmenv, tracingStateDB)
 	}
 	if p.config.IsPrague(block.Number()) {
```
manav2401, line 92: TODO: add `p.config.Bor == nil` check here

core/blockchain.go
```diff
@@ -2334,6 +2326,30 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness
...
+			if bc.vmConfig.StatelessSelfValidation || (makeWitness && len(chain) == 1) {
+				witness, err = stateless.NewWitness(block.Header(), bc)
+				if err != nil {
+					return nil, it.index, err
+				}
+			}
+			// Bor: We start the prefetcher in process block function called below
+			// and not here as we copy state for block-stm in that function. Also,
+			// we don't want to start duplicate prefetchers per block.
+			// statedb.StartPrefetcher("chain", witness)
```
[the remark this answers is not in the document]
manav2401: Here's the commit for full changes
https://github.com/maticnetwork/bor/pull/1499/commits/8ea52f50367c64ee99ccf340b4a407368a2dcc14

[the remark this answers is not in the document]
cffls: The change looks good to me.

internal/ethapi/api.go
manav2401: Created a task: https://polygon.atlassian.net/browse/POS-2964
Resolving this for now.

core/blockchain.go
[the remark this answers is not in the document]
manav2401: @lucca30 thanks for sharing reference. 

I assume witness generation is disabled in bor by default, right? In that case, it would use a `nil` witness while starting pre-fetcher for parallel executor and things should work fine I believe. As we enable witness generation in the `stateless_sync` branch, the reference you shared looks good to me. Just pattern wise, maybe it's better to pass `witness` into the `processBlock` function and use `nil` when doing it for block-stm explicitly. This will allow use to maintain compatibility with upstream (as it will try to start prefetcher outside while we do it inside the function). wdyt?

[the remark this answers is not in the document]
lucca30: Completely agreed. 

Specially because this is an already mapped issue, so there is no risk on missing it after merging and enabling the stateless feature.

PR #1507, opened 2025-03-28
Block Witness on Merkle Tree

lucca30:
# Description

The main goal with this DRAFT, is analyze the size of the witness block against mainnet blocks and also validate if it is properly working.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -365,7 +365,7 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 	bc.statedb = state.NewDatabase(bc.triedb, nil)
 	bc.validator = NewBlockValidator(chainConfig, bc)
 	bc.prefetcher = newStatePrefetcher(chainConfig, bc.hc)
-	bc.processor = NewStateProcessor(chainConfig, bc, bc.hc)
+	bc.processor = NewStateProcessor(chainConfig, bc.hc)
 
 	bc.genesisBlock = bc.GetBlockByNumber(0)
 	if bc.genesisBlock == nil {
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -293,7 +293,6 @@ type BlockChain struct {
 
 	// Bor related changes
 	borReceiptsCache *lru.Cache[common.Hash, *types.Receipt] // Cache for the most recent bor receipt receipts per block
-	stateSyncData    []*types.StateSyncData                  // State sync data
 	stateSyncFeed    event.Feed                              // State sync feed
 	chain2HeadFeed   event.Feed                              // Reorg/NewHead/Fork data feed
 }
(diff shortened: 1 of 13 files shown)


core/state_processor.go
```diff
@@ -77,7 +75,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		context vm.BlockContext
 		signer  = types.MakeSigner(p.config, header.Number, header.Time)
 	)
-	context = NewEVMBlockContext(header, p.hc, nil)
+	context = NewEVMBlockContext(header, p.chain, author)
```
cffls, line 78: Curious why we didn't need to pass `author` to the context before. iirc, the author could be derived by the signature of the block. Is it possible that this failed in stateless execution because the block signature wasn't copied correctly?

core/state_processor.go
```diff
@@ -38,16 +38,14 @@ import (
 // StateProcessor implements Processor.
 type StateProcessor struct {
 	config *params.ChainConfig // Chain configuration options
-	bc     *BlockChain         // Canonical header chain
-	hc     *HeaderChain
+	chain  *HeaderChain        // Canonical header chain
```
cffls, line 41: Seems like a code refactor/cleanup, is it absolutely required?

core/blockchain.go
```diff
@@ -641,24 +659,41 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header) (_
...
+			witness, err := stateless.NewWitness(block.Header(), bc)
+			if err != nil {
+				log.Error("error in witness generation", "caughterr", err)
+			}
+
+			statedb.StartPrefetcher("chain", witness)
 			pstart := time.Now()
-			res, err := bc.processor.Process(block, statedb, bc.vmConfig, ctx)
+
+			res, err := bc.processor.Process(block, statedb, bc.vmConfig, nil, ctx)
```
cffls, line 684: Would be interesting to see how much impact witness generation will have on block execution time.

core/state_processor.go
lucca30: Yeah, when we run stateless execution we basically delete some critical fiels in order to force recalculation. See the snippet below:

https://github.com/maticnetwork/bor/blob/f507b4b80df5319d2283a537d442c5ba26705ecf/core/blockchain.go#L732-L737

With this, when we try to get the Context from the signed block via `ecrecover` but missing these critical fields it basically generates random values. That's why when we were debugging we found an address which were not a block producer.

core/state_processor.go
lucca30: Yeah, this one I got from a recent version of geth. `bc` is actually never used, that's why we can remove.

core/blockchain.go
lucca30: Definetely another interesting value to measure into our plots. Thanks for raising this idea

PR #1512, opened 2025-04-09
Bor without heimdall

avalkov:
# Description

Supporting bor to run without heimdall


consensus/bor/bor.go
```diff
@@ -1196,24 +1339,127 @@ func (c *Bor) FetchAndCommitSpan(
...
+				if response, err = c.getLatestHeimdallSpanV2(); err != nil {
+					log.Error("Error while fetching last heimdallv2 span from db", "error", err)
+					return false, err
+				}
+				if response != nil {
+					response.Id = newSpanID
+					spanLength := getSpanLength(response.StartBlock, response.EndBlock)
+					response.StartBlock = getSpanStartBlock()
+					response.EndBlock = getSpanEndBlock(response.StartBlock, spanLength)
+					return true, nil
```
cffls, line 1390: It seems like the startblock and endblock are overridden here for every span. What will happen when heimdall just proposes a new span, will its values get overridden?

consensus/bor/bor.go
```diff
@@ -1116,6 +1119,12 @@ func (c *Bor) checkAndCommitSpan(
 	header *types.Header,
 	chain core.ChainContext,
 ) error {
+	if hmm.IsHeimdallV2 {
+		c.saveLatestHeimdallSpanV2()
+	} else {
+		c.saveLatestHeimdallSpanV1()
+	}
+
```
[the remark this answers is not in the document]
avalkov: Heimdall usually has 1 more span than the span that bor currently executes. We need to make sure we have the latest span from heimdall and we gonna use this same span as many times as bor wants to self commit span without heimdall.

consensus/bor/bor.go
avalkov: If heimdall is up this logic will not trigger, this logic triggers just when heimdall is down and we have saved latest heimdall span with the `saveLatestHeimdallSpanVX` call above. We need to update that span id, start block, end block and use same validators and producers set.

consensus/bor/bor.go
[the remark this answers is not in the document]
cffls: Thanks for the explanation. This makes sense. I think renaming the functions to `updateLatestHeimdallSpanVX` might reduce some confusions.

consensus/bor/bor.go
cffls: One thing I failed to understand is why don't we move this update logic to `saveLatestHeimdallSpanVX`, and here we can simply get latest span? In that way, there will be only one place where we know the span could be modified/updated.

consensus/bor/bor.go
```diff
@@ -1131,6 +1142,140 @@ func (c *Bor) checkAndCommitSpan(
...
+	var storedSpan borTypes.Span
+	if err := json.Unmarshal(storedSpanBytes, &storedSpan); err != nil {
+		log.Error("Error while unmarshalling heimdallv2 span", "error", err)
+		return nil, nil
+	}
+
+	return &storedSpan, nil
+}
```
lucca30, line 1185: Instead of returning nil on the error, shouldn't we pass it above? If no scenario should pass error, maybe it's worth removing it from function signature.

consensus/bor/bor.go
```diff
@@ -1131,6 +1142,140 @@ func (c *Bor) checkAndCommitSpan(
...
+	respSpanBytes, err := json.Marshal(respSpan)
+	if err != nil {
+		log.Error("Error while marshalling heimdallv1 span", "error", err)
+		return
+	}
+
+	if err := c.db.Put(rawdb.LastHeimdallV1SpanKey, respSpanBytes); err != nil {
+		log.Error("Error while saving heimdallv1 span to db", "error", err)
+		return
+	}
```
lucca30, line 1229: Should we protect this write operation with a rw mutex? 

Same for `rawdb.LastHeimdallV2SpanKey` on `saveLatestHeimdallSpanV2()`

maoueh:
Does this PR will enable running integration tests against `bor` alone without having to run both `heimdalld` and `bor`? Looking into ways to easily run tracing integrations tests against a `bor`, I'm especially interest in testing some of `bor` specific system transactions.

consensus/bor/bor.go
```diff
@@ -1196,24 +1341,127 @@ func (c *Bor) FetchAndCommitSpan(
...
 		}
-		chainId = response.BorChainId
+		retryWithBackoff := func(exec func() (bool, error)) error {
+			i := 1
+			for {
+				done, err := exec()
+				if err != nil {
+					return err
+				}
+
```
lucca30, line 1360: Is it possible it infinitely returns `(false, nil)` on `exec()`? Maybe it's worth including a `maxRetries`

consensus/bor/bor.go
```diff
@@ -1131,6 +1142,140 @@ func (c *Bor) checkAndCommitSpan(
...
+		return
+	}
+
+	storedSpanID := uint64(0)
+
+	if storedSpan != nil {
+		storedSpanID = storedSpan.ID
+	}
+
+	if respSpan.ID <= storedSpanID {
```
lucca30, line 1215: Is it possible both V1 and V2 use same patter for `id`?

In this function: `respSpan.ID`
In saveLatestHeimdallSpanV2(): `respSpan.Id`

consensus/bor/bor.go
```diff
@@ -1131,6 +1142,140 @@ func (c *Bor) checkAndCommitSpan(
...
+		log.Info("Latest heimdallv1 span is not updated", "storedSpanID", storedSpanID, "respSpanID", respSpan.ID)
+		return
+	}
+
+	respSpanBytes, err := json.Marshal(respSpan)
+	if err != nil {
+		log.Error("Error while marshalling heimdallv1 span", "error", err)
+		return
+	}
```
lucca30, line 1224: We already have respSpan in it "protobuf" at this part of the code. Isn't better store in this format instead of json?

Just recommend doing it if there is no extra effort than switching this now. It's a low gain of performance but worth if it's easy to apply now.

consensus/bor/heimdallgrpc/state_sync_v1.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/ethereum/go-ethereum/consensus/bor/clerk"
+
+	proto "github.com/maticnetwork/polyproto/heimdall"
+)
+
+func (h *HeimdallGRPCClient) StateSyncEventsV1(ctx context.Context, fromID uint64, to int64) ([]*clerk.EventRecordWithTime, error) {
+	eventRecords := make([]*clerk.EventRecordWithTime, 0)
+
+	req := &proto.StateSyncEventsRequest{
+		FromID: fromID,
```
lucca30, line 18: Is it worth protecting this and V2 from large return objects (because of big diff between `fromID` and `to`)?

consensus/bor/bor.go
avalkov: This logic here will trigger only at last sprint of the span, we want to fetch the latest heimdall span as soon as possible because we dont know when heimdall could go down.

avalkov:
> Does this PR will enable running integration tests against `bor` alone without having to run both `heimdalld` and `bor`? Looking into ways to easily run tracing integrations tests against a `bor`, I'm especially interest in testing some of `bor` specific system transactions.

These changes expect that heimdall was running atleast for a while and few spans were executed. But to your point, if you just create mock heimdall API service, you can do run bor and do tests on it.
Also there is `DeveloperConfig` that you can use to mine blocks without heimdall - https://github.com/maticnetwork/bor/blob/master/internal/cli/server/config.go#L591

consensus/bor/bor.go
```diff
@@ -1116,6 +1119,14 @@ func (c *Bor) checkAndCommitSpan(
 	header *types.Header,
 	chain core.ChainContext,
 ) error {
+	if header.Number.Uint64()%c.config.CalculateSprint(header.Number.Uint64()) == 0 {
```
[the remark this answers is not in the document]
avalkov: `CalculateSprint` returns the sprint length for given block number, because at earlier blocks the sprint length was different we need such function.
It should never return 0.

consensus/bor/bor.go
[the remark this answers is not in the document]
avalkov: Done.

consensus/bor/bor.go
avalkov: Good catch, removed and refactored the code where this function is used.

consensus/bor/bor.go
avalkov: Since read/write operations happen from block building and verification loop, there should be no way for conflict?

consensus/bor/bor.go
avalkov: The idea I have is, if we dont have stored span to return to just keep calling heimdall to get the span, I am not sure its better to return error from block building/verification loop?

consensus/bor/bor.go
avalkov: Good catch, refactored.

consensus/bor/heimdallgrpc/state_sync_v1.go
avalkov: The request has limit.

`	req := &proto.StateSyncEventsRequest{
		FromID: fromID,
		ToTime: uint64(to),
		Limit:  uint64(stateFetchLimit),
	}
`

consensus/bor/bor.go
lucca30: That's a good point. Not sure of when we should keep the application retrying until it recover. Maybe here is the better place.

consensus/bor/bor.go
lucca30: I think this one got missed on the fix commit:

https://github.com/maticnetwork/bor/blob/7fef2f61b099f21310520d0dccd9124216c58ba9/consensus/bor/bor.go#L1215

https://github.com/maticnetwork/bor/blob/7fef2f61b099f21310520d0dccd9124216c58ba9/consensus/bor/bor.go#L1260

consensus/bor/bor.go
avalkov: The current logic will just keep retrying on `GetSpan` so I try to keep the same behavior.

consensus/bor/bor.go
lucca30: Nice. It would only beneficial if it's an easy one. If not, keep it that way.

consensus/bor/bor.go
lucca30: Yeah, agree

consensus/bor/bor.go
lucca30: Yeah, it's seems there is no way for conflict.

maoueh:
> These changes expect that heimdall was running atleast for a while and few spans were executed

I see thanks for the extra details, I can see the use case.

> if you just create mock heimdall API service, you can do run bor and do tests on it.

Yeah I tried copying some existing "integration" tests and expand them, but mocking is a bit hard. How do you guys locally test bor + heimdalld combination? Or maybe you rely solely on unit/integration tests? 

`heimdalld start --chain=local` is quite easy, how to connect `bor` to it to form a valid 1 local network?

PR #1513, opened 2025-04-10
Enable Ancient Block Pruning with Path Based State Scheme

keefertaylor:
# Description

Path based state scheme is now the default for `bor`. [Ancient block pruning](https://forum.polygon.technology/t/pip-32-ancient-data-pruning/13346) is a feature that saves non-trivial amount of disk space for node operators.

After the merge of [geth v1.14.8](https://github.com/maticnetwork/bor/pull/1410), the ancient freezers support tail truncation for path based state schemes. This means we can enable ancient block pruning. 

MPT verification still fails with path based state scheme, and I'm not familiar enough with `geth`/`bor` code to get it working. Given that this is an optional flag I've simply added an error. 

The more or less [official path based state scheme snapshot](https://publicnode.com/snapshots#polygon) is 3.4TB including chaindata and ancients. After running ancient block pruning with the default `-block-amount-reserved=1024` flag, our node weighs in at 1.2TB (2.2TB saved / ~65% reduction in space usage). 

# Changes

# Nodes audience

N/A all node operators can uptake.

# Checklist

## Testing

I don't see an easy way to add appropriate unit tests, and most (if not all) of the freezer logic seems covered by unit tests that the `geth` team added and were [pulled in v1.14.8](https://github.com/maticnetwork/bor/pull/1410).

### Manual tests

[Tessellated](https://validator.info/polygon/59) has been running this on multiple `bor` nodes for the last two weeks on polygon mainnet and regularly re-pruning without issue.

Here's the command we use to prune: 
```
bor snapshot prune-block -datadir /home/bor -datadir.ancient /home/bor/bor/chaindata/ancient/chain
```

### DIFF
--- a/internal/cli/snapshot.go
+++ b/internal/cli/snapshot.go
@@ -24,8 +24,6 @@ import (
 	"github.com/mitchellh/cli"
 )
 
-var errPbssNotSupported = errors.New("ancient block pruning is not supporeted on path based storage scheme")
-
 // SnapshotCommand is the command to group the snapshot commands
 type SnapshotCommand struct {
 	UI cli.Ui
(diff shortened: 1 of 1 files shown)


manav2401:
Hi @keefertaylor, thanks for the PR but the reason why we haven't enabled ancient pruning in PBSS is because it requires some extra work than checking for PBSS. On top of my mind, there are 2 major things which needs to be done. 
1. Implement pruning for PBSS related tables in the ancient store (which is not done as of now)
2. Update the logic to check the recent most snapshot and MPT data for PBSS

I am not sure about the timeline but we will look into this soon. If you're up for implementing it, I would be happy to review it. For now, closing this. Thanks!

PR #1518, opened 2025-04-15
Add witnesses to downloader

cffls:
# Description

Enable downloader to download witnesses from peers. 
Note that enabling witnesses in download isn't sufficient for the stateless node to keep up with the network. We also need to add witness fetching from `block_fetcher.go`.


eth/handler.go
```diff
@@ -385,7 +385,7 @@ func (h *handler) runEthPeer(peer *eth.Peer, handler eth.Handler) error {
 		return errors.New("peer dropped during handling")
 	}
 	// Register the peer in the downloader. If the downloader considers it banned, we disconnect
-	if err := h.downloader.RegisterPeer(peer.ID(), peer.Version(), peer); err != nil {
```
pratikspatil024, line 388: this just makes it more convenient, right? Doesn't seem to change anything

pratikspatil024 (approved):

eth/handler.go
cffls: I had to make this change otherwise the compiler will complain that `peer` didn't implement `RequestWitnesses`.

pratikspatil024: Aha, okay.

PR #1519, opened 2025-04-16
Stateless execution

lucca30:
# Description
Stateless execution is now functional. To support it, we clear parts of the block header, which prevents retrieving the original coinbase (author) since that field is part of the signed data. As a result, the author must now be explicitly provided to avoid incorrect values.

# Changes


cffls (approved):
LGTM, thanks!

PR #1524, opened 2025-04-22
Fix Span Committing on Stateless Client

lucca30:
# Description

Related Fixes:

- Include witness StateDB on `GetCurrentSpan` call on stateless client
- Include state prefetching of `LastStateId` call on block producer 

# Changes

# Tests

- Tests made on Kurtosis, solution is open and running. But facing issues on syncing

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -1137,13 +1137,16 @@ func (c *Bor) checkAndCommitSpan(
 	var ctx = context.Background()
 	headerNumber := header.Number.Uint64()
 
-	span, err := c.spanner.GetCurrentSpan(ctx, header.ParentHash)
+	span, err := c.spanner.GetCurrentSpan(ctx, header.ParentHash, state)
 	if err != nil {
 		return err
 	}
 
 	if c.needToCommitSpan(span, headerNumber) {
-		return c.FetchAndCommitSpan(ctx, span.ID+1, state, header, chain)
+		err = c.FetchAndCommitSpan(ctx, span.ID+1, state, header, chain)
+		if err != nil {
+		}
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


consensus/bor/bor.go
```diff
@@ -1137,13 +1137,16 @@ func (c *Bor) checkAndCommitSpan(
...
 	if err != nil {
 		return err
 	}
 
 	if c.needToCommitSpan(span, headerNumber) {
-		return c.FetchAndCommitSpan(ctx, span.ID+1, state, header, chain)
+		err = c.FetchAndCommitSpan(ctx, span.ID+1, state, header, chain)
+		if err != nil {
+		}
+		return err
```
cffls, line 1149: Maybe a typo?

lucca30: Yes, it was. I included this err var to log and missed to replaced it back to original format.

lucca30: Fixed now

cffls (approved):
LGTM!

consensus/bor/bor.go
```diff
@@ -1265,7 +1266,15 @@ func (c *Bor) CommitStates(
...
-		lastStateIDBig, err = c.GenesisContractsClient.LastStateId(state.Copy(), number-1, header.ParentHash)
+		tempState := state.Copy()
+		lastStateIDBig, err = c.GenesisContractsClient.LastStateId(tempState, number-1, header.ParentHash)
+
+		if tempState.Witness() != nil {
+			// Clone the witness Codes and State sets from the temporary snapshot back into the original state
+			state.Witness().Codes = maps.Clone(tempState.Witness().Codes)
+			state.Witness().State = maps.Clone(tempState.Witness().State)
+		}
+
```
lucca30, line 1277: As discussed with @pratikspatil024 and @manav2401, we must keep state.Copy() on LastStateId call in order to avoid bad blocks. 

The solution so was copying the new witness into the witness of the original state.

lucca30: My only concern on doing it is that `AddCode` function actually requires the code on it byte state. So to do that I need to first convert the string on map back to bytes and also insert it one by one. Which in the end demands more lines of code and delivers the same output since while copying the state we do also create a deep copy of witness.

https://github.com/maticnetwork/bor/blob/d7dd23424bde03204110fdd206dc9ad02ecf1aec/core/stateless/witness.go#L88-L108

cffls: Ah I see, since `tempState` is a copy of the original witness, it would already contain the existing state and code, so it is okay to override here.

PR #1525, opened 2025-04-24
Djpolygon/packager updates

djpolygon:
# Description

Updated packager to use make_latest: false, instead of prerelease: true for releasing artifacts
Updated packagers to use commit pinned versions of other actions

# Changes

- [ x] Bugfix (non-breaking change that solves an issue)

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


rekibnikufesin (approved):

go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/gballet/go-libpcsclite v0.0.0-20191108122812-4678299bea08
 	github.com/gofrs/flock v0.8.1
-	github.com/golang-jwt/jwt/v4 v4.5.1
+	github.com/golang-jwt/jwt/v4 v4.5.2
```
cffls, line 37: Why is this required?

params/version.go
```diff
@@ -21,9 +21,9 @@ import (
 )
 
 const (
-	VersionMajor = 1  // Major version component of the current release
-	VersionMinor = 5  // Minor version component of the current release
-	VersionPatch = 5  // Patch version component of the current release
+	VersionMajor = 2  // Major version component of the current release
+	VersionMinor = 0  // Minor version component of the current release
+	VersionPatch = 1  // Patch version component of the current release
```
cffls, line 26: Patch version should be 2 for this branch?

.github/workflows/amoy_deb_profiles.yml
```diff
@@ -382,10 +382,10 @@ jobs:
         run: ls -ltr packaging/deb/ | grep bor
 
       - name: Release bor Packages
-        uses: softprops/action-gh-release@v2
+        uses: softprops/action-gh-release@da05d552573ad5aba039eaac05058a918a7bf631
```
marcello33, line 385: Any reason why we're using a commit hash instead of a stable release?

.github/workflows/mainnet_deb_profiles.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
```
marcello33, line 21: same here and for action-gh-release
Any reason why we're using a commit hash instead of a stable release?

go.mod
djpolygon: It is not, this came from some clean up. Will address

.github/workflows/amoy_deb_profiles.yml
djpolygon: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions

.github/workflows/mainnet_deb_profiles.yml
djpolygon: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions

.github/workflows/amoy_deb_profiles.yml
marcello33: Yeah I see. Not sure tho if we should not largely used actions such as `actions/checkout`, which are referenced everywhere with release versions.
Using commits SHA is inconvenient imho when it comes to tracking and updates, as it's not clear what version is being used, and it makes more complicated for 3rd party security tools to detect if that version is affected by sec vulns.

djpolygon: Ok will revert back on this PR and the heimdall PR.

marcello33: I mean it's basically harder to read and maintain (we don’t know which version or features we're using without looking it up). True that we have the risks of backdoor, but I think it's so minimal, and also pinning a commit SHA won't get us with security patches or improvements, unless we manually check for updates and bump the SHA (this can lead to technical debt if forgotten). Not sure, it feels like an overkill for "trusted" actions like actions/checkout, since GitHub maintains them "securely".
Up to you, just my 2 cents.

djpolygon:
Going to close this and reopen to prevent some conflicts

PR #1533, opened 2025-05-08
Sync block with witness in full sync

cffls:
# Description

This PR will enable a full node to use witness to sync a block. 

# Changes

- [X] New feature (non-breaking change that adds functionality)

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -670,13 +670,14 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header, com
 		processorCount++
 
 		go func() {
+			var witness *stateless.Witness
 			if computeWitness {
-				witness, err := stateless.NewWitness(block.Header(), bc)
+				witness, err = stateless.NewWitness(block.Header(), bc)
 				if err != nil {
 					log.Error("error in witness generation", "caughterr", err)
 				}
-				statedb.StartPrefetcher("chain", witness)
 			}
+			statedb.StartPrefetcher("chain", witness)
 
 			pstart := time.Now()
 			res, err := bc.processor.Process(block, statedb, bc.vmConfig, nil, ctx)
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -306,7 +306,6 @@ type BlockChain struct {
 	stateSyncData    []*types.StateSyncData                  // State sync data
 	stateSyncFeed    event.Feed                              // State sync feed
 	chain2HeadFeed   event.Feed                              // Reorg/NewHead/Fork data feed
-	computeWitness   bool                                    // Whether to compute witnesses
 }
 
 // NewBlockChain returns a fully initialised block chain using information
@@ -733,10 +732,6 @@ func (bc *BlockChain) empty() bool {
 	}
 
 	return true
-}
-
-func (bc *BlockChain) SetComputeWitness(computeWitness bool) {
-	bc.computeWitness = computeWitness
 }
 
 // loadLastState loads the last known chain state from the database. This method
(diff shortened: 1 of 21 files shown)


core/blockchain.go
```diff
@@ -671,12 +670,14 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header) (_
...
 		go func() {
-			witness, err := stateless.NewWitness(block.Header(), bc)
-			if err != nil {
-				log.Error("error in witness generation", "caughterr", err)
+			if computeWitness {
+				witness, err := stateless.NewWitness(block.Header(), bc)
+				if err != nil {
+					log.Error("error in witness generation", "caughterr", err)
+				}
+				statedb.StartPrefetcher("chain", witness)
```
lucca30, line 678: `statedb.StartPrefetcher` is always called, even when we are not computing a witness. The way we decide between computing or not is by passing `witness == nil`.

See reference from main branch:

https://github.com/maticnetwork/bor/blob/84fd09a04c86af63626ca59c0cb64b315b454911/core/blockchain.go#L657-L664

core/blockchain.go
```diff
@@ -2567,7 +2537,19 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness
 
 		// Process block using the parent state as reference point
 		pstart := time.Now()
-		receipts, logs, usedGas, statedb, vtime, err := bc.ProcessBlock(block, parent)
+
+		computeWitness := bc.computeWitness
+
+		if witnesses != nil && len(witnesses) > it.processed()-1 && witnesses[it.processed()-1] != nil {
+			memdb := witnesses[it.processed()-1].MakeHashDB()
+			bc.statedb.TrieDB().SetReadBackend(hashdb.New(memdb, triedb.HashDefaults.HashDB))
```
lucca30, line 2545: It seems `bc.computeWitness` have the same meaning of `makeWitness bool` in the function params. It would may be better either pass `bc.computeWitness` in this param on function call or remove it from the function signature.

eth/handler.go
```diff
@@ -168,8 +171,11 @@ func newHandler(config *handlerConfig) (*handler, error) {
 		quitSync:            make(chan struct{}),
 		handlerDoneCh:       make(chan struct{}),
 		handlerStartCh:      make(chan struct{}),
+		syncWithWitnesses:   config.syncWithWitnesses,
 	}
 
+	log.Info("Sync with witnesses", "enabled", config.syncWithWitnesses)
+
 	if config.Sync == downloader.StatelessSync {
```
lucca30, line 179: Doesn't `config.Sync == downloader.StatelessSync` means the same thing of `config.syncWithWitnesses`. Is there any scenario of one being true and the other dont?

core/blockchain.go
cffls: Make sense. Will change to passing nil.

eth/handler.go
cffls: No, they meant differently. When `downloader.StatelessSync` is `true`, the node won't keep any local state. However, when `config.syncWithWitnesses` is `true` but `downloader.StatelessSync` is `false`, the node *will* use witness for block execution but also keep and update its local state.

core/blockchain.go
cffls: Good observation. I thought about this too, but after spending some time trying, I realized that removing `makeWitness` will touch a lot of functions, especially those in `catalyst`, and it may become more difficult to merge from upstream in the future.

lucca30: Yeah, I see your point. Specially because of upstream merges, that can also be painful with this change.

Maybe passing `bc.computeWitness` as the parameter of the function call can help avoiding two variables with same behavior on this function. In this way whenever an upstream comes with a new functionality when `makeWitness` is true will also reflect in our code.

eth/handler.go
lucca30: That's an interesting behavior. I see your point.

core/blockchain.go
cffls: Managed to remove computeWitness. Thanks for your suggestion!

lucca30 (approved):

PR #1536, opened 2025-05-09
[bug fix] Avoid concurrent access to hasher buffer

cffls:
# Description

This is to fix hasher panic caused by a recent upstream merge.

```
/home/runner/work/bor/bor/core/state/reader.go:82 +0x6b
github.com/ethereum/go-ethereum/core/state.(*stateReader).Account(0xc101562400, {0xfb, 0xf8, 0xb7, 0xd2, 0xdd, 0x10, 0x44, 0x7d, 0xcb, ...})
/home/runner/work/bor/bor/crypto/crypto.go:92 +0x6b
github.com/ethereum/go-ethereum/crypto.HashData({0x7d7d9c1d3288, 0xc0b5bb81e0}, {0xc19359c558, 0x14, 0x14})
/home/runner/go/pkg/mod/golang.org/x/crypto@v0.31.0/sha3/sha3.go:117 +0x1c5
golang.org/x/crypto/sha3.(*state).Write(0xc138702f70?, {0xc19359c558?, 0x14?, 0x29edcc0?})
goroutine 362766824 [running]:
panic: sha3: Write after Read
```

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/state/reader.go
+++ b/core/state/reader.go
@@ -79,7 +79,7 @@ func newStateReader(root common.Hash, snaps *snapshot.Tree) (*stateReader, error
 //
 // The returned account might be nil if it's not existent.
 func (r *stateReader) Account(addr common.Address) (*types.StateAccount, error) {
-	ret, err := r.snap.Account(crypto.HashData(r.buff, addr.Bytes()))
+	ret, err := r.snap.Account(crypto.HashData(crypto.NewKeccakState(), addr.Bytes()))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


core/state/reader.go
```diff
@@ -79,7 +79,7 @@ func newStateReader(root common.Hash, snaps *snapshot.Tree) (*stateReader, error
 //
 // The returned account might be nil if it's not existent.
 func (r *stateReader) Account(addr common.Address) (*types.StateAccount, error) {
-	ret, err := r.snap.Account(crypto.HashData(r.buff, addr.Bytes()))
+	ret, err := r.snap.Account(crypto.HashData(crypto.NewKeccakState(), addr.Bytes()))
```
lucca30, line 82: By allocating a new buf each new account call wouldn't it has a significantly cpu or allocation cost (which also impacts on cpu because of gc)?

Because of the bug fix nature, we can keep this way, but maybe it's interesting create a ticket or left a TODO to replace it with a "buf pool". In this way we manage all buf availables and instead of deallocating we actually reset it.

lucca30 (approved):
Besides the CPU costs (which can be small), LGTM.

core/state/reader.go
cffls: It is okay becuase NewKeccakState is optimized: https://github.com/maticnetwork/bor/blob/573c343d54105db2510cbfea1a41ec91c17c6139/crypto/crypto.go#L77-L80

lucca30: Humm, that's true. Nice

ssandeep (approved):

PR #1538, opened 2025-05-13
miner: reduce commit interrupt timer for on-time block announcement

manav2401:
# Description

Previously, the commit interrupt timer triggers exactly at header's timestamp i.e. 2s (for non-sprint end blocks proposed by primary producer) to stop block building and announce the block to rest of the peers. 

There's some overhead after this timer is triggered which is consensus checks and state root calculation which isn't accounted in the 2s delay. On average, 300ms are spent on such operations (this is just a rough average from a sentry in a limited time range). This leads to delay in block announcement which may lead to reorgs in some scenarios (because the block reaches late to rest of the validators and the secondary producer kicks in). 

This PR reduces that delay and leaves a 500ms buffer for these misc operations before block is announced. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1557,8 +1557,10 @@ func resetAndCopyInterruptCtx(interruptCtx context.Context) context.Context {
 }
 
 func getInterruptTimer(interruptCtx context.Context, number, timestamp uint64) (context.Context, func()) {
-	delay := time.Until(time.Unix(int64(timestamp), 0))
-	interruptCtx, cancel := context.WithTimeout(interruptCtx, delay)
+	// Instead of interrupting block building at exact header time, we leave 500ms as buffer
+	// for consensus operations (in Finalize) and state root calculation.
+	adjustedDelay := time.Unix(int64(timestamp), 0).Add(-500 * time.Millisecond)
+	interruptCtx, cancel := context.WithTimeout(interruptCtx, time.Until(adjustedDelay))
 
 	go func() {
 		<-interruptCtx.Done()
--- a/miner/worker_test.go
+++ b/miner/worker_test.go
@@ -17,6 +17,7 @@
 package miner
 
 import (
+	"fmt"
 	"math/big"
 	"os"
 	"sync/atomic"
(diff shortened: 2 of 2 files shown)


pratikspatil024 (approved):

lucca30:
So just to understand better the behavior: the bigger is the time we reduce, the shorter is the time we have to produce a block. Which stands for more chance to empty blocks.

So the idea here is to choose the best time reduction in order to give the producer time to build the block but also give time to other peers to receive this block before accepting other blocks leading to a reorg. Right?

By applying this 500ms reduction, do you see any increase on empty blocks?

miner/worker.go
```diff
@@ -1557,8 +1557,10 @@ func resetAndCopyInterruptCtx(interruptCtx context.Context) context.Context {
 }
 
 func getInterruptTimer(interruptCtx context.Context, number, timestamp uint64) (context.Context, func()) {
-	delay := time.Until(time.Unix(int64(timestamp), 0))
-	interruptCtx, cancel := context.WithTimeout(interruptCtx, delay)
+	// Instead of interrupting block building at exact header time, we leave 500ms as buffer
+	// for consensus operations (in Finalize) and state root calculation.
+	adjustedDelay := time.Unix(int64(timestamp), 0).Add(-500 * time.Millisecond)
```
cffls, line 1562: Does it makes sense to make it a configurable parameter with default of 500ms? The use case may be that some higher spec machine will require less time to compute state root and they can configure this accordingly.

manav2401:
@lucca30 yes you're right. We need to choose an ideal time so that there's sufficient time to build new blocks and also the block reaches on time to other peers (validators specifically). 

I haven't been able to deploy it on a validator so I don't know the exact numbers but I do know that block building happens pretty fast as of now (specifically for validators running good machines) unless the tx pool takes time. Although the metric I captured yesterday was p2p delay (i.e. after how long was the block announced than it's ideal time) and the average seems to be around 1s. This change will surely reduce that.

miner/worker.go
manav2401: Yeah totally makes sense. I will do that change.

PR #1545, opened 2025-05-15
consensus/bor: allow early block announcements

manav2401:
# Description

This PR enables [PIP-66](https://github.com/maticnetwork/Polygon-Improvement-Proposals/blob/main/PIPs/PIP-66.md). It allows primary block producers to announce blocks early before the header time for optimal time utilisation and preventing reorgs and p2p delays. 

Checklist before merging
- [x] Address comments in Prepare and Verify headers
- [x] Add unit + integration tests (+fix existing tests if possible)
- [x] Add check for bhilai HF

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


consensus/bor/bor.go
```diff
@@ -349,8 +349,9 @@ func (c *Bor) verifyHeader(chain consensus.ChainHeaderReader, header *types.Head
 
 	number := header.Number.Uint64()
 
-	// Don't waste time checking blocks from the future
-	if header.Time > uint64(time.Now().Unix()) {
+	// Don't waste time checking blocks from the future but allow a buffer of block time for
+	// early block announcements.
+	if header.Time-c.config.CalculatePeriod(number) > uint64(time.Now().Unix()) {
```
manav2401, line 354: I believe we should check if the validator is primary here or not. 

Because with `period` set to 2s, non primary validators could also send blocks 2s early than their expected time. Wdyt?

consensus/bor/bor.go
```diff
@@ -827,6 +828,15 @@ func (c *Bor) Prepare(chain consensus.ChainHeaderReader, header *types.Header) e
...
 		header.Time = uint64(time.Now().Unix())
+	} else {
+		// For primary validators, wait until previous block time is completed before starting
+		// to produce next block
+		if succession == 0 {
+			startTime := time.Unix(int64(header.Time-c.config.CalculatePeriod(number)), 0)
+			if time.Now().Before(startTime) {
+				time.Sleep(time.Until(startTime))
+			}
+		}
```
manav2401, line 839: I believe we can avoid this section of code altogether. From what I understand, it's added to avoid bor to start building block before time. But I think that there's no harm in doing that because even if it starts and finishes before the actual block building window, the delay in seal function would avoid announcing block before `header.Time - 2s` timer hits. So, we'll anyways wait to announce block early. Wdyt?

consensus/bor/bor.go
manav2401: Will need access to snapshot (the one used in bor consensus) to find out the succession number of proposer and it's fetched very late in this `verifyHeader` call so we'll probably have to move this check after that instead of keeping it here.

consensus/bor/bor.go
```diff
@@ -1021,8 +1041,20 @@ func (c *Bor) Seal(chain consensus.ChainHeaderReader, block *types.Block, result
...
+		delay = time.Until(time.Unix(int64(header.Time), 0)) // Wait until we reach header time for non-primary validators
+		if successionNumber == 0 {
+			// For primary producers, set the delay to `header.Time - block time` instead of `header.Time`
+			// for early block announcement instead of waiting for full block time.
+			delay = time.Until(time.Unix(int64(header.Time-c.config.CalculatePeriod(number)), 0))
+		}
+	} else {
+		delay = time.Until(time.Unix(int64(header.Time), 0)) // Wait until we reach header time
+	}
+
```
lucca30, line 1057: I'm not familiar with the tame taken by `Sign()`

https://github.com/maticnetwork/bor/blob/66a1428048c25f2121428da375208b05f84d2ed9/consensus/bor/bor.go#L1061-L1063

But if it can potentially be a relevant time, I would suggest you calculating the `delay` right before waiting, so there is no risk on waiting more than expected. (Since some time can potentially pass while on `Sign()`)

___

Another approach could be instead of calculating `delay` you can calculate the `timeToSeal` so you can replace `<-time.After(delay)` by `<-time.After(time.Until(timeToSeal))`

Reference:

https://github.com/maticnetwork/bor/blob/66a1428048c25f2121428da375208b05f84d2ed9/consensus/bor/bor.go#L1075-L1079

manav2401: @lucca30 signing function is super fast and takes <1ms (also it's deterministic math library so I don't think there's any possibility of delay there). 

The `timeToSeal` approach is conditional (based on whether the HF is enabled or not, whether producer is primary or non-primary) and it will make the select thing hard to read. Assigning a delay seems readable and more clear for now.

lucca30: nice!

lucca30 (approved):

marcello33 (approved):

cffls:
A few tests have failed. Not sure if it is from the base branch though.

PR #1550, opened 2025-05-22
Pectra 2.0: upstream merge v1.15.7<>v1.15.11

pratikspatil024:
# Description

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


core/filtermaps/chain_view.go
```diff
@@ -122,10 +122,14 @@ func (cv *ChainView) RawReceipts(number uint64) types.Receipts {
 
 // SharedRange returns the block range shared by two chain views.
 func (cv *ChainView) SharedRange(cv2 *ChainView) common.Range[uint64] {
+	if cv == nil || cv2 == nil {
+		return common.Range[uint64]{}
+	}
+
```
lucca30, line 128: You are anticipating the `cv` nil check in order to avoid panic on `cv.lock.Lock()` call, right?

If it's not also reproduced on geth, should be interesting to share with them. Just not sure if `cv2` nil check is also necessary in this part. But it works anyway.

core/genesis.go
```diff
@@ -518,18 +518,15 @@ func (g *Genesis) toBlockWithRoot(root common.Hash) *types.Block {
 			head.BaseFee = new(big.Int).SetUint64(params.InitialBaseFee)
 		}
 	}
-
-	var withdrawals []*types.Withdrawal
+	var (
```
cffls: Why was this commit reverted?

cffls (commented):
Looks good! Left one question about reverting the commit where the withdraw is disabled in genesis.

core/genesis.go
pratikspatil024: Ideally we should not revert this, as it will give issues when using a custom genesis. But after making this change, there were a lot of unit tests that were failing and it was taking a lot of time to fix them.
So we decided to revert this for now. Created a [task](https://polygon.atlassian.net/browse/POS-3071) to track it. Thank you!

cffls: Got it, thanks!

cffls (approved):

PR #1551, opened 2025-05-23
FastForward on HeimdallV1

lucca30:
# Description

Implement a one-time fast-forward in the stateless client (bor-connected heimdall-v1) for tests by setting TotalDifficulty to the target block number on startup instead of fetching real TD and skipping formal validations. If it ever desyncs, simply restart to rerun the fast-forward.

# Changes


cffls (approved):
lgtm, thanks!

PR #1553, opened 2025-05-26
Witness Prefetching on Block Production

lucca30:
# Description

This changes aims to introduce Witness Prefetching right on producer nodes, reducing latency for stateless nodes and full nodes syncing via witness. In order to reach that we started pushing the witness on the `Seal` phase of consensus (something similar of what we already do in `FinalizeAndAssemble`). 

# Changes

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -769,8 +769,8 @@ func (h *handler) minedBroadcastLoop() {
 				delay := common.PrettyDuration(time.Millisecond * time.Duration(delayInMs))
 				log.Info("[block tracker] Broadcasting mined block", "number", ev.Block.NumberU64(), "hash", ev.Block.Hash(), "blockTime", ev.Block.Time(), "now", time.Now().Unix(), "delay", delay, "delayInMs", delayInMs)
 			}
-			h.BroadcastBlock(ev.Block, nil, true)  // First propagate block to peers
-			h.BroadcastBlock(ev.Block, nil, false) // Only then announce to the rest
+			h.BroadcastBlock(ev.Block, ev.Witness, true)  // First propagate block to peers
+			h.BroadcastBlock(ev.Block, ev.Witness, false) // Only then announce to the rest
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


eth/handler.go
```diff
@@ -769,8 +769,8 @@ func (h *handler) minedBroadcastLoop() {
 				delay := common.PrettyDuration(time.Millisecond * time.Duration(delayInMs))
 				log.Info("[block tracker] Broadcasting mined block", "number", ev.Block.NumberU64(), "hash", ev.Block.Hash(), "blockTime", ev.Block.Time(), "now", time.Now().Unix(), "delay", delay, "delayInMs", delayInMs)
 			}
-			h.BroadcastBlock(ev.Block, nil, true)  // First propagate block to peers
-			h.BroadcastBlock(ev.Block, nil, false) // Only then announce to the rest
+			h.BroadcastBlock(ev.Block, ev.Witness, true)  // First propagate block to peers
```
cffls, line 772: Do you think we should only pass this when witness protocol is enabled on the producer?

lucca30: I don't think so. Because only when the wit protocol is true we have `peersWithoutWitness`, making the code actually share the witness with other peers. But something we can actually think of is making the witness production itself optional for the producer. But it also sounds to not be interesting for us on [PIP-64](https://forum.polygon.technology/t/pip-64-validator-elected-block-producer/20918)

https://github.com/maticnetwork/bor/blob/ec76017637069f2a017881c06c9abc00a5dec1e5/eth/handler.go#L614-L620

https://github.com/maticnetwork/bor/blob/ec76017637069f2a017881c06c9abc00a5dec1e5/eth/handler.go#L663-L666

lucca30: Sure. So I'll include this in the same way we do on rpc nodes.

lucca30: Done on 6a72857 @cffls

cffls:
Looks good overall, thanks! I think the changes might break a few tests. Might be a good idea to merge from latest `stateless_sync` and rerun all unit tests.

cffls (approved):

PR #1556, opened 2025-05-29
Disable tx pool for stateless client

lucca30:
# Description

To handle this scenario I included it as `nil` on handler and properly handled each call scenario. The only one I had to generate an actual new behavior was the one regarding adding new transactions, which raises a new error.

# Changes


cffls (approved):
lgtm, thanks!

PR #1558, opened 2025-05-30
Prune old witnesses

lucca30:
# Description

This PR adds a background goroutine that wakes every h.witnessPruneInterval to prune any witnesses older than the retention threshold. A persistent database cursor ensures each run picks up exactly where the last left off—avoiding redundant scans—and bounds the pruning window. All eligible witness hashes are deleted in a single batched transaction, the cursor is advanced atomically, and the loop cleanly shuts down when the quit channel is closed.

# Changes

### DIFF
--- a/core/rawdb/accessors_state.go
+++ b/core/rawdb/accessors_state.go
@@ -282,3 +282,28 @@ func WriteWitness(db ethdb.KeyValueWriter, blockHash common.Hash, witness []byte
 		log.Crit("Failed to store witness", "err", err)
 	}
 }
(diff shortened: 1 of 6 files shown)


cffls (commented):
Looks good overall! Some unit tests will be awesome :D

lucca30:
Thanks @cffls .

I did a small refactor in order to make the code more testable. While doing it I do also included a few new mocks that will help us in future tests.

cffls (approved):

PR #1560, opened 2025-06-02
Fix ancient pruning check on stateless start

lucca30:
# Description

Freeze db has a check when it founds it as empty while the head hash is not matching the genesis hash. In this scenario geth checks if #1 block exists or not on database. This fix allows this scenario for stateless client since this client usually fast forward the block #1.

# Changes

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -483,10 +483,10 @@ func (c *Bor) verifyCascadingFields(chain consensus.ChainHeaderReader, header *t
 	// TODO:
 	// Verify validator statelessly
 
-	snap, err := c.snapshot(chain, number-1, header.ParentHash, parents)
-	if err != nil {
-		return err
-	}
+	// snap, err := c.snapshot(chain, number-1, header.ParentHash, parents)
+	// if err != nil {
+	// 	return err
+	// }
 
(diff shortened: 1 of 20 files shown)


consensus/bor/bor.go
```diff
@@ -525,7 +525,7 @@ func (c *Bor) verifyCascadingFields(chain consensus.ChainHeaderReader, header *t
 	}
 
 	// verify the validator list in the last sprint block
-	if IsSprintStart(number, c.config.CalculateSprint(number)) {
+	if false && IsSprintStart(number, c.config.CalculateSprint(number)) {
```
petejkim, line 528: is this something intentional that we want to keep in prod? @cffls

cffls: No, this won't be in prod. If you look at the latest version of `stateless_sync` branch, this chunk of code (checking validator set in the header) has been removed already. Instead of the relying on the valset from the header, a node will directly fetch the snapshot of valset from [span store](https://github.com/maticnetwork/bor/blob/stateless_sync/consensus/bor/span_store.go) and check if the signer of the incoming block is valid [here](https://github.com/maticnetwork/bor/blob/stateless_sync/consensus/bor/bor.go#L781-L795).

PR #1563, opened 2025-06-04
Allow mining if DevFakeAuthor is enabled

Raneet10:
# Description

In case of public networks, the worker doesn't start mining if the node isn't connected to peers. For the purposes of shadow forking, it is necessary to allow the node to mine blocks in its own fork whilst not listening to peers. This PR enables this. 

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -594,12 +594,13 @@ func (w *worker) mainLoop() {
 		}
 	}()
 
-	bor, _ := w.engine.(*bor.Bor)
+	bor, isBor := w.engine.(*bor.Bor)
+	devFakeAuthor := isBor && bor != nil && bor.DevFakeAuthor
 	for {
 		select {
 		case req := <-w.newWorkCh:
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -594,11 +594,12 @@ func (w *worker) mainLoop() {
 		}
 	}()
 
+	bor, _ := w.engine.(*bor.Bor)
```
praetoriansentry: It might be worth checking the type assertion here. E.g.

```golang
bor, isBor := w.engine.(*bor.Bor)
devFakeAuthor := isBor && bor != nil && bor.DevFakeAuthor
// ....
if w.eth.PeerCount() > 0 || devFakeAuthor {
      // ...
}

```

Raneet10: This did occur to me but I didn't pay much heed because we don't expect any consensus engine other than bor running.

Raneet10: But on second thoughts this looks neater.

ssandeep (approved):

cffls (approved):

PR #1567, opened 2025-06-05
VeBlop span rotation

cffls:
# Description

This PR introduces VeBlop span rotation.

Major changes:
* Today, each span is deterministically written to a contract for future blocks. After VeBlop, the span will be proactively retrieved from heimdall when the current block producer fails to deliver new blocks within a certain time range, which is currently set at 4s. 
* Remove span commit and validator set info in header after VeBlop, since the span info will be always retrieved from heimdall dynamically.

# Changes

# Breaking changes

A HF block number is introduced.

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/consensus/bor/span_store.go
+++ b/consensus/bor/span_store.go
@@ -3,6 +3,8 @@ package bor
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus/bor/heimdall/span"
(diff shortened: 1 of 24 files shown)


consensus/bor/span_store.go
```diff
@@ -110,60 +192,117 @@ func (s *SpanStore) spanById(ctx context.Context, spanId uint64) (*span.Heimdall
...
-func estimateSpanId(blockNumber uint64) uint64 {
-	if blockNumber > zerothSpanEnd {
+func (s *SpanStore) estimateSpanId(blockNumber uint64) uint64 {
+	if blockNumber > zerothSpanEnd && blockNumber > 0 {
+		if s.lastUsedSpan.Load() != nil {
+			lastUsedSpan := s.lastUsedSpan.Load()
+			startBlock := lastUsedSpan.Span.StartBlock
+			endBlock := lastUsedSpan.Span.EndBlock
+			if blockNumber > endBlock {
+				return lastUsedSpan.Span.ID + 1 + (blockNumber-endBlock)/defaultSpanLength
```
lucca30: Since the `startBlock` and `endBlock` are included in the span, I think it's missing a `-1` in the distance calculation `(blockNumber-endBlock)`

To make it clear, let's assume the following span intervals:

```
ID X:   [10,19]
ID X+1: [20,29]
(defaultSpanLength == 10)
```
So calculating for `block #29`:
```
X + 1 + (29 - 19)/10
X + 1 + 1
```
But it should be `X+1`.

So I'm assuming the calculation here should be:
```
return lastUsedSpan.Span.ID + 1 + (blockNumber-endBlock-1)/defaultSpanLength
```

consensus/bor/span_store.go
```diff
@@ -110,60 +192,117 @@ func (s *SpanStore) spanById(ctx context.Context, spanId uint64) (*span.Heimdall
...
+			} else if blockNumber < startBlock {
+				// Calculate how many spans to go back. (startBlock - blockNumber + defaultSpanLength - 1) / defaultSpanLength is ceil((startBlock - blockNumber)/defaultSpanLength)
+				spansToDecrement := (startBlock - blockNumber + defaultSpanLength - 1) / defaultSpanLength
```
lucca30: To keep it similar to the calculation of the `blockNumber > endBlock` case, I think it's good put `defaultSpanLength` outside the division.

```
spansToDecrement := 1+ (startBlock - blockNumber - 1) / defaultSpanLength
```

consensus/bor/span_store.go
```diff
@@ -204,3 +346,39 @@ func getMockSpan0(ctx context.Context, spanner Spanner, chainId string) (*span.H
...
+
+// Close cancels the background routine and cleans up resources
+func (s *SpanStore) Close() {
+	if s.cancel != nil {
+		s.cancel()
+	}
+}
+
+// Wait for a new span whose selected producers are different from the current header author
+func (s *SpanStore) waitForNewSpan(ctx context.Context, targetBlockNumber uint64, currentHeaderAuthor common.Address) (bool, error) {
```
lucca30: We need a built-in timeout here to ensure this function can’t block indefinitely. Instead of forcing every caller to wrap waitForNewSpan in context.WithTimeout, let’s accept a timeout time.Duration and create the cancellable context inside the function itself. For example:

```
func (s *SpanStore) waitForNewSpan(context context.Context, targetBlockNumber uint64, currentHeaderAuthor common.Address, timeout time.Duration) (bool, error) {
ctx, cancel := context.WithTimeout(context.Background(), timeout)
defer cancel()
```
This way, callers simply pass in a timeout value, and waitForNewSpan guarantees it will return (with an error) if it takes longer than timeout.

consensus/bor/span_store.go
cffls: Great catch and makes sense to me. Corrected the code according to your suggestions.

consensus/bor/span_store.go
cffls: Thanks. This is a great suggestion and made the code cleaner. Corrected accordingly.

lucca30 (approved):
lgtm, thanks!

PR #1570, opened 2025-06-10
Refactor and optimise commit interrupt

manav2401:
# Description

Currently, in bor, we use an interrupt context to notify the miner and EVM modules to stop block building when we hit the 2s mark to prevent delayed block announcement. While the check is pretty trivial, we do it before running every OPCODE which can lead to it using good chunk of CPU when processing transactions. An internal benchmark revealed that it used around 40% of CPU (out of the total CPU used in block building loop) on increased (~100M) gas limit block. 

This is because we check for `context.Done()` before running every opcode which can be non-trivial for a heavy block with too many transactions (and a lot more opcodes). We instead use a simple global atomic flag to toggle when the block building time is up and use this global flag in the EVM interpreter and miner loop to check if we want to interrupt or continue ahead. 

This PR also refactors how commit interrupt is handled in worker - evm interaction and simplifies it to a very good extent. 

Here's a very minimalistic benchmark for both the calls.

Context done approach:
```go
func BenchmarkCtxDone(b *testing.B) {
	ctx := context.Background()
	ctx, cancel := context.WithTimeout(ctx, time.Hour)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		// Non‐blocking check of ctx.Done()
		select {
		case <-ctx.Done():
			// (won't happen in this benchmark)
		default:
		}
	}
	cancel()
}
```

Results:
```
cpu: VirtualApple @ 2.50GHz
BenchmarkCtxDone-8      100000000               10.34 ns/op            0 B/op          0 allocs/op
```

Atomic flag approach
```go
func BenchmarkAtomic(b *testing.B) {
	ctx := context.Background()
	ctx, cancel := context.WithTimeout(ctx, time.Hour)

	var done atomic.Bool

	go func() {
		time.Sleep(time.Hour)
		done.Store(true)
	}()

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if done.Load() {
			b.Fatal("context cancelled")
		}
	}
	cancel()
}
```

Results:
```
cpu: VirtualApple @ 2.50GHz
BenchmarkAtomic-8       1000000000               0.4130 ns/op          0 B/op          0 allocs/op
```

Time for each operation is 0.4130 ns in this approach v/s 10.34 ns in the previous one.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### Manual tests

Tested it on shadow fork. 

# Additional comments

### DIFF
--- a/miner/worker_test.go
+++ b/miner/worker_test.go
@@ -93,11 +93,6 @@ func testGenerateBlockAndImport(t *testing.T, isClique bool, isBor bool) {
 	chain, _ := core.NewBlockChain(rawdb.NewMemoryDatabase(), nil, b.genesis, nil, engine, vm.Config{}, nil, nil, nil)
 	defer chain.Stop()
 
-	// Ignore empty commit here for less noise.
-	w.skipSealHook = func(task *task) bool {
-		return len(task.receipts) == 0
-	}
-
 	// Wait for mined blocks.
 	sub := w.mux.Subscribe(core.NewMinedBlockEvent{})
 	defer sub.Unsubscribe()
@@ -286,7 +281,7 @@ func (b *testWorkerBackend) newRandomTxWithNonce(creation bool, nonce uint64) *t
 	gasPrice := big.NewInt(100 * params.InitialBaseFee)
 
 	if creation {
-		tx, _ = types.SignTx(types.NewContractCreation(b.txPool.Nonce(TestBankAddress), big.NewInt(0), testGas, gasPrice, common.FromHex(testCode)), types.HomesteadSigner{}, testBankKey)
+		tx, _ = types.SignTx(types.NewContractCreation(b.txPool.PoolNonce(TestBankAddress), big.NewInt(0), testGas, gasPrice, common.FromHex(testCode)), types.HomesteadSigner{}, testBankKey)
 	} else {
 		tx, _ = types.SignTx(types.NewTransaction(nonce, testUserAddress, big.NewInt(1000), params.TxGas, gasPrice, nil), types.HomesteadSigner{}, testBankKey)
 	}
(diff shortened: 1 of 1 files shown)


common/commit_interrupt.go
```diff
@@ -0,0 +1,9 @@
+package common
+
+import "sync/atomic"
+
+// InterruptBlockBuilding is a global flag which will be set to true to notify
+// the miner to stop building the current block. It's also used in EVM interpreter
+// to interrupt an ongoing transaction execution. It's reset to false when block
+// building time of the next block starts.
+var InterruptBlockBuilding atomic.Bool
```
cffls: It seems a bit dangerous to use a global variable. Why can't this be created per mining task?

manav2401: It's because the EVM interpreter is very minimalistic and doesn't depend on miner. It creates a cyclic dependency if we use anything from miner in it. We could pass a pointer but that would lead to changing function definition at multiple places. Because the commit transactions loop is called only once, I though a global variable should be safe to use. If you have any other idea, happy to explore that approach.

cffls: It seems to me that the approach without using global variable is to directly change the interrupt context type [here](https://github.com/maticnetwork/bor/blob/0cebe7369811af60bb5cc01724f79d154f201a7c/miner/worker.go#L270). This would involve some type changes but won't be too much I guess? 
If going with global variable approach, the interruptCtx would basically become unless right?

manav2401: I don't think we'll be able to completely remove the `interruptCtx` as we still need to check for `Done()` in the EVM interpreter but let me see if there's some other workaround for that.

manav2401: Refactored a bunch of things which also handles this concern. Thanks!

cffls (approved):
Thanks for refactoring & simplifying! Looks pretty good to me.

lucca30 (approved):
lgtm, thanks!

PR #1573, opened 2025-06-11
Auto rewind blocks when their span rotates

cffls:
# Description

After veblop, a previously imported/produced block could become invalid, due to dynamic span rotation. This PR will automatically detect when this happens and rewind invalid blocks.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -303,6 +303,7 @@ type BlockChain struct {
 	stateSyncFeed    event.Feed                              // State sync feed
 	chain2HeadFeed   event.Feed                              // Reorg/NewHead/Fork data feed
 	chainSideFeed    event.Feed                              // Side chain data feed (removed from geth but needed in bor)
+	checker          ethereum.ChainValidator
 }
 
 // NewBlockChain returns a fully initialised block chain using information
@@ -355,6 +356,7 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, genesis *Genesis
 
 		borReceiptsCache: lru.NewCache[common.Hash, *types.Receipt](receiptsCacheLimit),
 		logger:           vmConfig.Tracer,
+		checker:          checker,
 	}
 
 	var err error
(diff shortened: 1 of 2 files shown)


core/blockchain.go
```diff
@@ -3501,3 +3506,99 @@ func (bc *BlockChain) ProcessBlockWithWitnesses(block *types.Block, witness *sta
...
+		log.Info("Started header verification loop")
+
+		for {
+			select {
+			case <-bc.quit:
+				log.Info("Stopping header verification loop")
+				return
+			case <-ticker.C:
+				bc.verifyPendingHeaders()
+			}
```
lucca30, line 3533: Even covered by a high frequent check of 2s, it's possible that `bc.verifyPendingHeaders()` can run slow forcing the next check to happen in an interval bigger than 2s.

In this scenario I suppose we would potentially "jump" the verification of 1 milestone (for example we checked milestone **B** and after 10s we verified milestone **B+2**). 

So what I want to clarify is if there is any problem by "jumping" the validation of one milestone. If so, you should consider always storing the `lastValidNumber` and continuously validating from it's value instead of querying milestone itself.

(from the milestone nature, I don't believe this is an actual issue since the milestone already means the block is verified. So I'm just double checking this with you)

cffls: This logic here is to address the problem when no milestone could be settled for pending blocks, so the case you mentioned above won't be the responsibility of this piece of code to resolve.

Jumping the milestone won't become an issue either. Consider the case below:

```
Milestone 1: 2-4 (producer1)
Milestone 2: 5-6 (producer2)
Milestone 3: 7-10 (producer1)
```

If a node synced block 5 and 6 created by producer 1 (wrong producer) instead of producer 2 (correct producer) and skipped milestone 2, when milestone 3 arrives, it will never be able to sync further because the blocks won't match milestone 3. As more blocks are produced, the node will eventually reorg to the correct fork due to lesser total difficulty from its local fork.
This is a great question in the end and it will help for future reference. Thanks!

lucca30: Nice, thanks for the detailed explanation.

PR #1578, opened 2025-06-13
Fix/issue conversion 1577

astudnev:
Fixed the issue with using stateDb in the processor
to fix https://github.com/maticnetwork/bor/issues/1577


manav2401:
Hi, found a more elegant way to handle this than just passing state db around. Fixed in https://github.com/maticnetwork/bor/pull/1623. Closing this for now, thanks for your contribution.

PR #1580, opened 2025-06-17
Add historical data retention related flags

manav2401:
# Description

Geth has multiple `history` related flags which control data retention for transactions, logs, and state. While doing upstream merge, we didn't implement those flags in our cli and hence they were not configurable until now. This PR adds those flags. Moreover, to match the naming convention it adds a deprecation notice for `txlookuplimit` flag which will be replaced by `history.transactions` flag in next release. 

# Changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -143,6 +143,24 @@ type Config struct {
 
 	// Pprof has the pprof related settings
 	Pprof *PprofConfig `hcl:"pprof,block" toml:"pprof,block"`
+
+	// HistoryConfig has historical data retention related settings
+	History *HistoryConfig `hcl:"history,block" toml:"history,block"`
+}
+
+type HistoryConfig struct {
+	// TransactionHistory denotes the maximum number of blocks from head whose tx indices are reserved.
+	TransactionHistory uint64 `hcl:"transactions,block" toml:"transactions,block"`
+
+	// LogHistory denotes the maximum number of blocks from head where a log search index is maintained.
+	LogHistory uint64 `hcl:"logs,block" toml:"logs,block"`
+
+	// LogNoHistory denotes whether log search index is maintained or not.
+	LogNoHistory bool `hcl:"logs.disable,block" toml:"logs.disable,block"`
+
+	// StateHistory denotes number of recent blocks to retain state history for (only relevant
+	// in state.scheme=path)
+	StateHistory uint64 `hcl:"state,block" toml:"state,block"`
 }
 
 type LoggingConfig struct {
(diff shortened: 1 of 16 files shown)


internal/cli/server/config.go
```diff
@@ -1131,10 +1155,21 @@ func (c *Config) buildEth(stack *node.Node, accountManager *accounts.Manager) (*
...
 		n.TriesInMemory = c.Cache.TriesInMemory
 		n.FilterLogCacheSize = c.Cache.FilterLogCacheSize
+
+		log.Warn("The flag txlookuplimit will be deprecated soon, please use history.transactions instead")
+		n.TransactionHistory = c.Cache.TxLookupLimit
+	}
+
+	// History
+	{
+		// TODO: uncomment this when txlookuplimit is completely removed
```
pratikspatil024, line 1172: it is already uncommented, right? Am I missing something?

manav2401: I meant uncomment the part where we use the new flag i.e. `history.transactions` once we deprecate the old flag.

manav2401: right now because we set the value from old flag, we can't again set it by new flag as it would break things for operators (not break really but will lead to unexpected results). once we remove old flag, we will set it from the value obtained from new flag.

pratikspatil024: okay, got it. Thanks!

internal/cli/server/config.go
```diff
@@ -1131,10 +1155,21 @@ func (c *Config) buildEth(stack *node.Node, accountManager *accounts.Manager) (*
 		n.TrieDirtyCache = calcPerc(c.Cache.PercGc)
 		n.NoPrefetch = c.Cache.NoPrefetch
 		n.Preimages = c.Cache.Preimages
-		n.TransactionHistory = c.Cache.TxLookupLimit
 		n.TrieTimeout = c.Cache.TrieTimeout
 		n.TriesInMemory = c.Cache.TriesInMemory
 		n.FilterLogCacheSize = c.Cache.FilterLogCacheSize
+
+		log.Warn("The flag txlookuplimit will be deprecated soon, please use history.transactions instead")
```
ngotchac: Shouldn't this warning be only printed when the flag is set?

manav2401: I intentionally skipped it as requires an explicit check very early while parsing the arguments and config leading to unnecessary complexity. Although it's just one flag so I'll handle it.

ngotchac: All right. It would just be a bit misleading to see this log is all.

manav2401: Handled in https://github.com/maticnetwork/bor/pull/1580/commits/6a57d197f3c8c09479c538014a686043f23dad88. Thanks!

lucca30 (approved):

PR #1584, opened 2025-06-18
Add mutex to block_fetcher to prevent concurrent map access

cffls:
Also added more unit tests for block fetcher and witness manager.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/fetcher/block_fetcher.go
+++ b/eth/fetcher/block_fetcher.go
@@ -20,6 +20,7 @@ package fetcher
 import (
 	"errors"
 	"math/rand"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -207,6 +208,9 @@ type BlockFetcher struct {
 
 	done chan common.Hash
 	quit chan struct{}
+
+	// Protect concurrent map access from goroutines
+	mu sync.RWMutex
 
 	// Announce states
 	announces  map[string]int                   // Per peer blockAnnounce counts to prevent memory exhaustion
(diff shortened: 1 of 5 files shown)


lucca30:
That's actually a really good protection over the map accesses. I'm just not sure if we should do it more granularly by including one mutex for each map instead of a single one.

cffls:
> I'm just not sure if we should do it more granularly by including one mutex for each map instead of a single one.

Most usages of this mutex is for `queue`, so I don't think it is necessary to add a mutex for each map.

eth/fetcher/block_fetcher.go
```diff
@@ -1058,27 +1080,33 @@ func (f *BlockFetcher) enqueue(peer string, header *types.Header, block *types.B
 		return
 	}
 
+	f.mu.Lock()
```
lucca30, line 1083: This lock is just to protect the `count := f.queues[peer] + 1` or it also covers `blockBroadcastDOSMeter.Mark(1)` operations?

If just the first one, maybe it's worth replacing by a `RLock()` and unlocking right after the `count` assignement.

lucca30 (approved):
Looks good to me, just one question in the comments.

eth/fetcher/block_fetcher.go
cffls: Thanks! I’d like to keep using the locks because we don’t know whether the code in the middle will start writing to the map. Since it doesn’t affect the performance much, it is okay to keep it.

PR #1586, opened 2025-06-18
fix(downloader): prevent duplicate WriteTd calls in fast-forward sync

kamuikatsurgi:
# Description

This fix ensures we avoid redundant WriteTd calls by checking if the existing total difficulty matches the incoming milestone. It also logs mismatches and overwrites them to maintain an accurate state during stateless sync.

### DIFF
--- a/eth/downloader/bor_downloader.go
+++ b/eth/downloader/bor_downloader.go
@@ -2133,24 +2133,32 @@ func (d *Downloader) fetchWitnesses(from uint64, beaconMode bool) error {
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


lucca30 (commented):
I think the fix should be on something before it: why `UpdateFastForwardBlockFromMilestone` is being called multiple times and not just once? 

Because this proposed fix solves the write issue but introduce multiple reads on DB instead.

kamuikatsurgi:
@lucca30 This comes from a go routine that checks and receives the milestones from Heimdall, so I can't fix that. At least reads are less expensive and safer than spam writes to the DB.

cffls (approved):

PR #1597, opened 2025-06-27
consensus/bor, params: added OverrideStateSyncRecordsInRange to overwrite the state sync records for a range of blocks

pratikspatil024:
# Description

Added OverrideStateSyncRecordsInRange to overwrite the state sync records for a range of blocks.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -1473,10 +1473,19 @@ func (c *Bor) CommitStates(
 		return stateSyncs, nil
 	}
 
+	// This if statement checks if there are any state sync record overrides configured for the current block number.
+	// If there are, it truncates the eventRecords array to the specified number of records.
 	if c.config.OverrideStateSyncRecords != nil {
 		if val, ok := c.config.OverrideStateSyncRecords[strconv.FormatUint(number, 10)]; ok {
 			eventRecords = eventRecords[0:val]
 		}
+	}
+
+	// This if statement checks if there are any state sync record overrides configured for the current block number.
+	// If there are, it truncates the eventRecords array to the specified number of records.
+	overrideStateSyncRecord, ok := c.config.GetOverrideStateSyncRecord(number)
+	if ok {
+		eventRecords = eventRecords[0:overrideStateSyncRecord]
 	}
 
 	fetchTime := time.Since(fetchStart)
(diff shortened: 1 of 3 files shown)


builder/files/genesis-amoy.json
```diff
@@ -39,6 +39,9 @@
         },
         "validatorContract": "0x0000000000000000000000000000000000001000",
         "stateReceiverContract": "0x0000000000000000000000000000000000001001",
+        "overrideStateSyncRecordsInRange": [
+          { "startBlock": 24184305, "endBlock": 23220127, "value": 0 }
```
ssandeep: How can startBlock be > endBlock?

cffls: +1 Start block could be the last settled milestone before v0.2.5 upgrade `23,220,015`

pratikspatil024: Nice catch, my bad, thanks Sandeep.
I created a separate PR for this - https://github.com/maticnetwork/bor/pull/1598.
Will add unit tests in this PR

pratikspatil024: @cffls - So the current range is `23184305` to `23220127`
Which is `last state sync (5709) before the migration + 1 ` to `first state sync (5710) after the migration + 1`.

This milestone block's ([23220015](https://amoy.polygonscan.com/block/23220015)) timestamp is `Jun-26-2025 01:31:25 PM` and the timestamp of state sync `5710` is `Jun-25-2025 08:00:12`. Which means the state sync took place before the milestone. Now when syncing from scratch, this will cause a problem, because the milestone block will be imported much later, and before that heimdall will return the state sync, which will cause a problem.
WDYT?

cffls: Sorry about the confusion. Yes, you are totally right about this. I didn't think of the case where state sync events will get filled even before 23220015.

cffls (approved):

builder/files/genesis-amoy.json
pratikspatil024: np! Thanks!

PR #1602, opened 2025-06-30
chore: improve smoke tests

kamuikatsurgi:
# Description

- Improves smoke tests by doing 100 state sync deposits and checking the count in Heimdall and the balance in Bor.

### DIFF
--- a/.github/matic-cli-config.yml
+++ b/.github/matic-cli-config.yml
@@ -21,7 +21,7 @@ blockNumber:
   - '0'
 
 sprintSize:
-  - '64'
+  - '16'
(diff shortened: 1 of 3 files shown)


kamuikatsurgi:
This is failing because of the upload logs case, and not because of the changes I made. The smoke test passes successfully.

manav2401 (approved):
LGTM, might be worth extending this approach to send L1 to L2 bridge transactions from different (at least 2) accounts instead of sending all from a single account.

kamuikatsurgi:
Yeah, good point. I can make a follow-up PR to do this. Merging this for now.

PR #1603, opened 2025-06-30
Support geth's log history flags

MqllR:
Bor 2.1.0 use Geth [1.15.6](https://github.com/ethereum/go-ethereum/releases/tag/v1.15.6)'s new log index. But as of now, we cannot control the settings using flags.


manav2401:
Hi, this is already handled in https://github.com/maticnetwork/bor/pull/1580. It was blocked because of some failing tests but it's fixed now and should be merged today. Thanks!

PR #1604, opened 2025-06-30
Improve stateless sync

cffls:
# Description

Improve stateless sync

This commit improves a few things about stateless sync:
 - Fix 'unknown ancestor' when a node is running in stateless sync mode
 - Extract all veblop snapshot logic to a separate function to make the code simpler. This also resolved a backward compatibility issue with non-veblop-enabled node.
 - Add witness cache to witness manager to handle the case when witness arrives before its corresponding block
 - Improve witness fetch reschedule event
 - Fix witness header hash when it is produced by a block producer

## Testing
Tested in kurtosis

### DIFF
--- a/eth/fetcher/witness_manager.go
+++ b/eth/fetcher/witness_manager.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ethereum/go-ethereum/eth/protocols/wit"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
+	ttlcache "github.com/jellydator/ttlcache/v3"
 )
 
 // Constants for witness unavailability handling
@@ -36,6 +37,9 @@ const (
 	// witness. Once the penalty window elapses the peer is automatically
 	// eligible again.
 	witnessFailurePenalty = 5 * time.Second
+
+	witnessCacheSize = 10
+	witnessCacheTTL  = 2 * time.Minute
 )
 
 // witnessRequestState tracks the state of a pending witness request.
@@ -43,6 +47,13 @@ type witnessRequestState struct {
 	op       *blockOrHeaderInject // The original block/header injection operation.
 	announce *blockAnnounce       // Announcement details, non-nil if a fetch is in flight.
 	retries  int                  // Number of fetch attempts already made
+}
+
+// cachedWitness represents a witness that arrived before its corresponding block
+type cachedWitness struct {
+	witness   *stateless.Witness
+	peer      string
+	timestamp time.Time
 }
 
 // witnessManager handles the logic specific to fetching and managing witnesses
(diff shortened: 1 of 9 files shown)


eth/fetcher/witness_manager.go
```diff
@@ -57,15 +68,22 @@ type witnessManager struct {
...
 	// Witness-specific state
-	pending               map[common.Hash]*witnessRequestState // Blocks waiting for witness or actively fetching.
-	failedWitnessAttempts map[string]int                       // Tracks witness fetch failures per peer.
-	peerPenalty           map[string]time.Time                 // Tracks peers currently penalised until a given time.
-	witnessUnavailable    map[common.Hash]time.Time            // Tracks hashes whose witnesses are known to be unavailable, with expiry times.
+	pending               map[common.Hash]*witnessRequestState         // Blocks waiting for witness or actively fetching.
+	failedWitnessAttempts map[string]int                               // Tracks witness fetch failures per peer.
+	peerPenalty           map[string]time.Time                         // Tracks peers currently penalised until a given time.
+	witnessUnavailable    map[common.Hash]time.Time                    // Tracks hashes whose witnesses are known to be unavailable, with expiry times.
+	witnessCache          *ttlcache.Cache[common.Hash, *cachedWitness] // TTL cache of witnesses that arrived before their blocks
```
lucca30, line 75: Nice catch with this caching idea. Tbh I'm just concerned with the ubiquity of the future Witness Caching feature we are planning to do. 

We can keep this one as the "Witness Cache", but we should rename the other one.

My suggestion will be something like: `DeltaWitnessSync`, `WitnessPatch` or something similar.

miner/worker.go
```diff
@@ -837,6 +837,10 @@ func (w *worker) resultLoop() {
 			log.Info("Successfully sealed new block", "number", block.Number(), "sealhash", sealhash, "hash", hash,
 				"elapsed", common.PrettyDuration(time.Since(task.createdAt)))
 
+			if witness != nil {
+				witness.SetHeader(block.Header())
```
lucca30, line 841: Were witness missing it's header when the block was mined instead of imported? Hum, makes sense because the header it's just done after sealing right?

Good catch!

lucca30 (approved):
LGTM.

I’ve added a few comments to help spark the discussion! 😊

eth/fetcher/witness_manager.go
cffls: Yeah, agreed it might have a name conflict with the real witness caching in the future. `DeltaWitness` sounds succinct and accurate to me. Another question is where do we want to implement and apply the witness delta/patch, such as in the witness protocol itself, or in blockchain.go. Keeping it low level in witness protocol will require least amount of changes to existing code, but it might not be as efficient. While implementing it in blockchain level might be most efficient, but will require changes in many components.

miner/worker.go
cffls: Caught this bug when inspect the payload of witnesses in debug log :D  You are right, the reason why the header would change was because the signing. The original header in the witness was created in `prepare()` function without the signature of block producer, but the final block header introduces the signature and therefore the block hash would be different.

PR #1617, opened 2025-07-04
core/txpool: decouple price heap for reduced miner wait time

manav2401:
# Description

The transaction pool in bor (derived from geth) follows a very **strict consistency** model as there's only a single lock which is being used for all internal operations. This means the reader of the pool will get the most updated results but this takes a hit in performance. This doesn't prioritise the miner which needs the next set of transactions to process immediately. The waiting time when the pool is filled with ~100k transactions can go >1s leaving very less time for execution. 

This PR does multiple changes to decouple the internal price heap (which is a costly operation when there are more elements) from the run reorg operation. This frees up the lock faster allowing miner to pull transactions without waiting much. It still aims to achieve **eventual consistency** by updating everything in background. 

Some more technical details on the changes
- Move operations needing lock in run reorg function together, release the lock as early as possible, and then call the reheap (which is costly). We're okay if this takes more time now as worker can immediately call the pending function and get whatever needed for building the next block. 
- Protect the functions of priced heap by it's internal lock (which is also used for reheap)
- Convert some functions in add transactions to run asynchronously i.e. run them in a go routine unblocking the parent function. This mostly involves all calls to the heap (adding / removing). 
- There are some blocking heap calls in add transactions which needs the lock. Unblock the global lock while heap operations are going on. Once the heap operations are done, we can acquire the lock again. This is safe to do as this will delay the addition of a new transaction in the worst case. On the other hand, it reduces the wait time of miner. 

On implementing these changes, pretty good performance improvements can be seen in the waiting time of worker to get pending transactions on a shadow forked mainnet node. While the experiment was conducted only a couple of times, the pending time has reduced to <200-300ms which is way better than ~1s earlier. This means we have way more time for execution and are not blocked by internal pool operations. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


core/txpool/legacypool/legacypool.go
```diff
@@ -823,6 +823,174 @@ func (pool *LegacyPool) add(tx *types.Transaction) (replaced bool, err error) {
...
 
+// addAsync is a semi-async version of `add`. It frees up the global pool lock
+// to allow other functions to use the pending pool. It also performs some
+// operations on `pricedList` async to avoid waiting for reheap.
+//
+// It also validates a transaction and inserts it into the non-executable queue
+// for later pending promotion and execution. If the transaction is a replacement
+// for an already pending or queued one, it overwrites the previous transaction
+// if its price is higher.
+func (pool *LegacyPool) addAsync(tx *types.Transaction) (replaced bool, err error) {
```
manav2401: While this is a new function, most of the contents (except the lock changes) are same as `add`. While the diff doesn't detect it, I'd recommend reviewing this side by side with `add` function.

core/txpool/legacypool/legacypool.go
```diff
@@ -1693,13 +1880,19 @@ func (pool *LegacyPool) demoteUnexecutables() {
...
+			// Delete the entire pending entry if it became empty.
 			delete(pool.pending, addr)
 			if _, ok := pool.queue[addr]; !ok {
 				pool.reserver.Release(addr)
 			}
+		} else {
+			// Update the latest known pending nonce
+			highestPending := list.LastElement()
+			nonces[addr] = highestPending.Nonce() + 1
+			pool.pendingNonces.setAll(nonces)
```
manav2401, line 1768: This change was done earlier in outer function but I have moved it here to simple save extra iteration on the whole pending pool.

core/txpool/legacypool/legacypool.go
cffls: Since they are mostly the same, would it be possible to combine `addAsync` and `add` by adding a new param to `add`, similar to what was done to `addTxs` below? This would reduce future maintenance when new changes are added to `add`, because then we wouldn't need to update `addAsync` separately.

manav2401: Okay makes sense, let me give it a try.

manav2401: Done in https://github.com/maticnetwork/bor/pull/1617/commits/5b84ffe57f5d43d4e2aac4f53fe5d876a16088f5

cffls:
Looks good! Regarding testing, would it possible to add some benchmark tests for add async?

manav2401:
> Looks good! Regarding testing, would it possible to add some benchmark tests for add async?

I tried to find ways to add a new test / benchmark in transaction pool but most of them are about testing individual functions and not a live chain (where we receive new blocks). This limits our testing as the main culprit was `reheap` which is only triggered if a new block is received. None of the existing tests are for that scenario. 

Hence, instead I tried to write a simple test in worker where I can mock a live chain, and interact with transaction pool. The results are promising for this branch. For a short lived chain, below are the time taken for pending call to return 20k transactions. 

This branch:
```
[11.302167ms 11.285166ms 10.580084ms 17.394709ms 7.569958ms 14.791375ms]
```

Develop:
```
[57.160291ms 72.235958ms 76.678625ms 81.324375ms 79.223416ms 60.344125ms]
```

Clearly, this branch performs better. Note that the overhead for other operations is very minimal as all the transactions are from same account and everything operates in a fresh in-memory db. 

Although, this is just a manual test/check. I am not sure how to make this an actual *test* which checks and validates something on every run as there's nothing to compare the results against.

miner/worker_test.go
```diff
@@ -899,6 +899,82 @@ func TestCommitInterruptPending(t *testing.T) {
 	w.stop()
 }
 
+// TestBenchmarkPending is a simple benchmark test to measure the performance of transaction pool. It inserts
+// large number of transactions into the pool and captures the time taken for `pending` to return the list
+// of pending transactions. The purpose is just to compare the performance on different branches.
+func TestBenchmarkPending(t *testing.T) {
```
cffls, line 905: Why not directly use benchmark? `func BenchmarkPending(b *testing.B)`

Then the test could be benchmarked by:

```go test -bench=.```

Comment (manav2401) [reply miner/worker_test.go:905]: There were a few reasons why I didn't go that route. 
1. It will just be a standalone benchmark and there's nothing to compare against. 
2. The whole tests needs a mock chain to be running which will be tough to achieve in a benchmark as it runs the same set of operation/code thousands of time which doesn't seem feasible as a single run takes around 10s. This test is more of en e2e test than standalone unit test. 

I tried to write a test in the legacy pool package itself so that I don't have to create a mock chain but it seems hacky as there's no way to trigger a reheap due to new head in those tests. Without the reheap, the performance improvements are negligible. None of the existing tests try to trigger a reheap which I can take for reference. Although, let me try checking this again. Thanks!

**Review by kamuikatsurgi — APPROVED** [review]

File: core/txpool/legacypool/list.go [hunk 4]
```diff
@@ -638,10 +638,24 @@ func newPricedList(all *lookup) *pricedList {
 
 // Put inserts a new transaction into the heap.
 func (l *pricedList) Put(tx *types.Transaction) {
+	l.reheapMu.Lock()
+	defer l.reheapMu.Unlock()
+
 	// Insert every new transaction to the urgent heap first; Discard will balance the heaps
 	heap.Push(&l.urgent, tx)
 }
```
lucca30, line 646: I think you should consider removing this function and unifying the whole `Put` calls on `PutMany`. In this way we avoid duplication of the lock pattern.

core/txpool/legacypool/list.go
```diff
@@ -638,10 +638,24 @@ func newPricedList(all *lookup) *pricedList {
...
+func (l *pricedList) PutMany(txs types.Transactions) {
+	l.reheapMu.Lock()
+	defer l.reheapMu.Unlock()
+
+	for _, tx := range txs {
+		// Insert every new transaction to the urgent heap first; Discard will balance the heaps
+		heap.Push(&l.urgent, tx)
+	}
+}
+
```
lucca30, line 658: There is a potential gain of performance if we add all elements to a single list and then init the heap instead of incrementally pushing.

- Single push: O(log N)
- Multiple pushes: O(N logN)
- Init: O(N)


So it will depend a bit on how many txs are we going to potentially add here to validate which approach is better.

core/txpool/legacypool/list.go
```diff
@@ -658,6 +672,9 @@ func (l *pricedList) Removed(count int) {
 // Underpriced checks whether a transaction is cheaper than (or as cheap as) the
 // lowest priced (remote) transaction currently being tracked.
 func (l *pricedList) Underpriced(tx *types.Transaction) bool {
+	l.reheapMu.Lock()
+	defer l.reheapMu.Unlock()
+
```
lucca30, line 677: What about replacing this for an `RLock` and just properly doing a `Lock` when `heap.Pop(h)` gets called inside of `underpricedFor`?

core/txpool/legacypool/list.go
manav2401: I wanted to keep the diff with upstream as low as possible and hence created a new `PutMany` as it was just called at one place while `Put` is called at few other places.

lucca30 (approved):
lgtm

core/txpool/legacypool/list.go
manav2401: As discussed, I think there can still be some scenarios which may cause a race condition. 

Firstly, we'll need to free the `RLock` before using `Lock` before popping from heap. If we release read lock, acquire the write lock, release it, and try to re-acquire the read lock again, it's possible that another process waiting to write gets preference over the last re-acquire of read lock. This may lead to some race conditions.

core/txpool/legacypool/list.go
manav2401: `PutMany` is called in cases where we're re-adding dropped transactions. The count IMO is very very low so it should be fine to just use `Push` instead of `Init`.

vicmik:
@manav2401: I updated to 2.2.10 and after about 7 minutes of runtime a crash which I think is related to the changes in this PR. I can confirm that I keep getting this crash after running the node for a couple of minutes. 

Here is the log and callstack:

Sep 09 14:26:50 xxx bash[1139962]: INFO [09-09|14:26:50.347] Imported new chain segment               number=76,245,902 hash=e62212..d4cd7a blocks=1  txs=66   mgas=12.666  elapsed=55.455ms  mgasps=228.401 snapdiffs=7.76MiB triediffs=220.65MiB triedirty=241.36MiB
**Sep 09 14:26:50 xxx bash[1139962]: panic: runtime error: makeslice: cap out of range**
Sep 09 14:26:50 xxx bash[1139962]: goroutine 15521 [running]:
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/core/txpool/legacypool.(*pricedList).Discard(0xc0fdaee300, 0xffffffffffffff22)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/core/txpool/legacypool/list.go:716 +0xc5
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/core/txpool/legacypool.(*LegacyPool).add(0xc0fdafc1e0, 0xc138d00aa0, 0x1)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/core/txpool/legacypool/legacypool.go:802 +0x5c6
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/core/txpool/legacypool.(*LegacyPool).addTxs(0xc0fdafc1e0, {0xc11efcfbe8, 0x1, 0xaa659bfc7b4e9f6b?}, 0x1)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/core/txpool/legacypool/legacypool.go:1086 +0xfc
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/core/txpool/legacypool.(*LegacyPool).Add(0xc0fdafc1e0, {0xc11efcfbe0, 0x1, 0x4087480?}, 0x0)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/core/txpool/legacypool/legacypool.go:1062 +0x45c
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/core/txpool.(*TxPool).Add(0xc0f4dceab0, {0xc030cbc080, 0x1, 0x3c1c820?}, 0x0)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/core/txpool/txpool.go:361 +0x27f
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth.newHandler.func5({0xc030cbc080?, 0x409555?, 0xc022a10d40?})
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/handler.go:259 +0x34
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth/fetcher.(*TxFetcher).Enqueue(0xc0143188f0, {0xc073a55b80, 0x40}, {0xc030cbc080, 0x1, 0x4}, 0x0)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/fetcher/tx_fetcher.go:342 +0x214
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth.(*ethHandler).Handle(0xc002002200, 0xc073a76d00, {0x5248218?, 0xc06d332c30})
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/handler_eth.go:83 +0x527
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth/protocols/eth.handleTransactions({0x527dee0, 0xc002002200}, {0x524b688, 0xc014383c80}, 0xc073a76d00)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/protocols/eth/handlers.go:534 +0x2ed
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth/protocols/eth.handleMessage({0x527dee0, 0xc002002200}, 0xc073a76d00)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/protocols/eth/handler.go:232 +0x53e
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth/protocols/eth.Handle({0x527dee0, 0xc002002200}, 0xc073a76d00)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/protocols/eth/handler.go:159 +0x37
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth/protocols/eth.MakeProtocols.func1.1(0xc11a457d88?)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/protocols/eth/handler.go:114 +0x25
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth.(*handler).runEthPeer(0xc002002200, 0xc073a76d00, 0xc073da1b60)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/handler.go:426 +0xf08
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth.(*ethHandler).RunPeer(0x44?, 0xc120ea8150?, 0x52489b8?)
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/handler_eth.go:41 +0x13
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/eth/protocols/eth.MakeProtocols.func1(0xc120ea8150, {0x52489b8, 0xc06fb73720})
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/eth/protocols/eth/handler.go:113 +0x102
Sep 09 14:26:50 xxx bash[1139962]: github.com/ethereum/go-ethereum/p2p.(*Peer).startProtocols.func1()
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/p2p/peer.go:476 +0x75
Sep 09 14:26:50 xxx bash[1139962]: created by github.com/ethereum/go-ethereum/p2p.(*Peer).startProtocols in goroutine 15518
Sep 09 14:26:50 xxx bash[1139962]:         /home/matic/bor/p2p/peer.go:474 +0xac
Sep 09 14:26:50 xxx systemd[1]: bor.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Sep 09 14:26:50 xxx systemd[1]: bor.service: Failed with result 'exit-code'.

PR #1620, opened 2025-07-08
Witness Pagination

lucca30:
# Description

This PR aims to solve an issue regarding limits of the P2P: **it only supports packets up to 16MB**. With witness pagination we deliver a feature that allows breaking the RLP bytes of a big witness in small pages. We designed this first version in a way all pages are requested from the same peer. But its open to future changes where we can distribute these requests across the network.
 
# Changes

# Breaking changes

It changes the p2p protocol, so it must be updated across all nodes. (I didnt create a new protocol version since stateless still in development phase).


eth/handler_eth.go
```diff
@@ -121,18 +120,18 @@ func (h *ethHandler) handleBlockAnnounces(peer *eth.Peer, hashes []common.Hash,
 		}
 	}
 
-	var witnessRequester func(hash common.Hash, sink chan *wit.Response) (*wit.Request, error)
+	var witnessRequester func(hash common.Hash, sink chan *eth.Response) (*eth.Request, error)
```
cffls, line 123: Why is this change required?

eth/handler_wit.go
```diff
@@ -105,21 +99,61 @@ func (h *witHandler) handleWitnessHashesAnnounce(peer *wit.Peer, hashes []common
...
-	for _, hash := range req.Hashes {
-		// Call the blockchain method which now returns raw RLP bytes
-		witnessBytes := rawdb.ReadWitness(h.Chain().DB(), hash)
-		if len(witnessBytes) > 0 {
-			log.Trace("Witness found in DB immediately", "hash", hash)
-			witnessesRLPBytes = append(witnessesRLPBytes, witnessBytes)
+func (h *witHandler) handleGetWitness(peer *wit.Peer, req *wit.GetWitnessPacket) (wit.WitnessPacketResponse, error) {
+	const (
+		PageSize = 15 * 1024 * 1024 // 15 MB
+	)
```
cffls: Let's move constants to the top of the file.

eth/handler_wit.go
```diff
@@ -105,21 +99,61 @@ func (h *witHandler) handleWitnessHashesAnnounce(peer *wit.Peer, hashes []common
...
+			}
+
+			start := PageSize * witnessPage.Page
+			end := start + PageSize
+			if end > uint64(len(witnessBytes)) {
+				end = uint64(len(witnessBytes))
+			}
+			witnessPageResponse.Data = witnessBytes[start:end]
+		}
+		response = append(response, witnessPageResponse)
```
cffls, line 159: Maybe I am missing something obvious. If the response of all pages are bundled here, wouldn't it exceed the 16MB limit of rlpx?

eth/handler_eth.go
lucca30: It was all about standardizing the way we request a witness in the protocol. Previously you can see that this `witnessRequester` was requesting using `ethPeer.witPeer.RequestWitness([]common.Hash{hash}, sink)` while other parts of the system was calling `ethPeer.RequestWitnesses([]common.Hash{hash}, sink)`.

Since all the changes was applied on `RequestWitnesses()`, I do also applied this function as the default call in all parts of the system.

eth/handler_wit.go
lucca30: done

eth/handler_wit.go
lucca30: Yes, you are totally right. 

I just wrote all the code supporting an array of pages in the request, even if all the requests **always** requests a single page. Just did that way so in a near future we can think of requesting multiple small witnesses in a single request. (For example, blocks with 2~4MB of witness).

eth/handler_eth.go
cffls: Okay makes sense. Thanks for the explanation.

eth/downloader/peer.go
```diff
@@ -149,7 +149,7 @@ func (p *peerConnection) ReceiptCapacity(targetRTT time.Duration) int {
 func (p *peerConnection) WitnessCapacity(targetRTT time.Duration) int {
 	cap := p.rates.Capacity(wit.MsgWitness, targetRTT)
 	// Placeholder limit, replace MaxWitnessFetch with actual constant if defined
-	const MaxWitnessFetch = 128 // Placeholder value
+	const MaxWitnessFetch = 1 // Placeholder value
```
cffls: Would be great if this could be moved to the same place where other max fetch, e.g. MaxHeaderFetch, is defined. Could you help me understand why is the max witness fetch limit only 1? Maybe it could be a number linear to max block fetch? For example, `MaxWitnessFetch = MaxBlockFetch / 10`.

eth/protocols/wit/handlers.go
```diff
@@ -40,7 +40,14 @@ func handleWitness(backend Backend, msg Decoder, peer *Peer) error {
...
 	log.Debug("Dispatching witness response packet", "peer", peer.ID(), "reqID", packet.RequestId, "count", len(packet.WitnessPacketResponse))
-	return peer.dispatchResponse(res, nil)
+	log.Info("[witpagination] starting dispatch response")
+	err := peer.dispatchResponse(res, nil)
+	if err != nil {
+		log.Info("[witpagination] error dispatching response", "err", err)
+	} else {
+		log.Info("[witpagination] successfully dispatched response")
+	}
+	return err
```
cffls: Maybe debug level is better?

eth/handler_wit.go
cffls: Nice, makes sense to me.

eth/peer.go
```diff
@@ -130,40 +153,61 @@ func (p *ethPeer) RequestWitnesses(hashes []common.Hash, dlResCh chan *eth.Respo
...
-				// If cancelled, exit the goroutine. The closing of witResCh
-				// will also eventually terminate the loop, but returning
-				// here ensures we don't block trying to send after cancellation.
+			err := p.receiveWitnessPage(witRes, receivedWitPages, reconstructedWitness, hashes, witReqs, &witReqsWg, witTotalPages, witTotalRequest, witResCh)
+			if err != nil {
 				return
 			}
+
+			witRes.Done <- nil
+			witReqsWg.Done()
```
cffls: One thing it could be improved is that witnesses that have more than one page would need to wait until all the responses are received before requesting the remaining pages, which could potentially slow down the download. A faster way is to make all intially requests and their handling of the remaining pages async.

cffls (commented):
Looks pretty good overall. Would be great if there are unit tests for new functions & features.

eth/peer.go
```diff
@@ -172,3 +216,83 @@ func (p *ethPeer) RequestWitnesses(hashes []common.Hash, dlResCh chan *eth.Respo
...
+			wit, err := p.reconstructWitness(receivedWitPages[page.Hash])
+			if err != nil {
+				return err
+			}
+			reconstructedWitness[page.Hash] = wit
+		}
+
+		// check and build any remaining witnessRequest for the witnesses we dont know previously the totalPages
+		witTotalPages[page.Hash] = page.TotalPages
+		p.buildWitnessRequests(hashes, witReqs, witReqsWg, witTotalPages, witTotalRequest, witResCh)
```
cffls: One adversarial scenario we might need to think about is when a malicious node returns an extremely large total page number, e.g. 100k. It might make the node create lots of requests, use lots of resources, and hang forever. Would be nice to put a max concurrent request limit. Also a timeout from the caller will be needed (I think it is already there but would be nice to double check).

kamuikatsurgi:
Just wondering here, because we want to add verification for witnesses after we receive them from peers, we would be able to do that after the pagination as well? Cause we can't verify chunks of witnesses. We'd need the full witness.

cffls:
> Just wondering here, because we want to add verification for witnesses after we receive them from peers, we would be able to do that after the pagination as well? Cause we can't verify chunks of witnesses. We'd need the full witness.

Yes, the validation will happen after all pages are collected.

eth/downloader/peer.go
lucca30: Yeah. Agreed on moving the variable. 

Regarding the limit, I created a task where we can better specify it but for now it's limited for 1 witness per request since we just know the size of the witnesses after requesting it. So in this future task we can look on requesting a batch of witnesses that fits in a single request

eth/protocols/wit/handlers.go
lucca30: It was my bad, these are logs I used while developing. Thanks for noticing

eth/peer.go
lucca30: Actually that's what we do. 
While handling the responses we already check and request the remaining pages.
Here's the reference on the handling method:
https://github.com/maticnetwork/bor/blob/ffbf495ed7fdd458d3976eb886f6b131788ca1e1/eth/peer.go#L252-L254


This happens inside the response handling loop:
https://github.com/maticnetwork/bor/blob/ffbf495ed7fdd458d3976eb886f6b131788ca1e1/eth/peer.go#L236

So it builds as soon as it knows there's more pages to request.

eth/peer.go
lucca30: That's a true and nice concern. Implementing here, thanks a lot. A nice and smart catch, as you always do.

eth/peer.go
```diff
@@ -108,17 +143,28 @@ func (p *ethPeer) RequestWitnesses(hashes []common.Hash, dlResCh chan *eth.Respo
...
-	// Make the actual request via the witness protocol peer.
-	witReq, err := p.witPeer.Peer.RequestWitness(hashes, witResCh)
-	if err != nil {
-		p.witPeer.Log().Error("RequestWitnesses failed to make wit request", "peer", p.ID(), "err", err)
-		return nil, err
-	}
-	p.witPeer.Log().Trace("RequestWitnesses made wit request", "peer", p.ID(), "hashes", len(hashes))
+	p.witPeer.Peer.Log().Trace("RequestWitnesses called", "peer", p.ID(), "hashes", len(hashes))
+
+	witResCh := make(chan *wit.Response, 10)
```
cffls: Please move constant to the top of the file.

eth/downloader/queue.go
```diff
@@ -1056,46 +1054,27 @@ func (q *queue) DeliverWitnesses(id string, witnessData interface{}, meta interf
...
-		// q.expire(id, q.witnessPendPool, q.witnessTaskQueue)
-		return 0, fmt.Errorf("invalid witness data type: %T", witnessData)
-	}
 
-	log.Debug("DeliverWitnesses: Received witness RLP data", "peer", id, "count", len(witnessRLP))
+	log.Debug("DeliverWitnesses: Received witnesses", "peer", id, "count", len(witnesses))
 
 	// Mark incoming data and time
-	witnessInMeter.Mark(int64(len(witnessRLP)))
+	witnessInMeter.Mark(int64(len(witnesses)))
```
cffls, line 1061: Not sure if this is the right place, but it would be great to add some additional metrics:
 * histogram of the total pages per witness
 * histogram of the time to fetch a witness
 * histogram of witness size (in byte)

eth/peer.go
```diff
@@ -18,11 +18,22 @@ package eth
...
 	"github.com/ethereum/go-ethereum/eth/protocols/eth"
 	"github.com/ethereum/go-ethereum/eth/protocols/snap"
 	"github.com/ethereum/go-ethereum/eth/protocols/wit"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+const (
+	DefaultPagesRequestPerWitness   = 1
+	DefaultConcurrentRequestsPerPee = 5
```
cffls: `DefaultConcurrentRequestsPerPee` -> `DefaultConcurrentRequestsPerPeer`

eth/peer.go
```diff
@@ -68,24 +79,42 @@ type witPeerInfo struct {
 	Version uint `json:"version"` // Witness protocol version negotiated
 }
 
-// witPeer is wrapper around wit.Peer to maintain a few extra metadata.
+type WitnessPeer interface {
```
cffls, line 84: Nice!

core/rawdb/accessors_state.go
```diff
@@ -280,18 +280,36 @@ func ReadWitness(db ethdb.KeyValueReader, blockHash common.Hash) []byte {
...
 
+func ReadWitnessSize(db ethdb.KeyValueReader, blockHash common.Hash) *uint64 {
+	log.Debug("ReadWitnessSize", "blockHash", blockHash)
+	data, err := db.Get(witnessSizeKey(blockHash))
+	if err != nil || len(data) == 0 {
+		return nil
+	}
+
+	number := binary.BigEndian.Uint64(data)
+	return &number
```
cffls, line 291: Why return a pointer? Seems safer to return a copy of the value.

eth/handler_wit.go
```diff
@@ -105,21 +100,57 @@ func (h *witHandler) handleWitnessHashesAnnounce(peer *wit.Peer, hashes []common
...
+			witnessSize[witnessBlockHash] = 0
+		} else {
+			witnessSize[witnessBlockHash] = *size
 		}
 	}
+
+	// query witnesses by demand
+	var response wit.WitnessPacketResponse
+	witnessCache := make(map[common.Hash][]byte, len(seen))
+	for _, witnessPage := range req.WitnessPages {
```
cffls, line 132: In this function, should there be a check to make sure the total size of the response to return won't exceed the limit of 16MB?

eth/peer.go
```diff
@@ -130,45 +176,166 @@ func (p *ethPeer) RequestWitnesses(hashes []common.Hash, dlResCh chan *eth.Respo
...
+) error {
+	witPacketPtr, ok := witRes.Res.(*wit.WitnessPacketRLPPacket)
+	if !ok {
+		p.witPeer.Peer.Log().Warn("RequestWitnesses received unexpected response type", "type", fmt.Sprintf("%T", witRes.Res), "peer", p.ID())
+	}
+
+	for _, page := range witPacketPtr.WitnessPacketResponse {
+
+		p.witPeer.Peer.Log().Trace("RequestWitnesses adapter received wit page response", "peer", p.ID(), "hash", page.Hash, "page", page.Page, "TotalPages", page.TotalPages, "lenData", len(page.Data))
+		if len(page.Data) == 0 {
```
cffls, line 308: Does it mean we've hit beyond the last page?

eth/peer.go
```diff
@@ -130,45 +176,166 @@ func (p *ethPeer) RequestWitnesses(hashes []common.Hash, dlResCh chan *eth.Respo
...
+			p.witPeer.Peer.Log().Debug("RequestWitnesses building a wit request", "peer", p.ID(), "hash", hash, "page", page)
+			witReqSem <- 1
+			witReq, err := p.witPeer.Peer.RequestWitness([]wit.WitnessPageRequest{{Hash: hash, Page: uint64(page)}}, witResCh)
+			if err != nil {
+				p.witPeer.Peer.Log().Error("RequestWitnesses failed to make wit request", "peer", p.ID(), "err", err)
+				return err
+			}
+			witReqsWg.Add(1)
+			witReqs = append(witReqs, witReq)
+			witTotalRequest[hash]++
```
cffls: If, for some reason, a page in the middle failed to deliver, are we able to detect it and resend the request?

eth/peer_test.go
```diff
@@ -0,0 +1,166 @@
...
+				}
+
+				muConcurrentCount.Lock()
+				concurrentCount--
+				muConcurrentCount.Unlock()
+			}()
+
+			return &wit.Request{}, nil
+		}).
+		Times(54)
```
cffls: Can move the calculation of `totalPages` to outside and use it here `Times(totalPages)`.

eth/protocols/wit/handlers.go
```diff
@@ -40,7 +40,8 @@ func handleWitness(backend Backend, msg Decoder, peer *Peer) error {
 
 	// Forward the response to the dispatcher
 	log.Debug("Dispatching witness response packet", "peer", peer.ID(), "reqID", packet.RequestId, "count", len(packet.WitnessPacketResponse))
-	return peer.dispatchResponse(res, nil)
+	err := peer.dispatchResponse(res, nil)
```
cffls: `return peer.dispatchResponse(res, nil)`

core/rawdb/accessors_state.go
lucca30: It was to allow `nil` returns which can easily tells us if there is a value or not.

eth/handler_wit.go
lucca30: I don't think so, since the protocol itself will block any invalid payload. So checking it seems as a double check that just do a "fast" fail validation (not so fast, because is in the end), imo.

eth/peer.go
lucca30: Yes, and empty `Data`  means beyond last page. Could also be checked trough `page.Page >= page.TotalPages`

eth/peer.go
lucca30: Hummm, not now. It will fail everything. Nice catch, I'll include this one. Thanks

eth/peer_test.go
lucca30: Agree, sounds better.

eth/protocols/wit/handlers.go
lucca30: Sure, it was probably while log debugging the code. Mb

eth/peer.go
lucca30: Agreed

eth/downloader/queue.go
lucca30: Yes, sure.

eth/handler_wit.go
cffls: Okay makes sense. In the current logic, is it possible that the client will make a request whose response would exceeds the limit? Do we have a protection mechanism from the client side?

eth/peer.go
lucca30: Done

eth/peer.go
lucca30: Thanks. Done

eth/handler_wit.go
lucca30: Yes, it's possible request something that exceeds since we allows multiple pages at once (even if now we just request one).

The protection mechanism is the p2p protocol itself.

eth/peer_test.go
lucca30: done

eth/protocols/wit/handlers.go
lucca30: done

eth/downloader/queue.go
lucca30: Actually these metrics are default from the queue. And it seems we are measuring in the way it should be

eth/handler_wit.go
lucca30: What I can do is protecting the server from building a super big response and consumes a lot of RAM while doing this. This is potentially an attack vector scenario.

lucca30: Done.

eth/peer.go
lucca30: Done

cffls (approved):
LGTM!

PR #1621, opened 2025-07-08
Stateless sync bytecode cache

cffls:
# Description

This PR mainly adds the ability to download all historical bytecode before starting stateless sync.

Also included a few improvements and bug fixes for stateless sync. 

# Changes

## Testing

- [x] I have added unit tests

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2274,16 +2274,15 @@ func (bc *BlockChain) InsertChainStateless(chain types.Blocks, witnesses []*stat
 		}
 
 		// Process the block using stateless execution
-		err := bc.ProcessBlockWithWitnesses(block, witness)
+		statedb, err := bc.ProcessBlockWithWitnesses(block, witness)
 		if err != nil {
 			return processed, err
 		}
 
-		statedb := state.StateDB{}
 		statedb.SetWitness(witness)
 
 		// Write the block to the chain without committing state
-		if _, err := bc.writeBlockAndSetHead(block, nil, nil, &statedb, false, true); err != nil {
+		if _, err := bc.writeBlockAndSetHead(block, nil, nil, statedb, false, true); err != nil {
 			return processed, err
 		}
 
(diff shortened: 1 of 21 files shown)


core/blockchain.go
```diff
@@ -997,7 +997,11 @@ func (bc *BlockChain) rewindHashHead(head *types.Header, root common.Hash) (*typ
 			logged = time.Now()
 			logger = log.Info
 		}
-		logger("Block state missing, rewinding further", "number", head.Number, "hash", head.Hash(), "elapsed", common.PrettyDuration(time.Since(start)))
+		if bc.cacheConfig.TriesInMemory == 0 {
```
lucca30: `bc.cacheConfig.TriesInMemory` is it the only way to check stateless mode here? Since we have some flags to enable it, it would be better using then here if possible.

core/blockchain.go
```diff
@@ -1037,6 +1046,13 @@ func (bc *BlockChain) rewindHashHead(head *types.Header, root common.Hash) (*typ
 		log.Debug("Skipping block with threshold state", "number", head.Number, "hash", head.Hash(), "root", head.Root)
 		head = bc.GetHeader(head.ParentHash, head.Number.Uint64()-1) // Keep rewinding
 	}
+	// If we exited the loop (e.g., for stateless nodes), return the current head
+	if bc.cacheConfig.TriesInMemory == 0 {
+		log.Info("Rewound to block (stateless mode)", "number", head.Number, "hash", head.Hash())
+	} else {
+		log.Info("Rewound to target block", "number", head.Number, "hash", head.Hash())
+	}
```
lucca30: Is it missing the return of the current head when the statement is true?

core/blockchain.go
```diff
@@ -1011,6 +1015,11 @@ func (bc *BlockChain) rewindHashHead(head *types.Header, root common.Hash) (*typ
 		}
 		// If the associated state is not reachable, continue searching
 		// backwards until an available state is found.
+		// Skip state checking for stateless nodes (TriesInMemory == 0)
+		if bc.cacheConfig.TriesInMemory == 0 {
+			// For stateless nodes, we can use any block as head without state checking
+			break
+		}
```
lucca30: I totally agree on skipping state check. But since we are now using bytecodes stored, we need somehow to verify if contracts stored matches. Maybe trough a new "codeHash" in the block.

core/blockchain.go
cffls: The sync mode flag isn't passed to blockchain, but I will try to add a new flag.

core/blockchain.go
cffls: Nice catch! Will add.

core/blockchain.go
cffls: It should be okay to skip because code are stored by code hash key. If the witness is correct, it will be able  to find the contract code from top down.

core/blockchain.go
cffls: Decided to revert this function and use a simpler solution - not calling this function when running in stateless mode.

lucca30 (approved):
lgtm, thanks

PR #1623, opened 2025-07-09
core, consensus: handle statedb interface conversion for hooked state db

manav2401:
# Description

Fixes https://github.com/maticnetwork/bor/issues/1577

Earlier in bor consensus, we just used to extract the underlying db by type conversion (as it's just wrapped by the `vm.StateDB` interface) like this:
```
state := wrappedState.(*state.StateDB)
```

This may panic (as the issue suggests) when the underlying db is not simple `*state.StateDB` but a hooked state db if tracing is enabled. It's a wrapper over `*state.StateDB` but handles tracing hooks for abstraction. 

This PR adds 2 interface methods
- `SetBalance`: For setting balance which is used by bor consensus when there's a hard fork (the new interface doesn't support it so need to add it)
- `Inner`: Returns the inner state db instance (for hooked as well as normal state db)

We could have handled the interface conversion for hooked state db as well to solve this simply but the `SetBalance` call wouldn't be tracked if tracing is enabled. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -886,9 +886,9 @@ func (c *Bor) Finalize(chain consensus.ChainHeaderReader, header *types.Header, 
 		return
 	}
 
-	// Extract the underlying state to access methods like `IntermediateRoot` and `Copy`
-	// required for bor consensus operations
-	state := wrappedState.(*state.StateDB)
+	// Get the underlying state db instance as we'll only interact with genesis contracts
+	// going ahead for consensus verification.
+	state := wrappedState.Inner()
 
 	var (
 		stateSyncData []*types.StateSyncData
@@ -916,7 +916,10 @@ func (c *Bor) Finalize(chain consensus.ChainHeaderReader, header *types.Header, 
 		state.BorConsensusTime = time.Since(start)
 	}
 
-	if err = c.changeContractCodeIfNeeded(headerNumber, state); err != nil {
+	// Check if any hardfork needs change in genesis contract code. Note that we use
+	// the wrapped state here as it may have a hooked state db instance which can help
+	// in tracing if it's enabled.
+	if err = c.changeContractCodeIfNeeded(headerNumber, wrappedState); err != nil {
 		log.Error("Error changing contract code", "error", err)
 		return
 	}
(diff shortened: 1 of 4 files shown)


maoueh:
@manav2401 ~While fixing the problem at directly, the fix is incorrect from a Live Tracer perspective. Indeed, you are now passing around a non-hooked StateDB, which means tracing will never be enabled for state changes happening within `Finalize`.~

~This is problematic as it means a bunch of Polygon state changes wouldn't be traced correctly anymore, since all operation here are now performed using a "vanilla" StateDB.~

**Review** Let me double check my claim, as you passed a hooked stateDB at least in `changeContractCodeIfNeeded`, I'll verify if it would traced everywhere correctly.

After review, there is still some places where the `wrappedState` would need to be passed along, in https://github.com/maticnetwork/bor/pull/1623/files#diff-e2acd76a9636afbcc13d882e2f46ca9bf74e0f9fa6c190c7ba75328a98fd364fL902 and the same below it.

But everything should work properly, we can update our own PR on top of yours just to adjust the changes related to `commitSpan` and the like.

manav2401:
Hi @maoueh, thanks for your inputs. I initially thought the remaining functions like `checkAndCommitSpans` and `CommitStates` won't need the wrapped db as they're not invoking functions where tracing is implemented. But they're doing calls to update the genesis contracts for spans and bridge events which means it needs tracing. I can update my PR to reflect those changes if that's fine. 

I think your PR might also work but there are some extra interface methods you've added which may not be needed.

manav2401:
Okay, I found a few minor issues around using state db correctly which I've fixed. I've also updated the PR to pass around `vm.StateDB` for other consensus functions.

maoueh:
@manav2401 yeah with the remaining passing around of `vm.StateDB` interface, I think the PR will not be needed anymore.

> But they're doing calls to update the genesis contracts for spans and bridge events which means it needs tracing. I can update my PR to reflect those changes if that's fine.

Yeah we have follow up PRs to wrap more nicely the tracing around those part too.

cffls (approved):
LGTM

kamuikatsurgi (approved):

PR #1625, opened 2025-07-10
core, eth, miner: interrupt block building during fill transactions incase of timeout

manav2401:
# Description

Use commit interrupt for fill transactions as well. 

We interrupt executing new transactions when the 2s block time mark is crossed. Although before execution, we call `Pending()` function of transaction pool which can be time consuming in some cases. If the 2s time limit is up, we'd like to interrupt that call as well to avoid spending any more time on doing any work for this block.

This PR passes the `interrupt` atomic boolean (which is used across the modules to trigger interrupt) when we're fetching the pending transactions and sorting them based on nonce and fee before executing. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1321,26 +1321,38 @@ func (w *worker) fillTransactions(interrupt *atomic.Int32, env *environment) err
 	filter.OnlyPlainTxs, filter.OnlyBlobTxs = true, false
 	pendingPlainTxs := w.eth.TxPool().Pending(filter, &w.interruptBlockBuilding)
 
+	filter.OnlyPlainTxs, filter.OnlyBlobTxs = false, true
+	pendingBlobTxs := w.eth.TxPool().Pending(filter, &w.interruptBlockBuilding)
+
 	// Split the pending transactions into locals and remotes.
 	prioPlainTxs, normalPlainTxs := make(map[common.Address][]*txpool.LazyTransaction), pendingPlainTxs
+	prioBlobTxs, normalBlobTxs := make(map[common.Address][]*txpool.LazyTransaction), pendingBlobTxs
 
 	for _, account := range prio {
 		if txs := normalPlainTxs[account]; len(txs) > 0 {
 			delete(normalPlainTxs, account)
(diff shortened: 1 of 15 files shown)


miner/worker.go
```diff
@@ -1337,22 +1339,11 @@ func (w *worker) fillTransactions(interrupt *atomic.Int32, env *environment) err
...
-	// bor: kept for reference, can remove later
-	/*
-		var (
-			localPlainTxs, remotePlainTxs, localBlobTxs, remoteBlobTxs map[common.Address][]*txpool.LazyTransaction
-		)
-	*/
-
 	filter.OnlyPlainTxs, filter.OnlyBlobTxs = true, false
-	pendingPlainTxs := w.eth.TxPool().Pending(filter)
+	pendingPlainTxs := w.eth.TxPool().Pending(filter, &w.interruptBlockBuilding)
```
cffls, line 1343: Should we use a lower timeout limit for this (I am assuming interruptBlockBuilding is 2s) so that there will still be time left for tx executions?

manav2401: We could reduce it but with the [tx pool optimisations PR](https://github.com/maticnetwork/bor/pull/1617), the wait time will be very less compared to earlier. 

My aim behind this PR was just to handle cases when last block was late and you've very less time (let's say 100ms) to process current block. It won't be worth spending that time for the pending call and instead would be good to exit asap so that you've full 2s for next block. 

But if you still think so, we could reduce the wait time to half so that there's some time for execution. So even if there's 100ms left, we spend 50ms fetching transactions and 50ms executing them so that we get at least something.

cffls: Thanks, your explanation makes sense. It would be hard to determine a good value for the timeout of `pending()`. In this case, I am fine with the current implementation.

kamuikatsurgi (approved):

PR #1637, opened 2025-07-15
Start GRPC server after backend assignment

lucca30:
# Description

This PR fixes a bug of a time window at startup which `s.backend == nil`, which causes panic in any GRPC call if enabled by heimdall and queried right after starting bor.

It was reported on this [issue](https://github.com/maticnetwork/bor/issues/1632)

# Changes


cffls:
Looks like a unit test has failed. lgtm otherwise.

lucca30:
Hum, that's true. Fixing it. Thanks

pratikspatil024 (approved):

PR #1641, opened 2025-07-16
Upload genesis as assets on releases

lucca30:
# Description

This PR aims to start uploading our genesis in our assets page so users can use it as the source of truth. Currently we faced an issue (https://github.com/maticnetwork/bor/issues/1618) where we updated the genesis and the hash mismatches the file on a tutorial. With this we can instruct users on trusting directly on the checksum from github.

# Changes

### DIFF
--- a/.github/workflows/mainnet_deb_profiles.yml
+++ b/.github/workflows/mainnet_deb_profiles.yml
@@ -400,3 +400,5 @@ jobs:
             packaging/deb/bor-pbss-mainnet-**.deb
             packaging/deb/bor-mainnet-**.deb.checksum
             packaging/deb/bor-pbss-mainnet-**.deb.checksum
+            builder/files/genesis-mainnet-v1.json
+            builder/files/genesis-amoy.json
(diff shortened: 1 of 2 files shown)


djpolygon (approved):

.github/workflows/mainnet_deb_profiles.yml
```diff
@@ -400,3 +400,5 @@ jobs:
             packaging/deb/bor-pbss-mainnet-**.deb
             packaging/deb/bor-mainnet-**.deb.checksum
             packaging/deb/bor-pbss-mainnet-**.deb.checksum
+            builder/files/genesis-mainnet-v1.json
+            builder/files/genesis-amoy.json
```
cffls: Should this be put in `amoy_deb_profiles.yml`?

lucca30: Humm, agree. Nice catch

lucca30: fixed, thanks

PR #1646, opened 2025-07-17
feat: witness validation

kamuikatsurgi:
# Description 

Adds witness validation while receiving witness from peers.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2673,19 +2673,18 @@ func (bc *BlockChain) insertChainWithWitnesses(chain types.Blocks, setHead bool,
 		// Process block using the parent state as reference point
 		pstart := time.Now()
 
-		// Validate the witness.
+		computeWitness := makeWitness
+
 		if witnesses != nil && len(witnesses) > it.processed()-1 && witnesses[it.processed()-1] != nil {
+			// 1. Validate the witness.
 			if err := stateless.ValidateWitnessPreState(witnesses[it.processed()-1], bc); err != nil {
 				log.Error("Witness validation failed during chain insertion", "blockNumber", block.Number(), "blockHash", block.Hash(), "err", err)
 				bc.reportBlock(block, &ProcessResult{}, err)
 				followupInterrupt.Store(true)
 				return nil, it.index, fmt.Errorf("witness validation failed: %w", err)
 			}
-		}
 
-		computeWitness := makeWitness
-
-		if witnesses != nil && len(witnesses) > it.processed()-1 && witnesses[it.processed()-1] != nil {
+			// 2. Set the witness to the statedb.
 			memdb := witnesses[it.processed()-1].MakeHashDB(bc.statedb.TrieDB().Disk())
 			bc.statedb.TrieDB().SetReadBackend(hashdb.New(memdb, triedb.HashDefaults.HashDB))
 			computeWitness = false
--- a/eth/fetcher/witness_manager.go
+++ b/eth/fetcher/witness_manager.go
@@ -10,10 +10,9 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/stateless"
 	"github.com/ethereum/go-ethereum/core/types"
-
-	//	"github.com/ethereum/go-ethereum/eth/protocols/eth" // Not directly needed here
 	"github.com/ethereum/go-ethereum/eth/protocols/eth"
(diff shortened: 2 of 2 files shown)


eth/downloader/queue.go
```diff
@@ -1094,6 +1100,17 @@ func (q *queue) DeliverWitnesses(id string, witnessData interface{}, meta interf
 			return
 		}
 		log.Trace("DeliverWitnesses: Successfully decoded witness RLP", "peer", id, "index", index)
+
+		// Log detailed witness information after successful decoding
+		log.Info("[Stateless] Successfully decoded witness from peer",
```
cffls: Let's use debug log level here.

eth/fetcher/witness_manager.go
```diff
@@ -328,6 +329,16 @@ func (m *witnessManager) handleBroadcast(msg *injectedWitnessMsg) {
 	hash := msg.witness.Header().Hash()
 	log.Debug("[wm] Processing injected witness", "peer", msg.peer, "hash", hash, "number", msg.witness.Header().Number.Uint64())
 
+	// Log witness reception from peer
+	log.Info("[Stateless] Received witness via broadcast injection",
```
cffls: Same as above. There are a few other places where log level should be debug.

eth/protocols/wit/handlers.go
```diff
@@ -3,9 +3,55 @@ package wit
 import (
 	"fmt"
 
+	"github.com/ethereum/go-ethereum/core/stateless"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rlp"
 )
 
+// validateWitnessPreState validates that the witness pre-state root matches the parent block's state root
+func validateWitnessPreState(witness *stateless.Witness, backend Backend, peerID string) error {
```
cffls: This function is almost the same as the one in `handler_wit.go`. Let's write one common function and use it in both files. The interface could be something like this:

```go
func validateWitnessPreState(witness *stateless.Witness, chain *core.BlockChain, peerID string) error
```

cffls (commented):
Looks pretty good! Please add unit tests.

eth/downloader/queue.go
kamuikatsurgi: These are just the logs I have used for myself. Would be removing and cleaning up the PR as we go.

eth/fetcher/witness_manager.go
kamuikatsurgi: Same.

eth/protocols/wit/handlers.go
kamuikatsurgi: Done.

kamuikatsurgi:
> Looks pretty good! Please add unit tests.

Addressed comments and added unit tests. This is ready for review.

core/stateless/witness.go
```diff
@@ -18,21 +18,61 @@ package stateless
...
+	if contextHeader == nil {
+		return fmt.Errorf("witness context header is nil")
+	}
+
+	// Get the parent block header from the chain.
+	parentHeader := headerReader.GetHeader(contextHeader.ParentHash, contextHeader.Number.Uint64()-1)
+	if parentHeader == nil {
+		return fmt.Errorf("parent block header not found: parentHash=%x, parentNumber=%d",
+			contextHeader.ParentHash, contextHeader.Number.Uint64()-1)
+	}
```
cffls, line 58: One concern of this approach is that it might prevent a client from bulk downloading witnesses. We might need to move this to the place where the witness is actually being used for importing its block (in `blockchain.go`), since at that point we knew the block's parent must have been known at that point.

kamuikatsurgi: Good point. I changed the witness validation logic to be only in the `blockchain.go`.

core/blockchain.go
```diff
@@ -2673,6 +2673,16 @@ func (bc *BlockChain) insertChainWithWitnesses(chain types.Blocks, setHead bool,
 		// Process block using the parent state as reference point
 		pstart := time.Now()
 
+		// Validate the witness.
+		if witnesses != nil && len(witnesses) > it.processed()-1 && witnesses[it.processed()-1] != nil {
```
cffls, line 2678: This statement is essentially the same as line 2688, can these two be combined?

kamuikatsurgi: Addressed.

cffls (approved):

PR #1652, opened 2025-07-22
core: witness compression

pratikspatil024:
# Description

This PR introduces gzip compression and deduplication for witness data in Bor, reducing the size of witness objects stored and transmitted across the network.

These are the configurable params
```
type CompressionConfig struct {
    Enabled          bool  // Enable/disable compression
    Threshold        int   // Minimum size to consider compression
    CompressionLevel int   // Gzip compression level (1-9)
    UseDeduplication bool  // Enable witness optimization
}
```

- Compression Marker: Uses 0x01 for compressed, 0x00 for uncompressed
- Smart Compression: Only compresses when it actually reduces size
- Configurable Threshold: Only compresses data larger than a configurable threshold
- Code Normalization: Removes trailing zeros from bytecode
- State Deduplication: Removes empty or redundant state nodes

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1962,7 +1962,7 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 
 	if statedb.Witness() != nil {
 		var witBuf bytes.Buffer
-		if err := statedb.Witness().EncodeRLP(&witBuf); err != nil {
+		if err := statedb.Witness().EncodeCompressed(&witBuf); err != nil {
 			log.Error("error in witness encoding", "caughterr", err)
 		}
 
--- a/core/stateless/encoding.go
+++ b/core/stateless/encoding.go
@@ -21,63 +21,71 @@ import (
 	"compress/gzip"
 	"errors"
 	"io"
-	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/metrics"
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
 // Compression metrics
 var (
-	compressionRatio    int64
-	compressionCount    int64
-	uncompressedCount   int64
-	totalOriginalSize   int64
-	totalCompressedSize int64
+	// Counters for operations
+	compressionCount   = metrics.NewRegisteredCounter("witness/compression/count", nil)
+	uncompressedCount  = metrics.NewRegisteredCounter("witness/uncompressed/count", nil)
+	decompressionCount = metrics.NewRegisteredCounter("witness/decompression/count", nil)
 
-	// Compression timing and rate metrics
-	totalCompressionTime int64 // nanoseconds
-	totalCompressionSize int64 // total size of compressed data
-	compressionRate      int64 // bytes per second
+	// Gauges for current values
+	compressionRatio    = metrics.NewRegisteredGauge("witness/compression/ratio", nil)
+	totalOriginalSize   = metrics.NewRegisteredGauge("witness/compression/original_size", nil)
+	totalCompressedSize = metrics.NewRegisteredGauge("witness/compression/compressed_size", nil)
+	spaceSavedBytes     = metrics.NewRegisteredGauge("witness/compression/space_saved", nil)
 
-	// Decompression metrics
-	decompressionCount     int64
-	totalDecompressionTime int64 // nanoseconds
-	totalDecompressionSize int64 // total size of decompressed data
-	decompressionRate      int64 // bytes per second
+	// Timers for timing operations
+	compressionTimer   = metrics.NewRegisteredTimer("witness/compression/timer", nil)
+	decompressionTimer = metrics.NewRegisteredTimer("witness/decompression/timer", nil)
+
+	// Meters for rates
+	compressionRate   = metrics.NewRegisteredMeter("witness/compression/rate", nil)
+	decompressionRate = metrics.NewRegisteredMeter("witness/decompression/rate", nil)
 )
 
 // CompressionStats returns current compression statistics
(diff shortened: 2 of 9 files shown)


core/stateless/encoding.go
```diff
@@ -17,12 +17,128 @@
...
+// CompressionConfig holds configuration for witness compression
+type CompressionConfig struct {
+	Enabled          bool // Enable/disable compression
+	Threshold        int  // Threshold in bytes. Only compress if witness is larger than this.
+	CompressionLevel int  // Gzip compression level (1-9)
+	UseDeduplication bool // Enable witness optimization
+}
+
+// DefaultCompressionConfig returns the default compression configuration
+func DefaultCompressionConfig() *CompressionConfig {
```
pratikspatil024, line 128: @cffls @lucca30 - we need to decide on the default values.

core/stateless/encoding.go
```diff
@@ -64,9 +194,126 @@ func (w *Witness) DecodeRLP(s *rlp.Stream) error {
...
+	}
+
+	return w.fromExtWitness(&ext)
+}
+
 // extWitness is a witness RLP encoding for transferring across clients.
 type extWitness struct {
 	Context *types.Header
 	Headers []*types.Header
+	Codes   [][]byte
```
pratikspatil024: I'm not 100% sure about this, but had to do it because without it this field would not be compressed/decompressed.

core/blockchain.go
```diff
@@ -1962,7 +1962,7 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 
 	if statedb.Witness() != nil {
 		var witBuf bytes.Buffer
-		if err := statedb.Witness().EncodeRLP(&witBuf); err != nil {
+		if err := statedb.Witness().EncodeCompressed(&witBuf); err != nil {
```
lucca30, line 1965: `writeBlockWithState` it's part of the lifecycle of creating/importing a block. So this will potentially slow the production/import time of a block.

I would suggest calling it in a goroutine so it not blocks the process itself. The only concerns you should take care of are the ones regarding reading this witness after this step. Because probably some steps are assuming the previous step already wrote the witness.

core/stateless/encoding.go
lucca30: you can delete this part, since code it's not more transferred

core/blockchain.go
cffls: If we use a go routine to store it, a peer may request this witness before it is actually ready in the db and cause problems. Maybe we should use a memory cache to store the witness so they will be instantly available to server witness queries.

core/stateless/encoding.go
cffls: Let's use the best compression by default and check the compression metrics in benchmarking tests.

core/stateless/witness.go
```diff
@@ -134,6 +135,61 @@ func (w *Witness) AddState(nodes map[string]struct{}) {
...
+	defer w.lock.Unlock()
+
+	// Remove duplicate codes by normalizing them
+	normalizedCodes := make(map[string]struct{})
+	for code := range w.Codes {
+		// Remove leading/trailing zeros that don't affect execution
+		normalized := bytes.TrimRight([]byte(code), "\x00")
+		normalizedCodes[string(normalized)] = struct{}{}
+	}
+	w.Codes = normalizedCodes
```
cffls: This could be removed because codes won't be used.

core/blockchain.go
lucca30: That's a nice idea!
To save memory, we can also mark the witness’s write operation as “pending” and, if a read arrives while it’s pending, simply wait for it to complete before returning.

I don't see any read scenario where waiting would be a problem.

core/stateless/witness.go
pratikspatil024: Removed, thanks!

core/stateless/encoding.go
pratikspatil024: Removed, thanks!

core/blockchain.go
pratikspatil024: Created https://polygon.atlassian.net/browse/POS-3161 for this

core/stateless/encoding.go
```diff
@@ -17,12 +17,128 @@
...
+// Compression metrics
+var (
+	compressionRatio    int64
+	compressionCount    int64
+	uncompressedCount   int64
+	totalOriginalSize   int64
+	totalCompressedSize int64
+
+	// Compression timing and rate metrics
+	totalCompressionTime int64 // nanoseconds
```
cffls: We can use a timer metric and it would be added to prometheus automatically.
Similar to block execution timer: https://github.com/maticnetwork/bor/blob/master/core/blockchain.go#L97

core/stateless/encoding.go
```diff
@@ -17,12 +17,128 @@
...
+	totalCompressionSize int64 // total size of compressed data
```
cffls: Similarly, we can use a meter here.

core/stateless/encoding.go
pratikspatil024: Updated, [thanks](https://github.com/maticnetwork/bor/pull/1652/commits/4042e8063ba7edb6809858a4b303a5e0b5da8fd9)!

core/stateless/encoding.go
pratikspatil024: Updated, [thanks](https://github.com/maticnetwork/bor/pull/1652/commits/4042e8063ba7edb6809858a4b303a5e0b5da8fd9)!

PR #1662, opened 2025-07-25
Parallel block import

Raneet10:
# Description

This PR adds the feature to allow a node to import blocks in parallel in stateless mode. 

```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/core
cpu: Apple M2 Pro
                                  │ bench/stateless_seq_100.txt │ bench/stateless_par_100.txt │
                                  │           sec/op            │        sec/op         vs base   │
InsertChainStatelessSequential-10                   56.51µ ± 0%
InsertChainStatelessParallel-10                                            42.52µ ± 1%
geomean                                             56.51µ                 42.52µ       ? ¹ ²
¹ benchmark set differs from baseline; geomeans may not be comparable
² ratios must be >0 to compute geomean
```
Parallel import currently is ~25% faster. 

# Changes

# Checklist

## Testing

- [x] I have added unit tests

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -696,12 +696,6 @@ type WitnessConfig struct {
 
 	// Minimum necessary distance between local header and peer to fast forward
 	FastForwardThreshold uint64 `hcl:"fastforwardthreshold,optional" toml:"fastforwardthreshold,optional"`
-
-	// Minimum necessary distance between local header and latest non pruned witness
-	PruneThreshold uint64 `hcl:"prunethreshold,optional" toml:"prunethreshold,optional"`
-
-	// The time interval between each witness prune routine
-	PruneInterval time.Duration `hcl:"pruneinterval,optional" toml:"pruneinterval,optional"`
 }
 
 func DefaultConfig() *Config {
@@ -919,8 +913,6 @@ func DefaultConfig() *Config {
 			ParallelStatelessImportWorkers: 0,
 			WitnessAPI:                     false,
 			FastForwardThreshold:           6400,
-			PruneThreshold:                 64000,
-			PruneInterval:                  120 * time.Second,
 		},
 		History: &HistoryConfig{
 			TransactionHistory: ethconfig.Defaults.TransactionHistory,


cffls:
From the [run log](https://github.com/0xPolygon/bor/actions/runs/17155850554?pr=1662), it looks like one issue is that the node won't be able to process a block/header unless its parent is known.

Screenshot 2025-08-22 at 10 54 45 AM

core/blockchain.go
```diff
@@ -2243,36 +2304,253 @@ func (bc *BlockChain) InsertChainStateless(chain types.Blocks, witnesses []*stat
...
+			// stop header verification from producing more results
+			abortCloser.Do(func() { close(abort) })
+			// unblock any WaitForParent waiters
+			bs.Abort()
+		})
+	}
 	for i, block := range chain {
-		// Check if block is already known before attempting to process
+		wg.Add(1)
+		go func(block *types.Block, i int) {
```
cffls: This is a very big function. Could you move it to a separate function and potentially break it down?

core/blockchain.go
```diff
@@ -2243,36 +2304,253 @@ func (bc *BlockChain) InsertChainStateless(chain types.Blocks, witnesses []*stat
...
+		return bc.insertChainStatelessSequential(chain, witnesses, errChans, &stats)
+	}
+
+	var wg sync.WaitGroup
+
+	// In stateless mode, we process blocks parallely without committing the state
+	var processed atomic.Int32
+
+	bs := NewBlockState() // Block state tracker to manage parallel block processing
+	bc.parallelImport.Store(true)
```
cffls: Let's add a config to enable/disable parallel block processing.

core/blockchain_test.go
```diff
@@ -4302,7 +4303,7 @@ func TestDeleteThenCreate(t *testing.T) {
...
...806100206000396000f3fe608060405234801561001057600080fd5b506004361061002a5760003560e01c80627743601461002f575b600080fd5b610049600480360381019061004491906100d8565b61004b565b005b6000808251602084016000f59050803b61006457600080fd5b5050565b600061007b61007684610146565b610121565b905082815260208101848484011115610097576100966101eb565b5b6100a2848285610177565b509392505050565b600082601f8301126100bf576100be6101e6565b5b81356100cf848260208601610068565b91505092915050565b6000602082840312156100ee576100ed6101f5565b5b600082013567ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff191690815f1a90535050565b6000604051905090565b600067ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff1115610161576101606101b7565b5b61016a826101fa565b9050602081019050919050565b82818337600083830152505050565b61018f826101fa565b810181811067ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff17156101ae576101ad6101b7565b5b80604052505050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052604160045260246000fd5b600080fd5b600080fd5b600080fd5b600080fd5b6000601f19601f830116905091905056fea2646970667358221220ea8b35ed310d03b6b3deef166941140b4d9e90ea2c92f6b41eb441daf49a59c364736f6c63430008070033")
```
cffls: Why is it required to change this?

core/blockchain.go
Raneet10: Done.

core/blockchain.go
Raneet10: Done.

core/blockchain_test.go
Raneet10: Ouch. Yeah reverted this.

pratikspatil024:
After the recent changes, in testing, I observed that a block is never imported sequentially.
All the blocks, even 0 txn blocks are imported in parallel.

Raneet10:
> After the recent changes, in testing, I observed that a block is never imported sequentially. All the blocks, even 0 txn blocks are imported in parallel.

This is expected since I reverted the logic where sequential import is done in [this](https://github.com/0xPolygon/bor/pull/1662/commits/e990f09e64aa80c9af389cfe2a9d73802c510f67) commit. Let's see whether latency tests pass in the CI otherwise we'll re-enable the condition.

core/blockchain.go
```diff
@@ -2208,142 +2209,314 @@ func (bc *BlockChain) InsertChainStateless(chain types.Blocks, witnesses []*stat
...
+	type execResult struct {
+		sdb        *state.StateDB
+		err        error
+		needsRetry bool
+		witness    *stateless.Witness
+	}
+	results := make([]execResult, len(chain))
+	workCh := make(chan int, len(chain))
+	var wg sync.WaitGroup
+	numWorkers := runtime.GOMAXPROCS(0)
```
cffls, line 2556: Let's make this configurable

core/blockchain.go
```diff
@@ -2208,142 +2209,314 @@ func (bc *BlockChain) InsertChainStateless(chain types.Blocks, witnesses []*stat
...
+					continue
+				}
+				var witness *stateless.Witness
+				if idx < len(witnesses) {
+					witness = witnesses[idx]
+				}
+				sdb, perr := bc.ProcessBlockWithWitnesses(blk, witness)
+				if perr != nil {
+					// If stateless self-validation depends on parent's commit, mark for retry in writer stage
+					if idx > 0 && strings.Contains(perr.Error(), "stateless self-validation state root mismatch") {
```
cffls: Please create a specific error type (or reuse one if there is an existing one) for state root mismatch.

core/blockchain.go
```diff
@@ -2208,142 +2209,314 @@ func (bc *BlockChain) InsertChainStateless(chain types.Blocks, witnesses []*stat
...
+	if !bc.parallelStatelessImportEnabled.Load() {
+		return bc.insertChainStatelessSequential(chain, witnesses, errChans, &stats)
+	}
+
+	return bc.insertChainStatelessParallel(chain, witnesses, errChans, &stats, stopHeaders)
+}
 
+// insertChainStatelessSequential imports a small batch of blocks sequentially in stateless mode.
+func (bc *BlockChain) insertChainStatelessSequential(chain types.Blocks, witnesses []*stateless.Witness, errChans []chan error, stats *insertStats) (int, error) {
+	log.Info("Perfoming sequential stateless import", "chain length", len(chain))
```
cffls: Use debug?

core/blockchain.go
```diff
@@ -2208,142 +2209,314 @@ func (bc *BlockChain) InsertChainStateless(chain types.Blocks, witnesses []*stat
...
-			continue // Skip to the next block
+			continue
 		}
-		// If the chain is terminating, don't even bother starting up.
 		if bc.insertStopped() {
-			return processed, errInsertionInterrupted
-		}
-
-		// Wait for the block's verification to complete
-		if err := <-results; err != nil {
```
cffls, line 2508: The following checks are removed. I guess they haven't been equivalently checked elsewhere?

core/blockchain_test.go
```diff
@@ -5261,6 +5261,88 @@ func TestStatelessInsertChain(t *testing.T) {
...
+		t.Fatalf("expected processed=%d, got %d", len(blocksSequential), processed)
+	}
+
+	if chain.cacheConfig.TriesInMemory != 0 {
+		t.Error("Expected blockchain to be configured for stateless mode (TriesInMemory = 0)")
+	}
+}
+
+// TestStatelessInsertChainInvalidInputs tests invalid or edge inputs for InsertChainStateless
+func TestStatelessInsertChainInvalidInputs(t *testing.T) {
```
cffls: Can we add more adversarial test cases where bad inputs are passed. e.g. bad witness with one byte altered, missing trie nodes, witnesses that will compute to incorrect state roots.

internal/cli/server/config.go
```diff
@@ -854,12 +857,13 @@ func DefaultConfig() *Config {
...
-			Enable:               false,
-			SyncWithWitnesses:    false,
-			ProduceWitnesses:     false,
-			FastForwardThreshold: 6400,
-			PruneThreshold:       64000,
-			PruneInterval:        120 * time.Second,
+			Enable:                        false,
+			SyncWithWitnesses:             false,
+			ProduceWitnesses:              false,
+			EnableParallelStatelessImport: false,
```
cffls: Would be great to have a node with parallel block import enabled in stateless e2e test.

core/blockchain.go
Raneet10: Maybe @pratikspatil024 had a reason: [5da6523](https://github.com/0xPolygon/bor/pull/1662/commits/5da6523ee2747a627736062fdf46d2cdbeddc655) ?

core/blockchain_test.go
Raneet10: Good idea. Just wondering, were these cases not a problem in sequential import ? They should be AFAIK.

internal/cli/server/config.go
Raneet10: So basically have it enabled by default ?

Raneet10: So you mean the stateless tests running till now were sequential ? 💀 
But I wonder how come the latency tests were failing in the earlier approach ?

core/blockchain.go
Raneet10: They're [here](https://github.com/0xPolygon/bor/pull/1662/files#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215R2241) I suppose.

internal/cli/server/config.go
cffls: It seems like it was always sequential before. Maybe previous failure were due to issues in sequential execution itself.

core/blockchain_test.go
cffls: In these test cases, there will be errors in sequential regardless. However, it is important to test them in parallel to ensure parallel execution doesn't miss catching these errors.

internal/cli/server/config.go
Raneet10: Done.

core/blockchain_test.go
Raneet10: Done. Please take a look and let me know if it looks good.

core/blockchain.go
Raneet10: Done.

core/blockchain.go
Raneet10: Done.

core/blockchain.go
pratikspatil024: reverted https://github.com/0xPolygon/bor/pull/1662/commits/029bc3e19f56b00f41ad97a0b5d5071a53b68082

core/blockchain.go
cffls: Is it configurable now?

internal/cli/server/config.go
```diff
@@ -669,6 +669,12 @@ type WitnessConfig struct {
 	// ProduceWitnesses enables producing witnesses while syncing
 	ProduceWitnesses bool `hcl:"producewitnesses,optional" toml:"producewitnesses,optional"`
 
+	// Parallel stateless import (download path) toggle
+	EnableParallelStatelessImport bool `hcl:"parallelstatelessimport,optional" toml:"parallelstatelessimport,optional"`
```
cffls: Can we put `-` between words for this flag?

.github/workflows/kurtosis-stateless-e2e.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - 'master'
       - 'develop'
+      - v2.3.0-candidate # Remove after testing
```
pratikspatil024: should we remove this?

.github/workflows/kurtosis-stateless-e2e.yml
```diff
@@ -58,7 +59,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           repository: 0xPolygon/kurtosis-pos
-          ref: stateless
+          ref: raneet10/stateless-parallel-import # Revert after testing
```
pratikspatil024: and update this?

internal/cli/server/config.go
```diff
@@ -685,11 +685,23 @@ type WitnessConfig struct {
...
+	ParallelStatelessImportWorkers int `hcl:"parallel-stateless-import-workers,optional" toml:"parallel-stateless-import-workers,optional"`
+
 	// WitnessAPI enables witness API endpoints
 	WitnessAPI bool `hcl:"witnessapi,optional" toml:"witnessapi,optional"`
 
 	// Minimum necessary distance between local header and peer to fast forward
 	FastForwardThreshold uint64 `hcl:"fastforwardthreshold,optional" toml:"fastforwardthreshold,optional"`
+
+	// Minimum necessary distance between local header and latest non pruned witness
+	PruneThreshold uint64 `hcl:"prunethreshold,optional" toml:"prunethreshold,optional"`
```
pratikspatil024: I don't see these 2 flags (`PruneThreshold` and `PruneInterval`) in flags.go file

pratikspatil024: Also, where are these 2 flags used?

.github/workflows/kurtosis-stateless-e2e.yml
Raneet10: We need this because parallel import config is in this branch.

internal/cli/server/config.go
Raneet10: Not sure. These appeared when I merged `develop`.

.github/workflows/kurtosis-stateless-e2e.yml
Raneet10: Done

internal/cli/server/config.go
Raneet10: Done.

internal/cli/server/config.go
Raneet10: Reverted.

PR #1665, opened 2025-07-28
Stateless sync code cache fix

cffls:
Fixing byte code download when a stateless node needs to fast forward a recent block. 
The previous issue was that trie node healer in snap sync was disabled completely for stateless node. However, this led to a problem when the pivot of snap sync moves, resulting the node not being able to download all the contracts correctly.

This change will try to allow snap sync pivot to move when downloading bytecodes. Because of moving pivot, the node will need to go through trie node and byte code healing process. In order to speed up healing speed, the merkle path whose lengths are less than 6 nibbles or who leads to a bytecode trie node will be stored. 

In PoS mainnet (~74M blocks), the total contract size is approximately 30GB, and their associated trie nodes is 60GB.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1981,11 +1981,7 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 		}
 	}
 
-	batch := bc.db.NewBatch()
-	rawdb.WriteBytecodeSyncLastBlock(batch, block.NumberU64())
-	if err := batch.Write(); err != nil {
-		log.Crit("Failed to write bytecode sync last block into disk", "err", err)
-	}
+	rawdb.WriteBytecodeSyncLastBlock(bc.db, block.NumberU64())
 
 	// If node is running in path mode, skip explicit gc operation
 	// which is unnecessary in this mode.
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1986,6 +1986,13 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 			return []*types.Log{}, err
 		}
 	}
+
+	batch := bc.db.NewBatch()
+	rawdb.WriteBytecodeSyncLastBlock(batch, block.NumberU64())
+	if err := batch.Write(); err != nil {
+		log.Crit("Failed to write bytecode sync last block into disk", "err", err)
+	}
+
 	// If node is running in path mode, skip explicit gc operation
 	// which is unnecessary in this mode.
 	if bc.triedb.Scheme() == rawdb.PathScheme {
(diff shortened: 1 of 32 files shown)


core/blockchain.go
```diff
@@ -1986,6 +1986,13 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 			return []*types.Log{}, err
 		}
 	}
+
+	batch := bc.db.NewBatch()
+	rawdb.WriteBytecodeSyncLastBlock(batch, block.NumberU64())
+	if err := batch.Write(); err != nil {
+		log.Crit("Failed to write bytecode sync last block into disk", "err", err)
+	}
```
lucca30: What about just 

```
rawdb.WriteBytecodeSyncLastBlock(bc.db, block.NumberU64())
```

Instead of batch, since there is no other write/delete operation?

eth/downloader/bor_downloader.go
```diff
@@ -758,15 +761,43 @@ func (d *Downloader) syncWithPeer(p *peerConnection, hash common.Hash, td, ttd *
...
+
+		// Ensure snap syncer is fully terminated before proceeding
+		d.Cancel()
+		d.cancelWg.Wait()
+
+		batch := d.stateDB.NewBatch()
+		rawdb.WriteBytecodeSyncLastBlock(batch, origin)
+		if err := batch.Write(); err != nil {
+			return err
+		}
```
lucca30: Same here

eth/downloader/bor_downloader.go
```diff
@@ -758,15 +761,43 @@ func (d *Downloader) syncWithPeer(p *peerConnection, hash common.Hash, td, ttd *
...
+		d.pivotLock.Unlock()
+
+		err := d.spawnSyncDominant(fetchers, func() error { return d.processSnapSyncContent(false) })
+		if err != nil {
+			return err
+		}
+
+		// Ensure snap syncer is fully terminated before proceeding
+		d.Cancel()
+		d.cancelWg.Wait()
```
lucca30: `d.Cancel()` already waits for the `d.cancelWg`, so no need to wait again

eth/downloader/bor_downloader.go
```diff
@@ -1838,7 +1919,14 @@ func (d *Downloader) processSnapSyncContent() error {
...
 				log.Warn("Pivot became stale, moving", "old", pivot.Number.Uint64(), "new", height-uint64(fsMinFullBlocks)+uint64(reorgProtHeaderDelay))
-				pivot = results[len(results)-1-fsMinFullBlocks+reorgProtHeaderDelay].Header // must exist as lower old pivot is uncommitted
+
+				newPivotNum := len(results) - 1 - fsMinFullBlocks + reorgProtHeaderDelay
+
+				if newPivotNum < 0 || newPivotNum > len(results)-1 {
+					newPivotNum = len(results) - 1
+				}
+
+				pivot = results[newPivotNum].Header // must exist as lower old pivot is uncommitted
```
lucca30: Should the `"new"` on the log be also updated?

core/blockchain.go
cffls: You are right, there isn't much benefit of using new batch here. Will delete it.

eth/downloader/bor_downloader.go
cffls: Nice catch! Update the logging.

PR #1666, opened 2025-07-29
Fix lint issues

lucca30:
# Description

Fixing lint. It helped finding a potential bug on `witReqs` appending.

# Changes

### DIFF
--- a/consensus/bor/snapshot.go
+++ b/consensus/bor/snapshot.go
@@ -179,32 +179,32 @@ func (s *Snapshot) apply(headers []*types.Header, c *Bor) (*Snapshot, error) {
 	return snap, nil
 }
 
-func (s *Snapshot) applyNumber(number uint64, c *Bor) (*Snapshot, error) {
-	if s.Number != number-1 {
-		return nil, errOutOfRangeChain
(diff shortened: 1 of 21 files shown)


cffls (approved):

consensus/bor/snapshot.go
```diff
@@ -179,32 +179,32 @@ func (s *Snapshot) apply(headers []*types.Header, c *Bor) (*Snapshot, error) {
...
-	}
-	v := getUpdatedValidatorSet(snap.ValidatorSet.Copy(), producers)
-
-	v.IncrementProposerPriority(1)
-
-	snap.ValidatorSet = v
-
-	return snap, nil
-}
+// func (s *Snapshot) applyNumber(number uint64, c *Bor) (*Snapshot, error) {
```
kamuikatsurgi: If not needed, should we remove the code? This would reduce code debt.

kamuikatsurgi (approved):

consensus/bor/snapshot.go
lucca30: agree

PR #1681, opened 2025-08-04
Stateless sync fix wit download

cffls:
This change will avoid sending witness request to peers that don't support `wit` protocol. Also fixes a sync issue where witness queue is hanging due to the node is incorrectly waiting for peers that don't support `wit` protocol.

### DIFF
--- a/eth/downloader/bor_fetchers_concurrent.go
+++ b/eth/downloader/bor_fetchers_concurrent.go
@@ -145,9 +145,18 @@ func (d *Downloader) concurrentFetch(queue typedQueue, beaconMode bool) error {
 				caps  []int
 			)
 
+			// Check if we're fetching witnesses to filter peers appropriately
+			isWitnessQueue := reflect.TypeOf(queue) == reflect.TypeOf(&witnessQueue{})
+
 			for _, peer := range d.peers.AllPeers() {
 				pending, stale := pending[peer.id], stales[peer.id]
 				if pending == nil && stale == nil {
+					// For witness fetching, skip peers that don't support the witness protocol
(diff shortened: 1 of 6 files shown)


eth/downloader/bor_fetchers_concurrent.go
```diff
@@ -221,7 +230,20 @@ func (d *Downloader) concurrentFetch(queue typedQueue, beaconMode bool) error {
...
+					if peer.peer.SupportsWitness() {
+						capablePeers++
+					}
+				}
+			} else {
+				// For other queues, all peers are capable
+				capablePeers = d.peers.Len()
+			}
+
+			if !progressed && !throttled && len(pending) == 0 && len(idles) == capablePeers && queued > 0 && !beaconMode {
```
manav2401, line 246: This check does `len(idles) == capablePeers`. Even if there's a non-overlapping set of idle peers and peers being able to support witness protocol, won't this check return an error in those cases? 

E.g. if total peers are 20, 10 are idle and 10 support witness. But, we don't know if those 10 peers are same ones so it would be unfair to return an error in such case.

cffls: When the current queue is witness type, a peer will only be added to idle list if they support witness protocol, see the code here: https://github.com/0xPolygon/bor/pull/1681/files#diff-e25923cb3dd24f59e479ec1d00b185cd4f566e7264b4379966622c0572714800R154-R160 

We'd like to return the error early if all peers are tried by not progress is being made.

manav2401: Ah okay didn't see that as it wasn't in the diff. Thanks!

Issue #1654, opened 2025-07-23
Bor missing state sync transactions and errors on logs


quickchase:
This is a probably a continuation of #1607 - although that issues was closed, the only reason it was "fixed" was  because we re-sync the nodes.

`bor/v2.2.9/linux-amd64/go1.24.4`

Our full nodes are returning the following error for `eth_getBlockReceipts`:

```bash
curl -s -H 'Content-Type: application/json' http://localhost:8545 -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x1721E00"]}'  | jq
```

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32000,
    "message": "receipts length mismatch: 1 vs 0"
  }
}
```

And when we do `eth_getBlockByNumber`, it's returning 1 transaction, not 2:

```bash
curl -s -H 'Content-Type: application/json' http://localhost:8545 -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x1721E00",false]}' | jq
```

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "baseFeePerGas": "0x3f",
    "difficulty": "0x5",
    "extraData": "0xd783010f0b83626f7288676f312e32342e34856c696e75780000000000000000c380c1c01453a9cbd107a90384575d3a7f6b065a63d284cd813afd98ed610ab3925f5ca225db0c0931bb045c965a980016ade2f9cba57942706413a69d99a7c9ffeada1100",
    "gasLimit": "0x2aea540",
    "gasUsed": "0x9ca70e",
    "hash": "0x0704da40e68666ffb35a933ec2bf662436dde1c5cfefa22a808e5ac1798e9d17",
[...]

PR #1684, opened 2025-08-06
State Sync check consistency endpoint

lucca30:
# Description

Adds a new endpoint, checkStateSyncConsistency(startBlock, endBlock), which scans a block-range on a Bor chain and reports any missing Heimdall‐originated state‐sync transactions in the local database.

This check provides a simple, automated way to verify consistency between on-chain state-sync events (Heimdall’s source of truth) and the node’s rawdb.

# Changes

## Testing

- [ ] I have added unit tests


lucca30:
Referencing https://github.com/0xPolygon/bor/issues/1654 , since it will help us easily identify any inconsistency throughout the nodes.

kamuikatsurgi:
We should first deploy this to sentry as well as val, since we have some API changes regarding state syncs.

consensus/bor/IHeimdallWSClient.go
```diff
@@ -0,0 +1,14 @@
+package bor
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum/consensus/bor/heimdall/milestone"
+)
+
+//go:generate mockgen -source=IHeimdallWSClient.go -destination=../../tests/bor/mocks/MockIHeimdallWSClient.go -package=mocks . IHeimdallWSClient
```
cffls, line 9: We can simplify this by moving this to the same file as `IHeimdallClient` and remove the `go:generate` annotation and all mock files could be auto-generated through `make generate-mocks` command.

In stateless branch, this practice is followed: https://github.com/0xPolygon/bor/blob/stateless_sync/consensus/bor/heimdall.go#L13, and all mocks are generated in https://github.com/0xPolygon/bor/blob/stateless_sync/tests/bor/mocks/IHeimdallClient.go

lucca30: I'm actually using the `make generate-mocks` here in this branch. afaik it runs the generate based exactly on the interfaces that have the `go:generate` annotation, so I think it's necessary. I decided separating it to make it more clear and visible. In the same way I renamed the file to make it more clear its an interface.

cffls: Okay that sounds good, thanks for the clarification.

cffls (approved):

PR #1694, opened 2025-08-12
Upstream v1.16.1

avalkov:
# Description

Upstream v1.16.1


eth/downloader/bor_downloader_test.go
```diff
@@ -333,9 +332,10 @@ func (dlp *downloadTesterPeer) RequestBodies(hashes []common.Hash, sink chan *et
 // peer in the download tester. The returned function can be used to retrieve
 // batches of block receipts from the particularly requested peer.
 func (dlp *downloadTesterPeer) RequestReceipts(hashes []common.Hash, sink chan *eth.Response) (*eth.Request, error) {
-	blobs := eth.ServiceGetReceiptsQuery(dlp.chain, hashes)
+	// TODO v1.16.1: Should we change these tests to version 69?
```
manav2401: Yes, this should be changed to eth/69 or we can have 2 tests one with 68 and another with 69.

avalkov: Changed to protocol 69

cmd/utils/flags.go
```diff
@@ -113,6 +113,11 @@ var (
 		Usage:    "Root directory for ancient data (default = inside chaindata)",
 		Category: flags.EthCategory,
 	}
+	EraFlag = &flags.DirectoryFlag{
```
kamuikatsurgi, line 116: Should we add a similar flag to the bor flag set, because iirc we don't use flags from `cmd/geth`? Enable this via that bor flag?

avalkov: Good point, I think we should discuss if we need this eradb for historical pruned blocks, I think it may need some additional work as we have some custom pruning logic.

core/blockchain.go
```diff
@@ -644,42 +713,32 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header, wit
...
-			return nil, nil, 0, nil, 0, err
-		}
-
 		processorCount++
 
 		go func() {
-			parallelStatedb.StartPrefetcher("chain", witness)
 			pstart := time.Now()
-			res, err := bc.parallelProcessor.Process(block, parallelStatedb, bc.vmConfig, nil, ctx)
+			statedb.StartPrefetcher("chain", witness)
```
cffls: Parallel statedb can't be shared with serial statedb. They need to be created separately. I guess that was the reason why there was issue when block-stm was enabled.

core/blockchain.go
```diff
@@ -1793,8 +1894,10 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 				}
 			}
 			// Write all the data out into the database
-			rawdb.WriteBody(batch, block.Hash(), block.NumberU64(), block.Body())
-			rawdb.WriteReceipts(batch, block.Hash(), block.NumberU64(), receiptChain[i])
+			// TODO v1.16.1: Which of the two?
```
cffls: I think it makes sense to keep the way it is. WriteBlock = Write body + header, and it makes sense to call WriteCanonicalHash here because this function is called in snapsync. When the pivot block of the snapsync has moved forward, the delta between old pivot and new pivot would be inserted as receipt chain, which makes sense to be marked as canonical.

cffls (commented):
Pruner-related changes might need review from @manav2401, who had worked on pruning in the past.

go.mod
```diff
@@ -18,15 +18,14 @@ require (
 	github.com/aws/aws-sdk-go-v2/credentials v1.17.18
 	github.com/aws/aws-sdk-go-v2/service/route53 v1.40.10
 	github.com/cespare/cp v1.1.1
-	github.com/cloudflare/cloudflare-go v0.97.0
-	github.com/cockroachdb/pebble v1.1.2
+	github.com/cloudflare/cloudflare-go v0.114.0
```
marcello33, line 21: Shall we take this opportunity to bump the following deps?
- github.com/0xPolygon/heimdall-v2 to `v0.3.1`
- github.com/cosmos/cosmos-sdk v0.50.13 to `v0.50.14` (it gets replaced but it's just to quit some sec warnings)
- github.com/0xPolygon/bor to latest stable version

beacon/blsync/engineclient.go
```diff
@@ -104,7 +105,11 @@ func (ec *engineClient) callNewPayload(fork string, event types.ChainHeadEvent)
 		method = "engine_newPayloadV4"
 		parentBeaconRoot := event.BeaconHead.ParentRoot
 		blobHashes := collectBlobHashes(event.Block)
-		params = append(params, blobHashes, parentBeaconRoot, event.ExecRequests)
+		hexRequests := make([]hexutil.Bytes, len(event.ExecRequests))
+		for i := range event.ExecRequests {
+			hexRequests[i] = hexutil.Bytes(event.ExecRequests[i])
```
marcello33, line 110: type conversion here is redundant

beacon/engine/types.go
```diff
@@ -336,7 +341,9 @@ func BlockToExecutableData(block *types.Block, fees *big.Int, sidecars []*types.
 		for j := range sidecar.Blobs {
 			bundle.Blobs = append(bundle.Blobs, hexutil.Bytes(sidecar.Blobs[j][:]))
 			bundle.Commitments = append(bundle.Commitments, hexutil.Bytes(sidecar.Commitments[j][:]))
-			bundle.Proofs = append(bundle.Proofs, hexutil.Bytes(sidecar.Proofs[j][:]))
+		}
+		for _, proof := range sidecar.Proofs {
+			bundle.Proofs = append(bundle.Proofs, hexutil.Bytes(proof[:]))
```
marcello33, line 346: same here with redundant conversions for Blobs and Proofs

core/rawdb/eradb/eradb.go
```diff
@@ -0,0 +1,345 @@
...
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// Package eradb implements a history backend using era1 files.
```
marcello33, line 17: Just wondering how this will be used. Does it show improvements in terms of efficiency in accessing db files?

core/rawdb/accessors_indexes.go
```diff
@@ -189,6 +191,187 @@ func ReadReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig)
...
+		log.Error("Transaction not found", "number", *blockNumber, "hash", blockHash, "txhash", hash, "err", err)
+		return nil, common.Hash{}, 0, 0
+	}
+	return tx, blockHash, *blockNumber, txIndex
+}
+
+// ReadCanonicalReceipt retrieves a specific transaction receipt from the database,
+// along with its added positional metadata. Notably, only the receipt in the canonical
+// chain is visible.
+func ReadCanonicalReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig) (*types.Receipt, common.Hash, uint64, uint64) {
```
marcello33, line 263: Isn't this function exactly the same as `ReadReceipt` defined before? Also, I see it's only used in tests. Maybe we don't need it?

core/state/access_events.go
```diff
@@ -92,127 +93,156 @@ func (ae *AccessEvents) Copy() *AccessEvents {
...
+	if chunkRead {
+		gas += params.WitnessChunkReadCost
+	}
+	if branchWrite {
+		gas += params.WitnessBranchWriteCost
+	}
+	if chunkWrite {
+		gas += params.WitnessChunkWriteCost
+	}
+	if chunkFill {
```
marcello33, line 222: Isn't this condition always false?

core/types/transaction.go
```diff
@@ -370,44 +370,53 @@ func (tx *Transaction) GasTipCapIntCmp(other *big.Int) int {
...
 
 // EffectiveGasTipCmp compares the effective gasTipCap of two transactions assuming the given base fee.
 func (tx *Transaction) EffectiveGasTipCmp(other *Transaction, baseFee *big.Int) int {
 	if baseFee == nil {
 		return tx.GasTipCapCmp(other)
 	}
-	return tx.EffectiveGasTipValue(baseFee).Cmp(other.EffectiveGasTipValue(baseFee))
+	// Use more efficient internal method.
+	txTip, otherTip := new(big.Int), new(big.Int)
+	tx.calcEffectiveGasTip(txTip, baseFee)
```
marcello33, line 407: what is `calcEffectiveGasTip` returns an error? Shall we handle it here and in other places?

core/rawdb/accessors_indexes.go
avalkov: Its really not used and doesnt seems to be used in go-ethereum v1.16.1, but for easier resolving of conflicts, I think we can just keep it and delete it with upstream merge when they delete it.

core/state/access_events.go
avalkov: It is, but I guess its preparation for some upcoming features for the witnesses, I think better to keep it as it is, because it will be one more conflict on next upstream merge.

core/rawdb/eradb/eradb.go
avalkov: If you prune your node from historical data but then you want to access this historical data for some reason, you can download it as era1 files https://geth.ethereum.org/docs/fundamentals/downloadera 

I think we wont be using this feature atleast for now @manav2401 @pratikspatil024 ?

core/types/transaction.go
avalkov: Thats how it is in upstream, maybe we can add panic here if error is returned ?

beacon/blsync/engineclient.go
avalkov: It really redundant, but thats how it is in upstream, I suggest to keep it like this to avoid future conflicts.

beacon/engine/types.go
avalkov: It really redundant, but thats how it is in upstream, I suggest to keep it like this to avoid future conflicts.

core/types/transaction.go
marcello33: Not sure I'd drift from upstream here. Resolving for now.

core/blockchain.go
avalkov: So no change if I understand correctly ?

core/blockchain.go
avalkov: Fixed it.

go.mod
avalkov: Done.

core/blockchain.go
cffls: Correct

manav2401 (approved):

pratikspatil024 (approved):

PR #1699, opened 2025-08-13
core: Key-Value Priority Read on BorReceipts

lucca30:
# Description

This PR aims to prioritize Key Value Reads on Bor Receitps over the ones from Ancient db.

# Changes

# Breaking changes

No breaks

### DIFF
--- a/core/rawdb/bor_receipt.go
+++ b/core/rawdb/bor_receipt.go
@@ -36,21 +36,22 @@ func ReadBorReceiptRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.Raw
 
 	// First, try to fetch from KV db
 	data, _ = db.Get(borReceiptKey(number, hash))
+	if data != nil {
+		return data
+	}
 
-	if data == nil {
-		err := db.ReadAncients(func(reader ethdb.AncientReaderOp) error {
-			// Check if the data is in ancients
-			if isCanon(reader, number, hash) {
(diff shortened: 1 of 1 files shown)


core/rawdb/bor_receipt.go
```diff
@@ -34,22 +34,23 @@ func borTxLookupKey(hash common.Hash) []byte {
 func ReadBorReceiptRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValue {
 	var data []byte
 
-	err := db.ReadAncients(func(reader ethdb.AncientReaderOp) error {
-		// Check if the data is in ancients
-		if isCanon(reader, number, hash) {
-			data, _ = reader.Ancient(freezerBorReceiptTable, number)
+	// First, try to fetch from KV db
+	data, _ = db.Get(borReceiptKey(number, hash))
```
cffls, line 38: If returning here when data isn't empty, we can probably avoid wrapping ancient reading in an if statement.

```go
if data != nil {
    return data
}
```

lucca30: The `if data == nil` already does that by only wrapping ancient read when no data is retrieved

lucca30: done, thanks

cffls (approved):
lgtm, thanks!

kamuikatsurgi (approved):

PR #1700, opened 2025-08-13
miner: Veblop rotation timer

cffls:
# Description

Use a separate timer for veblop span rotation instead of relying on recommit.

# Changes

## Testing

- [x] I have added unit tests
- [x] I have added tests to CI

### DIFF
--- a/integration-tests/kurtosis_stateless_test.sh
+++ b/integration-tests/kurtosis_stateless_test.sh
@@ -412,7 +412,7 @@ test_polycli_load_with_rotation() {
 
 	# Record initial state
 	test_account="0x97538585a02A3f1B1297EB9979cE1b34ff953f1E"
-	num_txs=6000  # More transactions for longer test
+	num_txs=6000 # More transactions for longer test
 	initial_nonce=$(cast nonce "$test_account" --rpc-url "$first_rpc_url")
 
 	echo "Initial account nonce: $initial_nonce"
@@ -428,7 +428,7 @@ test_polycli_load_with_rotation() {
 		--requests $num_txs \
 		--rate-limit 100 \
 		--mode t \
-		--gas-price 35000000000 > /tmp/polycli_rotation_test.log 2>&1 &
+		--gas-price 35000000000 >/tmp/polycli_rotation_test.log 2>&1 &
 
 	LOAD_PID=$!
 	echo "Load test started with PID: $LOAD_PID"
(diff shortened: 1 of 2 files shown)


integration-tests/kurtosis_stateless_test.sh
```diff
@@ -393,6 +393,135 @@ test_polycli_load_test() {
 	fi
 }
 
+# Test 8: Combined load test with block producer rotation
```
kamuikatsurgi, line 396: nit: can you just run `shfmt -w integration-tests/*.sh` to format if needed?

internal/cli/server/config.go
```diff
@@ -726,7 +726,7 @@ func DefaultConfig() *Config {
 			GasCeil:             miner.DefaultConfig.GasCeil,
 			GasPrice:            big.NewInt(params.BorDefaultMinerGasPrice), // bor's default
 			ExtraData:           "",
-			Recommit:            1 * time.Second,
+			Recommit:            125 * time.Second,
```
kamuikatsurgi, line 729: Any specific reason for `125 seconds`?

miner/worker.go
```diff
@@ -1519,7 +1541,9 @@ func (w *worker) commitWork(interrupt *atomic.Int32, noempty bool, timestamp int
 
 	// Create an empty block based on temporary copied state for
 	// sealing in advance without waiting block execution finished.
-	if !noempty && !w.noempty.Load() {
+	// If the block is a veblop block, we will never create an empty block
```
kamuikatsurgi: Do you know if this assumption holds every time? Even if there are no transactions in the transaction pool, like in Amoy?

kamuikatsurgi: If yes, can you remove the above comments?
```
// Create an empty block based on temporary copied state for
// sealing in advance without waiting block execution finished.
```

internal/cli/server/config.go
cffls: 125 seconds was the previous value. This change will revert it back.

miner/worker.go
cffls: This comment might be a bit confusing. When creating a block, worker will create two commits, one is simply empty, and the other one will try to include transactions. In veblop, we don't want to create the first commit where block is empty, but it doesn't mean the final block will always be non-empty.

integration-tests/kurtosis_stateless_test.sh
cffls: Done

miner/worker.go
kamuikatsurgi: Ohh okay, got it.

kamuikatsurgi (approved):

PR #1710, opened 2025-08-18
eth, core: handle state-sync data in eth/69 messages

manav2401:
# Description

This PR adds the ability to send state-sync transaction receipts (or bor receipts in short) in `eth/69` protocol (which is newly added from upstream). This will allow nodes using snap sync to persist the state-sync transaction receipts (as there's no execution involved in snap sync and hence they're not generated locally).

This PR changes the implementation of 2 p2p messages:
- GetReceiptsMsg (code: 0x0f): Fetches bor receipts for requested blocks and include them in the final encoded message
- ReceiptsMsg (code: 0x10): Accepts the bor receipts from incoming p2p packet. They are excluded from receipt root (in header) and hence this needs to be taken care of while handling incoming receipts. 

Moreover, this PR also changes how receipts are handled in `InsertReceiptChain` function which didn't handle bor receipts earlier. Bor receipts can be identified by their 0 gas used field. Upon insertion, as they're stored separately from normal block receipts, they need to be extracted from the encoded packet which is handled in this PR.

Note that this change is forward compatible and allows to do a hard fork in future where we include bor receipts as a part of receipt root without any p2p changes.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


core/blockchain.go
```diff
@@ -1620,6 +1620,41 @@ const (
...
+		// If no receipts left, return
+		if len(decoded[:len(decoded)-1]) == 0 {
+			return nil, encodedStateSyncReceipt
+		}
+
+		// Encode back the normal (non state-sync) receipts and return
+		encodedReceipts, err := rlp.EncodeToBytes(decoded[:len(decoded)-1])
+		if err != nil {
+			log.Warn("Failed to encode remaining receipts after excluding state-sync receipt", "number", number, "hash", hash, "err", err)
+			return receipts, encodedStateSyncReceipt
```
lucca30, line 1746: On this function, wouldn't it be better retrieving the error so you know how to properly handle it?

I see it's called once at:

```
			receiptChain[i], borReceipts[i] = splitReceipts(receipts, blockChain[i].NumberU64(), blockChain[i].Hash())
```

And even if this return scenario I pointed is unlikely to happen, it seem it would not behave good receiving this returned value because it will returns the `receipts` with the `encodedStateSyncReceipt` in it.

manav2401: I actually thought about those cases when encoding/decoding fails and I wasn't sure if we should return an error or not. There are 2 broad options, either to return error which would reject the whole receipt chain import or accept the error with the possibility of storing some data differently (note that it's not corrupted data -- it's either missing data or duplicate data). 

I am hoping to conduct some e2e tests and devnet testing and see what's the frequency of these errors. If it's low, I can return an error from here so that it fails for crucial cases. 

Re. the values being returned, I think it's valid instead of returning nil. This case is when we have extracted and encoded the state sync receipts separately and we're trying to re-encode the remaining receipts (as they're stored separately as of now). If this re-encoding fails, we return the original receipts list back (which preserves the normal receipts + has state-sync receipt as well) and we also return the encoded state-sync receipt. Hence, the state-sync receipt is stored perfectly fine and the normal receipts has an extra state-sync receipt (which is duplicate) but we're fine with that (instead of returning nil and storing nothing). Wdyt?

core/bor_blockchain_reader.go
```diff
@@ -4,14 +4,24 @@ import (
...
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/metrics"
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
+var (
+	borReceiptsCacheHit     = metrics.NewRegisteredGauge("bor/receipts/cache/hit", nil)
+	borReceiptsCacheMiss    = metrics.NewRegisteredGauge("bor/receipts/cache/miss", nil)
+	borReceiptsRLPCacheHit  = metrics.NewRegisteredGauge("bor/rlpreceipts/cache/hit", nil)
+	borReceiptsRLPCacheMiss = metrics.NewRegisteredGauge("bor/rlpreceipts/cache/miss", nil)
```
manav2401, line 15: These metrics have been added to see the hit/miss ratio. For normal receipts, they've been removed from upstream. If the ratio is bad, probably worth removing for bor receipts as well.

core/blockchain.go
lucca30: Yeah, I agree with the approach. Specially because I think there is no possible scenario to reach this failure statement. So lgtm

lucca30 (approved):

eth/downloader/queue.go
```diff
@@ -86,7 +86,11 @@ func newFetchResult(header *types.Header, syncMode SyncMode) *fetchResult {
 		item.Withdrawals = make(types.Withdrawals, 0)
 	}
 
-	if syncMode == SnapSync && !header.EmptyReceipts() {
+	fetchReceipts := true
+	if header.EmptyReceipts() && !isSprintEndBlock(borCfg, header.Number.Uint64()) {
+		fetchReceipts = false
+	}
+	if syncMode == SnapSync && fetchReceipts {
```
manav2401, line 93: Will be writing a new test around this change.

core/blockchain.go
```diff
@@ -1659,6 +1659,45 @@ const (
 	SideStatTy
 )
 
+// splitReceipts separates out the state-sync receipt from the whole receipt list
+// of a block and returns the encoded lists back separately. In case of errors or
+// empty receipt, it returns `nil` instead of `rlp.EncodeToBytes(nil)`.
+func splitReceipts(receipts rlp.RawValue, number uint64, hash common.Hash) (rlp.RawValue, rlp.RawValue) {
```
cffls: Can we optimize this by only decoding receipt if they are from a sprint end block? It would reduce unnecessary cpu cycles.

core/blockchain.go
```diff
@@ -1659,6 +1659,45 @@ const (
...
+	}
+
+	var decoded []*types.ReceiptForStorage
+	if err := rlp.DecodeBytes(receipts, &decoded); err != nil {
+		log.Warn("Failed to decode block receipts", "number", number, "hash", hash, "err", err)
+		return receipts, nil
+	}
+	// Find if there's a state-sync transaction receipt present. They are always
+	// appended at the end of list and can be identified by 0 gas usage.
+	if len(decoded) > 0 && decoded[len(decoded)-1].CumulativeGasUsed == 0 {
```
cffls: Could you help me understand why `CumulativeGasUsed` is used to determine whether it is a state sync receipt instead of `GasUsed`?

eth/protocols/eth/receipt.go
```diff
@@ -337,6 +337,15 @@ func (rl *ReceiptList68) EncodeRLP(_w io.Writer) error {
 	return w.Flush()
 }
 
+// ExcludeStateSync removes the state sync transaction from the list. We use
+// the property of zero gas used to identify a state-sync transaction in the
+// list.
+func (rl *ReceiptList68) ExcludeStateSync() {
+	if len(rl.items) > 0 && rl.items[len(rl.items)-1].GasUsed == 0 {
```
cffls: Here `GasUsed` is used, which isn't consistent with other places where `CumulativeGasUsed`.

eth/protocols/eth/receipt.go
```diff
@@ -403,10 +412,20 @@ func (rl *ReceiptList69) EncodeRLP(_w io.Writer) error {
 	return w.Flush()
 }
 
+// ExcludeStateSync removes the state sync transaction from the list. We use
+// the property of zero gas used to identify a state-sync transaction in the
+// list.
+func (rl *ReceiptList69) ExcludeStateSync() {
+	if len(rl.items) > 0 && rl.items[len(rl.items)-1].GasUsed == 0 {
```
cffls: Since this is the same implementation as ReceiptList68, it could be extracted as a common function and used by both ReceiptList68 and ReceiptList69.

eth/protocols/eth/receipt.go
```diff
@@ -421,10 +440,14 @@ func blockReceiptsToNetwork69(blockReceipts, blockBody rlp.RawValue) ([]byte, er
 	)
 	outer := enc.List()
 	for i := 0; it.Next(); i++ {
-		txType, _ := nextTxType()
 		content, _, _ := rlp.SplitList(it.Value())
 		receiptList := enc.List()
-		enc.WriteUint64(uint64(txType))
+		if isStateSyncReceipt(i) {
+			enc.WriteUint64(uint64(0)) // TxType is always 0 for state-sync transactions
```
cffls, line 460: It seems incompatible with [PIP-74](https://github.com/0xPolygon/Polygon-Improvement-Proposals/pull/509/files#diff-e65b1d8c93a5833fe6e4d0823528eb7664568c550ab1e9d35d1af0f713685cc1R39). I suppose this will change after a HF?

manav2401: Yes, based on whatever tx type we decide to have for state-sync txs, this will have a hard fork check.

core/blockchain.go
manav2401: Sure, done!

core/blockchain.go
marcello33: I am using GasUsed in Erigon. Let's agree on what's best here, thanks

manav2401: Sure. So bor uses the struct called `ReceiptForStorage` to store the raw receipts into db (after rlp encoding them). If you see the encoding / decoding functions [here](https://github.com/0xPolygon/bor/blob/c6944926972edbc451b17c05e6db2cd4184dab14/core/types/receipt.go#L340), you'll see that it always reads and writes `CumulativeGasUsed` into the database. Even the original receipt struct defined [here](https://github.com/0xPolygon/bor/blob/c6944926972edbc451b17c05e6db2cd4184dab14/core/types/receipt.go#L53) has `CumulativeGasUsed` as a primary field and fields like `GasUsed` are derived on the fly (see [this](https://github.com/0xPolygon/bor/blob/c6944926972edbc451b17c05e6db2cd4184dab14/core/types/receipt.go#L408) derive fields function). 

If you check the [p2p spec](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#receipt-encoding-and-validity), that also involves cumulative gas instead of just gas used. 

This field is zero for bor receipts (which is wrong IMO but that's a separate discussion for later). For now, we can identify bor receipts via this.

eth/protocols/eth/receipt.go
manav2401: This is a bummer because the it represents cumulative gas used and not gas used. It's just that the struct being used in p2p has named the field `GasUsed` instead of `CumulativeGasUsed` and I haven't touched this as it's coming from geth. As mentioned in the comment above, the p2p spec also uses cumulative gas.

marcello33 (approved):

eth/protocols/eth/receipt.go
manav2401: Done. There were few more extra changes needed which are also extracted to a common function.

core/blockchain.go
manav2401: Hey @marcello33, so I've updated the logic in bor to identify a state-sync receipt. 
- We first check if the cumulative gas used is 0. If yes, it's a state-sync receipt. 
- If not, we check if the cumulative gas used of last receipt is same as that of second last receipt (indicating a state-sync receipt with 0 gas used). 
- In every other case, there's no state-sync transaction. 

Maybe worth ensuring this logic in erigon as well.

marcello33: Ok, thanks @manav2401 
Just included the changes

cffls: Makes sense. Thanks for the clarification.

cffls (approved):
All e2e tests are passing now. LGTM!

PR #1594, opened 2025-06-26
Add-in BalanceChangePolygonBurn hook

Rampex1:
# Description

Added new hook to replicate BalanceChange_REASON_BURN behaviour in polygon-1.x

# Changes

- [X] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)
- [ ] Changes only for a subset of nodes

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Nodes audience

In case this PR includes changes that must be applied only to a subset of nodes, please specify how you handled it (e.g. by adding a flag with a default value...)

# Checklist

- [ ] I have added at least 2 reviewer or the whole pos-v1 team
- [ ] I have added sufficient documentation in code
- [ ] I will be resolving comments - if any - by pushing each fix in a separate commit and linking the commit hash in the comment reply
- [ ] Created a task in Jira and informed the team for implementation in Erigon client (if applicable)
- [ ] Includes RPC methods changes, and the Notion documentation has been updated

# Cross repository changes

- [ ] This PR requires changes to heimdall
    - In case link the PR here:
- [ ] This PR requires changes to matic-cli
    - In case link the PR here:

## Testing

- [ ] I have added unit tests
- [ ] I have added tests to CI
- [ ] I have tested this code manually on local environment
- [ ] I have tested this code manually on remote devnet using express-cli
- [ ] I have tested this code manually on amoy
- [ ] I have created new e2e tests into express-cli

### Manual tests

Please complete this section with the steps you performed if you ran manual tests for this functionality, otherwise delete it

# Additional comments

Please post additional comments in this section if you have them, otherwise delete it

PR #1711, opened 2025-08-19
Live tracer support

nebojsa94:
# Description

Adds Live Tracer support for  `bor server`

# Changes

- [X] New feature (non-breaking change that adds functionality)

# Checklist

- [X] I will be resolving comments - if any - by pushing each fix in a separate commit and linking the commit hash in the comment reply

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -841,6 +841,8 @@ func DefaultConfig() *Config {
 			LogNoHistory:       ethconfig.Defaults.LogNoHistory,
 			StateHistory:       params.FullImmutabilityThreshold,
 		},
+		VMTrace:           "",
+		VMTraceJsonConfig: "{}",
 	}
 }
 
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -146,6 +146,12 @@ type Config struct {
 
 	// HistoryConfig has historical data retention related settings
 	History *HistoryConfig `hcl:"history,block" toml:"history,block"`
+
+	// VMTrace Name of tracer which should record internal VM operations (costly)
+	VMTrace string `hcl:"vmtrace,optional" toml:"vmtrace,optional"`
+
+	// VMTraceJsonConfig Tracer configuration (JSON)
+	VMTraceJsonConfig string `hcl:"vmtrace.jsonconfig,optional" toml:"vmtrace.jsonconfig,optional"`
 }
 
 type HistoryConfig struct {
@@ -1256,6 +1262,8 @@ func (c *Config) buildEth(stack *node.Node, accountManager *accounts.Manager) (*
 	}
 
 	n.EnableBlockTracking = c.Logging.EnableBlockTracking
+	n.VMTrace = c.VMTrace
+	n.VMTraceJsonConfig = c.VMTraceJsonConfig
 
 	return &n, nil
 }
--- a/internal/cli/server/flags.go
+++ b/internal/cli/server/flags.go
@@ -131,6 +131,20 @@ func (c *Command) Flags(config *Config) *flagset.Flagset {
 		Default: c.cliConfig.Logging.Vmodule,
 		Group:   "Logging",
 	})
+	f.StringFlag(&flagset.StringFlag{
+		Name:    "vmtrace",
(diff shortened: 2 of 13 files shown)


internal/cli/server/config.go
```diff
@@ -146,6 +146,12 @@ type Config struct {
 
 	// HistoryConfig has historical data retention related settings
 	History *HistoryConfig `hcl:"history,block" toml:"history,block"`
+
+	// VMTrace Name of tracer which should record internal VM operations (costly)
+	VMTrace string `hcl:"vmtrace,optional" toml:"vmtrace,optional"`
```
kamuikatsurgi, line 151: Can you set default values of these in the `DefaultConfig()` func?

internal/cli/server/flags.go
```diff
@@ -131,6 +131,20 @@ func (c *Command) Flags(config *Config) *flagset.Flagset {
 		Default: c.cliConfig.Logging.Vmodule,
 		Group:   "Logging",
 	})
+	f.StringFlag(&flagset.StringFlag{
```
kamuikatsurgi, line 134: Can you add some examples in the docs on how to use this, and also run `make docs` afterwards?

nebojsa94:
@kamuikatsurgi resolved comments, can you take a look?

internal/cli/server/config.go
```diff
@@ -1256,6 +1264,8 @@ func (c *Config) buildEth(stack *node.Node, accountManager *accounts.Manager) (*
 	}
 
 	n.EnableBlockTracking = c.Logging.EnableBlockTracking
+	n.VMTrace = c.VMTrace
```
pratikspatil024, line 1267: Better to assign only if value is not empty string.
```
	if c.VMTrace != nil {
		n.VMTrace = c.VMTrace
		n.VMTraceJsonConfig = c.VMTraceJsonConfig
	}
```

internal/cli/server/config.go
```diff
@@ -146,6 +146,12 @@ type Config struct {
 
 	// HistoryConfig has historical data retention related settings
 	History *HistoryConfig `hcl:"history,block" toml:"history,block"`
+
+	// VMTrace Name of tracer which should record internal VM operations (costly)
```
pratikspatil024, line 150: These two fields are added in the `Config` structure, and I see that you have added the flags in the `[log]` section of the config file. This will not work.

I would recommend creating another structure/section for this, or move them to the one which is most relevant.

pratikspatil024 (approved):

nebojsa94:
I just realised that we would need to pass tracingEVM in Finalise() function of StateDB interface in order to have tracer context

maoueh:
@nebojsa94 You can retrieve the live tracer via `chain`, in the `Finalize method:

```
tracer := chain.(*core.BlockChain).GetTracingHooks()
```

This is correct has the Finalize itself make that assumption that it's of type `(*core.BlockChain)` due to https://github.com/0xPolygon/bor/blob/master/consensus/bor/bor.go#L924-L926 and https://github.com/0xPolygon/bor/blob/master/core/state_processor.go#L147-L150.

On the actual Live tracing, here stuff that we had to add/tweak:

What is missing is the actual instrumentation of `ApplyMessage` at https://github.com/0xPolygon/bor/blob/master/consensus/bor/statefull/processor.go#L69 which does some state changes. You can see how current implementation at:
 - https://github.com/streamingfast/go-ethereum/blob/release/polygon-2.x-fh3.0/consensus/bor/statefull/processor.go#L139-L147
  - https://github.com/streamingfast/go-ethereum/blob/release/polygon-2.x-fh3.0/consensus/bor/statefull/processor.go#L182-L200

There is also https://github.com/0xPolygon/bor/blob/master/consensus/bor/bor.go#L957-L958 that is unspecified but shouldn't, a specific Polygon balance change should be introduced here IMO. 

There is also a fee Transfers that IMO are not actual transfer see https://github.com/0xPolygon/bor/pull/1594/files, would like to reach consensus about the right values here.

PR #1724, opened 2025-08-27
Remove explicit wait for span rotation

cffls:
# Description

This change will remove a explicit wait for span changes if an incoming block has a different author from its parent block. Previously, this wait was added to prevent the case where a span rotation happens but local heimdall hasn't received the new span.

This waiting mechanism is now replaced by a check that will make sure bor will be blocked until heimdall has synced to the tip. The same check will also resolve an issue where bor might retrieve incorrect state sync events when heimdall isn't in sync.

# Changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [x] I have added unit tests

### DIFF
--- a/consensus/bor/heimdallgrpc/client.go
+++ b/consensus/bor/heimdallgrpc/client.go
@@ -1,6 +1,7 @@
 package heimdallgrpc
 
 import (
+	"context"
 	"strings"
 	"time"
 
@@ -8,7 +9,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 
-	protoV1 "github.com/0xPolygon/polyproto/heimdall"
+	"github.com/ethereum/go-ethereum/consensus/bor/heimdall"
 	"github.com/ethereum/go-ethereum/log"
 	grpcRetry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
 
@@ -16,6 +17,7 @@ import (
 	checkpointTypes "github.com/0xPolygon/heimdall-v2/x/checkpoint/types"
 	clerkTypes "github.com/0xPolygon/heimdall-v2/x/clerk/types"
 	milestoneTypes "github.com/0xPolygon/heimdall-v2/x/milestone/types"
+	ctypes "github.com/cometbft/cometbft/rpc/core/types"
 )
 
 const (
@@ -25,15 +27,15 @@ const (
 
 type HeimdallGRPCClient struct {
 	conn                  *grpc.ClientConn
-	client                protoV1.HeimdallClient
+	client                *heimdall.HeimdallClient
 	borQueryClient        borTypes.QueryClient
 	checkpointQueryClient checkpointTypes.QueryClient
 	clerkQueryClient      clerkTypes.QueryClient
 	milestoneQueryClient  milestoneTypes.QueryClient
 }
 
-func NewHeimdallGRPCClient(address string) *HeimdallGRPCClient {
-	address = removePrefix(address)
+func NewHeimdallGRPCClient(grpcAddress string, heimdallURL string, timeout time.Duration) *HeimdallGRPCClient {
+	grpcAddress = removePrefix(grpcAddress)
 
 	opts := []grpcRetry.CallOption{
 		grpcRetry.WithMax(10000),
@@ -41,7 +43,7 @@ func NewHeimdallGRPCClient(address string) *HeimdallGRPCClient {
 		grpcRetry.WithCodes(codes.Internal, codes.Unavailable, codes.Aborted, codes.NotFound),
 	}
 
-	conn, err := grpc.NewClient(address,
+	conn, err := grpc.NewClient(grpcAddress,
 		grpc.WithStreamInterceptor(grpcRetry.StreamClientInterceptor(opts...)),
 		grpc.WithUnaryInterceptor(grpcRetry.UnaryClientInterceptor(opts...)),
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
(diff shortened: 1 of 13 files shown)


consensus/bor/heimdallgrpc/client.go
```diff
@@ -50,11 +52,11 @@ func NewHeimdallGRPCClient(address string) *HeimdallGRPCClient {
 		log.Crit("Failed to connect to Heimdall gRPC", "error", err)
 	}
 
-	log.Info("Connected to Heimdall gRPC server", "address", address)
+	log.Info("Connected to Heimdall gRPC server", "grpcAddress", grpcAddress)
 
 	return &HeimdallGRPCClient{
 		conn:                  conn,
-		client:                protoV1.NewHeimdallClient(conn),
+		client:                heimdall.NewHeimdallClient(heimdallURL, timeout),
```
cffls, line 59: Created a heimdall http client in grpc client here because there isn't a corresponding grpc api for status in heimdall, which is probably also an overkill to add. 
There isn't really any strong reason why we shouldn't create a heimdall http client in grpc. The main reason why we created grpc heimdall client in the first place was to enable heimdall to run as part of bor, but that is disabled.

consensus/bor/span_store.go
```diff
@@ -85,6 +92,45 @@ func (s *SpanStore) getLatestSpan(ctx context.Context) (*borTypes.Span, error) {
...
+		if err != nil {
+			return err
+		}
+	}
+	s.heimdallStatus.Store(syncInfo)
+	return nil
+}
+
+func (s *SpanStore) waitUntilHeimdallIsSynced(ctx context.Context) {
+	// If there's no heimdall client, don't wait
```
lucca30, line 120: If there's no heimdall client, shouldn't we fail or panic?

consensus/bor/span_store.go
```diff
@@ -173,6 +219,8 @@ func (s *SpanStore) spanById(ctx context.Context, spanId uint64) (*borTypes.Span
 // asked for a future span. This is safe to assume as we don't have a way to find out span id for a future block
 // unless we hardcode the span length (which we don't want to).
 func (s *SpanStore) spanByBlockNumber(ctx context.Context, blockNumber uint64) (res *borTypes.Span, err error) {
+	s.waitUntilHeimdallIsSynced(ctx)
```
lucca30, line 237: Shouldn't we include a similar wait call before StateSync fecthing to also ensure it's synced? I saw you mentioned that it do also fixes the state sync fetching but I was supposing it demands a similar call on `Finalize` and `FinalizeAndAssemble`  in `bor.go`

lucca30 (commented):
Huge delivery! Just a few comments left

consensus/bor/span_store.go
cffls: We don't want to fail or panic here because this will happen in unit tests or when dev mode is enabled.

consensus/bor/span_store.go
cffls: Good point! Previously I thought this header validation would block `Finalize` until heimdall is ready, but I realized that header validation and `Finalize` are running asynchronously. Will add the check to `Finalize` and `FinalizeAndAssemble` as well. Thanks!

kamuikatsurgi:
Kurtosis now supports this feature, where we can stop Heimdall, let the bor run, see what happens with these PR changes, and start Heimdall after some time. For ref https://0xpolygon.github.io/kurtosis-pos/guides/restart-clients-without-stopping-containers.

avalkov (approved):

consensus/bor/span_store.go
```diff
@@ -85,6 +92,45 @@ func (s *SpanStore) getLatestSpan(ctx context.Context) (*borTypes.Span, error) {
...
+
+	for {
+		syncInfo := s.heimdallStatus.Load()
+		if syncInfo == nil || syncInfo.CatchingUp {
+			log.Warn("Heimdall isn't synced, waiting for update", "syncInfo", syncInfo)
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(timeout):
+				s.updateHeimdallStatus(ctx)
```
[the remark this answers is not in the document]
kamuikatsurgi: It would be nice to have this one included, @cffls.

consensus/bor/heimdallgrpc/client.go
Raneet10: > The main reason why we created grpc heimdall client in the first place was to enable heimdall to run as part of bor, but that is disabled.

I think heimdall running as part of bor is different; we just spin the heimdall binary when starting up bor. IIRC , adding grpc was to enable faster communication between the services.

consensus/bor/span_store.go
```diff
@@ -85,6 +92,45 @@ func (s *SpanStore) getLatestSpan(ctx context.Context) (*borTypes.Span, error) {
 	return s.latestSpanCache.Load(), nil
 }
 
+func (s *SpanStore) updateHeimdallStatus(ctx context.Context) (err error) {
+	var syncInfo *ctypes.SyncInfo
+	if s.heimdallClient == nil {
```
Raneet10, line 107: This check looks redundant since we only call this function when `s.heimdallClient != nil` ?

cffls: I added this additional check just in case the behavior of its caller changes. Didn't want to make too strong assumption about how it would be used.

consensus/bor/span_store.go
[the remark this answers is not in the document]
cffls: I am not sure if returning an error here is the best idea. If heimdall fails, we do want to keep waiting indefinitely until it is back up. Returning an error will result in bor failing the block execution and therefore dropping all good peers. There is already a background process that checks and reports the health of heimdall above, so an additional logging won't be necessary.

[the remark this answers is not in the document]
cffls: @kamuikatsurgi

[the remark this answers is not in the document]
kamuikatsurgi: Okay, works.

consensus/bor/heimdallgrpc/client.go
cffls: Thanks for the context. I saw the PR of GRPC also introduced the feature of running heimdall in bor. Anyway, if it is for the reason of faster communication, it won't be a concern in this use case.

PR #1726, opened 2025-08-27
StateSync Tx on Body

lucca30:
# Description

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2061,7 +2061,7 @@ func (bc *BlockChain) writeBlockAndSetHead(block *types.Block, receipts []*types
 		if emitHeadEvent {
 			bc.chainHeadFeed.Send(ChainHeadEvent{Header: block.Header()})
 			// BOR state sync feed related changes
-			for _, data := range bc.stateSyncData {
+			for _, data := range bc.GetStateSync() {
 				bc.stateSyncFeed.Send(StateSyncEvent{Data: data})
 			}
 			// BOR
@@ -2466,7 +2466,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool, makeWitness 
 		}
 
 		// BOR state sync feed related changes
-		for _, data := range bc.stateSyncData {
+		for _, data := range bc.GetStateSync() {
 			bc.stateSyncFeed.Send(StateSyncEvent{Data: data})
 		}
 		// BOR
--- a/core/parallel_state_processor.go
+++ b/core/parallel_state_processor.go
@@ -319,6 +319,9 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 	}
 	// Iterate over and process the individual transactions
 	for i, tx := range block.Transactions() {
+		if tx.Type() == types.StateSyncTxType {
+			continue
+		}
 		msg, err := TransactionToMessage(tx, types.MakeSigner(p.config, header.Number, header.Time), header.BaseFee)
 		if err != nil {
 			log.Error("error creating message", "err", err)
@@ -408,12 +411,12 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 	var requests [][]byte
 
 	// Finalize the block, applying any consensus engine specific extras (e.g. block rewards)
-	p.engine.Finalize(p.bc, header, statedb, block.Body())
+	receipts = p.engine.Finalize(p.bc, header, statedb, block.Body(), receipts)
 
 	return &ProcessResult{
 		Receipts: receipts,
 		Requests: requests,
-		Logs:     allLogs,
+		Logs:     statedb.Logs(),
 		GasUsed:  *usedGas,
 	}, nil
 }
--- a/core/types/tx_state_sync.go
+++ b/core/types/tx_state_sync.go
@@ -1,0 +1,118 @@
+// Copyright 2021 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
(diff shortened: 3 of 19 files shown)


core/blockchain.go
```diff
@@ -1935,7 +1935,8 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 			return blockLogs[i].Index < blockLogs[j].Index
 		})
 
-		if len(blockLogs) > len(logs) {
+		// After StateSync HF we dont write bor receipts separately
+		if !(bc.chainConfig.Bor != nil && bc.chainConfig.Bor.IsStateSync(block.Number())) && len(blockLogs) > len(logs) {
```
manav2401: We can do this check earlier to avoid sorting logs just above this.

core/parallel_state_processor.go
```diff
@@ -329,6 +329,9 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 	}
 	// Iterate over and process the individual transactions
 	for i, tx := range block.Transactions() {
+		if tx.Type() == types.StateSyncTxType {
```
manav2401, line 332: Just curious to know if this check works as expected or not. Because the transaction as part of block body is sent via p2p, I am not sure if the underlying type (i.e. StateSyncTxType) is maintained or not.

core/types/tx_state_sync.go
```diff
@@ -0,0 +1,113 @@
...
+		return errors.New("nil StateSyncTx")
+	}
+	enc := make([]StateSyncData, 0, len(tx.StateSyncData))
+	for _, d := range tx.StateSyncData {
+		if d == nil {
+			continue
+		}
+		enc = append(enc, *d)
+	}
+	return rlp.Encode(buf, enc)
```
manav2401, line 88: IIRC, you need an encoder function for type `StateSyncData` for `rlp.Encode` to work. Was that already implemented? Same for decode.

core/parallel_state_processor.go
manav2401: Oh actually I saw the encoding/decoding functions which appends type of tx as the first byte. LGTM in that case.

manav2401:
The core logic LGTM. There are a few places where we might have to make changes. 
1. There are certain RPC methods which has custom logic for handling bor events. After this PR, we will have to maintain 2 separate logic (the post HF one would be very simple). 
2. Right now we use a separate transaction lookup which maps a transaction hash to bor receipt. That logic won't be used after HF. We need to confirm if queries done via tx hash works correctly or not (e.g. getTransactionReceipt by transaction hash). Ideally they should as the lookup entries are updated by iterating over block.Transactions which already has state-sync transactions. 
3. There will be some issues w/ snap sync post this HF as my changes related to eth/69 are merged. For e.g. right now, we exclude state-sync receipt from receipt root calculation but after the HF, we should be including it. If you want, I can take a look into this as I am mostly familiar with where the changes would be. 
Thanks!

core/parallel_state_processor.go
lucca30: Yeah. It was a fix I specially did after running over the kurtosis environment.

core/types/tx_state_sync.go
lucca30: Not necessarily since the struct has just encodable structs. I double checked it on the unit tests for this file. If you see, I'm checking each value change impacting on the hash itself

lucca30:
Thanks for the comments @manav2401 

> 1. There are certain RPC methods which has custom logic for handling bor events. After this PR, we will have to maintain 2 separate logic (the post HF one would be very simple).

I think I already checked all of them and just `eth_getTransactionByHash` showed an issue but not directly related to it. But it's crucial checking all of them specially on e2e.

> 2. Right now we use a separate transaction lookup which maps a transaction hash to bor receipt. That logic won't be used after HF. We need to confirm if queries done via tx hash works correctly or not (e.g. getTransactionReceipt by transaction hash). Ideally they should as the lookup entries are updated by iterating over block.Transactions which already has state-sync transactions.
 Yeah, they not impact

> 3. There will be some issues w/ snap sync post this HF as my changes related to eth/69 are merged. For e.g. right now, we exclude state-sync receipt from receipt root calculation but after the HF, we should be including it. If you want, I can take a look into this as I am mostly familiar with where the changes would be.

Yes, it will demands a HF check. As soon as one of the PR's got merged we can fix on the other one.

core/blockchain.go
lucca30: Yeah, you are right. We can do this sort just after the check

lucca30: done

manav2401:
> The core logic LGTM. There are a few places where we might have to make changes.
> 
> 1. There are certain RPC methods which has custom logic for handling bor events. After this PR, we will have to maintain 2 separate logic (the post HF one would be very simple).
> 2. Right now we use a separate transaction lookup which maps a transaction hash to bor receipt. That logic won't be used after HF. We need to confirm if queries done via tx hash works correctly or not (e.g. getTransactionReceipt by transaction hash). Ideally they should as the lookup entries are updated by iterating over block.Transactions which already has state-sync transactions.
> 3. There will be some issues w/ snap sync post this HF as my changes related to eth/69 are merged. For e.g. right now, we exclude state-sync receipt from receipt root calculation but after the HF, we should be including it. If you want, I can take a look into this as I am mostly familiar with where the changes would be.
>    Thanks!

All 3 things seems to handled. First one is pending to be merged and needs some e2e and devnet testing.

PR #1739, opened 2025-08-29
Rename VEBloP HF to Rio

cffls:
Change the HF name from `veblop` to `Rio`.

### DIFF
--- a/consensus/bor/bor_test.go
+++ b/consensus/bor/bor_test.go
@@ -163,7 +163,7 @@ func TestEncodeSigHeaderJaipur(t *testing.T) {
 	require.Equal(t, hash, hashWithoutBaseFee)
 }
 
-func TestCalcProducerDelayVeBlop(t *testing.T) {
+func TestCalcProducerDelayRio(t *testing.T) {
 	t.Parallel()
 
 	// Test cases for VeBlop condition in CalcProducerDelay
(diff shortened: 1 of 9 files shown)


consensus/bor/bor_test.go
```diff
@@ -163,7 +163,7 @@ func TestEncodeSigHeaderJaipur(t *testing.T) {
 	require.Equal(t, hash, hashWithoutBaseFee)
 }
 
-func TestCalcProducerDelayVeBlop(t *testing.T) {
+func TestCalcProducerDelayRio(t *testing.T) {
 	t.Parallel()
 
 	// Test cases for VeBlop condition in CalcProducerDelay
```
manav2401, line 169: The comments and test case description still use veblop. Maybe it's okay as it helps to identify the technical aspect of the change but just pointing out.

manav2401 (approved):

kamuikatsurgi (approved):
LGTM, can you just run `go mod tidy` and include the changes in this PR only? It seems there are some changes in the `go.mod` file.

consensus/bor/bor_test.go
cffls: Thanks for pointing this out. Yes, keeping veblop in some functions was intentional. It will be easier to understand what the function actually does in certain context.

cffls:
> LGTM, can you just run go mod tidy and include the changes in this PR only? It seems there are some changes in the go.mod file.

I don't think go.mod file has changed. Maybe it was due to a cache in your browser?

kamuikatsurgi:
> > LGTM, can you just run go mod tidy and include the changes in this PR only? It seems there are some changes in the go.mod file.
> 
> I don't think go.mod file has changed. Maybe it was due to a cache in your browser?

I was referring to this https://github.com/0xPolygon/bor/pull/1740; my apologies if I wasn't clear.

PR #1743, opened 2025-09-03
Customizable coinbase address

cffls:
# Description

This PR enables coinbase to be a different address, which is specified in the genesis, from the block producer. The motivation is to redirect all tx fees in veblop era to a smart contract address, where the fee could be redistributed based on certain rules.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [x] I have added unit tests

### DIFF
--- a/core/evm.go
+++ b/core/evm.go
@@ -54,7 +54,16 @@ func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common
 
 	// If we don't have an explicit author (i.e. not mining), extract from the header
 	if author == nil {
-		beneficiary, _ = chain.Engine().Author(header) // Ignore error, we're past header validation
+		if chain.Config().Bor != nil && chain.Config().Bor.IsRio(header.Number) {
+			beneficiary = common.HexToAddress(chain.Config().Bor.CalculateCoinbase(header.Number.Uint64()))
+
+			// In case of coinbase is not set post Rio, use the default coinbase
+			if beneficiary == (common.Address{}) {
+				beneficiary, _ = chain.Engine().Author(header)
+			}
+		} else {
+			beneficiary, _ = chain.Engine().Author(header) // Ignore error, we're past header validation
+		}
 	} else {
 		beneficiary = *author
 	}
--- a/core/parallel_state_processor.go
+++ b/core/parallel_state_processor.go
@@ -290,8 +290,6 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 
 	shouldDelayFeeCal := true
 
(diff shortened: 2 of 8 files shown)


core/parallel_state_processor.go
```diff
@@ -290,8 +290,6 @@ func (p *ParallelStateProcessor) Process(block *types.Block, statedb *state.Stat
 
 	shouldDelayFeeCal := true
 
-	coinbase, _ := p.bc.Engine().Author(header)
```
cffls, line 293: This will break parallel process if the fee is sent to a smart contract after veblop. Replaced by `coinbase := blockContext.Coinbase` below.

internal/cli/server/chains/amoy.go
```diff
@@ -59,6 +59,10 @@ var amoyTestnet = &Chain{
 					"0":     "0x000000000000000000000000000000000000dead",
 					"73100": "0xeCDD77cE6f146cCf5dab707941d318Bd50eeD2C9",
 				},
+				Coinbase: map[string]string{
+					"0":        "0x0000000000000000000000000000000000000000",
+					"26201856": "0x000000000000000000000000000000000000ba5e",
```
cffls: `0x000000000000000000000000000000000000ba5e` is just a place hold. The final address is TBD.

core/evm.go
```diff
@@ -54,7 +54,16 @@ func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common
...
+			beneficiary = common.HexToAddress(chain.Config().Bor.CalculateCoinbase(header.Number.Uint64()))
+
+			// In case of coinbase is not set post Rio, use the default coinbase
+			if beneficiary == (common.Address{}) {
+				beneficiary, _ = chain.Engine().Author(header)
+			}
+		} else {
+			beneficiary, _ = chain.Engine().Author(header) // Ignore error, we're past header validation
+		}
 	} else {
```
kamuikatsurgi, line 67: I don't have the full context here, but we only want to set the beneficiary to coinbase if `author == nil`? As per the description, we want to do this for all the blocks. Please correct me if I'm wrong.

kamuikatsurgi: I am talking about this case:
```go
else {
	beneficiary = *author
}
```

cffls: That's correct. If the author is specified,  we will use the author passed by caller. Otherwise, there isn't a point to keep `author` in the argument of this function. I think the only case when author will be passed in is in mining and unit tests. For mining, the same logic has been implemented in `worker.go`.

Issue #1737, opened 2025-08-28
Data race on SpanStore cache between concurrent spanById operations


praetoriansentry:
Concurrent read/write access to SpanStore's internal cache at 0x00c0011f1f08. Both operations occur in spanById (span_store.go:164-165) when caching span data during block verification. Race between parallel VerifyHeader and VerifyHeaders goroutines accessing validator snapshots.

https://github.com/0xPolygon/bor/commit/8509ba2

```
INFO [08-28|00:52:54.441] Span check complete                      foundNewSpan=true
INFO [08-28|00:52:55.004] Imported new chain segment               number=387 hash=96800c..6d3f8a blocks=1 txs=0  mgas=0.000 elapsed=1.360ms     mgasps=0.000   snapdiffs=6.53KiB triediffs=41.80KiB triedirty=0.00B
INFO [08-28|00:52:55.397] Whitelisting new milestone from heimdall block=386 hash=0a449e..76d92b
==================
WARNING: DATA RACE
Read at 0x00c0011f1f08 by goroutine 7351:
  github.com/ethereum/go-ethereum/consensus/bor.(*SpanStore).spanById()
      /var/lib/bor/consensus/bor/span_store.go:164 +0x424
  github.com/ethereum/go-ethereum/consensus/bor.(*SpanStore).getFutureSpan()
      /var/lib/bor/consensus/bor/span_store.go:236 +0x10d
  github.com/ethereum/go-ethereum/consensus/bor.(*SpanStore).spanByBlockNumber()
      /var/lib/bor/consensus/bor/span_store.go:208 +0x3a8
  github.com/ethereum/go-ethereum/consensus/bor.(*Bor).getVeBlopSnapshot()
      /var/lib/bor/consensus/bor/bor.go:665 +0x16b
  github.com/ethereum/go-ethereum/consensus/bor.(*Bor).snapshot()
      /var/lib/bor/consensus/bor/bor.go:547 +0x2a4
  github.com/ethereum/go-ethereum/consensus/bor.(*Bor).verifySeal()
      /var/lib/bor/consensus/bor/bor.go:785 +0x129
  github.com/ethereum/go-ethereum/consensus/bor.(*Bor).verifyCascadingFields()
      /var/lib/bor/consensus/bor/bor.go:527 +0x635
  github.com/ethereum/go-ethereum/consensus/bor.(*Bor).verifyHeader()
      /var/lib/bor/consensus/bor/bor.go:455 +0xb10
[...]

Issue #1735, opened 2025-08-28
Data race on milestone finality field between IsValidPeer reader and Process writer


praetoriansentry:
Concurrent access to milestone's finality field at 0x00c000262588. Read occurs in IsValidPeer (milestone.go:115) during peer validation in downloader while write happens in finality.Process (finality.go:76) when processing milestone updates. Race between sync downloader and milestone handler goroutines.

https://github.com/0xPolygon/bor/commit/8509ba2

```
==================
WARNING: DATA RACE
Read at 0x00c000262588 by goroutine 4061:
  github.com/ethereum/go-ethereum/eth/downloader/whitelist.(*milestone).IsValidPeer()
      /var/lib/bor/eth/downloader/whitelist/milestone.go:115 +0x132
  github.com/ethereum/go-ethereum/eth/downloader/whitelist.(*Service).IsValidPeer()
      /var/lib/bor/eth/downloader/whitelist/service.go:104 +0x72
  github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).findAncestor()
      /var/lib/bor/eth/downloader/bor_downloader.go:1080 +0x3be
  github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).syncWithPeer()
      /var/lib/bor/eth/downloader/bor_downloader.go:651 +0xba4
  github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).fetchHead()
      /var/lib/bor/eth/downloader/bor_downloader.go:949 +0xcf
  github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).syncWithPeer()
      /var/lib/bor/eth/downloader/bor_downloader.go:548 +0x93d
  github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).synchronise()
      /var/lib/bor/eth/downloader/bor_downloader.go:510 +0x959
  github.com/ethereum/go-ethereum/eth/downloader.(*Downloader).LegacySync()
      /var/lib/bor/eth/downloader/bor_downloader.go:380 +0xaa
  github.com/ethereum/go-ethereum/eth.(*handler).doSync()
      /var/lib/bor/eth/sync.go:276 +0x2e4
  github.com/ethereum/go-ethereum/eth.(*chainSyncer).startSync.func1()
      /var/lib/bor/eth/sync.go:252 +0x50

Previous write at 0x00c000262588 by goroutine 201:
[...]

Issue #1731, opened 2025-08-28
Data race on BlockChain state sync field between SetStateSync writer and insertChainWithWitnesses reader


praetoriansentry:
Concurrent access to state sync field at 0x00c0011bd460. Write occurs in SetStateSync (blockchain_reader.go:544) during block finalization while read happens in insertChainWithWitnesses (blockchain.go:2710) during chain insertion. Race between miner worker and block fetcher import goroutines.

https://github.com/0xPolygon/bor/commit/8509ba2

``` 
==================
WARNING: DATA RACE
Write at 0x00c0011bd460 by goroutine 103:
  github.com/ethereum/go-ethereum/core.(*BlockChain).SetStateSync()
      /var/lib/bor/core/blockchain_reader.go:544 +0x44
  github.com/ethereum/go-ethereum/consensus/bor.(*Bor).FinalizeAndAssemble()
      /var/lib/bor/consensus/bor/bor.go:1110 +0x983
  github.com/ethereum/go-ethereum/miner.(*worker).commit()
      /var/lib/bor/miner/worker.go:1634 +0x2c4
  github.com/ethereum/go-ethereum/miner.(*worker).commitWork()
      /var/lib/bor/miner/worker.go:1584 +0x94f
  github.com/ethereum/go-ethereum/miner.(*worker).mainLoop()
      /var/lib/bor/miner/worker.go:640 +0x958
  github.com/ethereum/go-ethereum/miner.newWorker.gowrap1()
      /var/lib/bor/miner/worker.go:354 +0x33

Previous read at 0x00c0011bd460 by goroutine 391:
  github.com/ethereum/go-ethereum/core.(*BlockChain).insertChainWithWitnesses()
      /var/lib/bor/core/blockchain.go:2710 +0x4aca
  github.com/ethereum/go-ethereum/core.(*BlockChain).InsertChainWithWitnesses()
      /var/lib/bor/core/blockchain.go:2185 +0x1212
  github.com/ethereum/go-ethereum/eth.newHandler.func3()
      /var/lib/bor/eth/handler.go:274 +0x5bd
  github.com/ethereum/go-ethereum/eth/fetcher.(*BlockFetcher).importBlocks.func1()
      /var/lib/bor/eth/fetcher/block_fetcher.go:1223 +0xa71

Goroutine 103 (running) created at:
  github.com/ethereum/go-ethereum/miner.newWorker()
      /var/lib/bor/miner/worker.go:354 +0x10d4
[...]

Issue #1728, opened 2025-08-28
Data race on BlockChain state sync field between writeBlockAndSetHead and SetStateSync


praetoriansentry:
Concurrent access to BlockChain's state sync field at address 0x00c0010f23e0. Read occurs in writeBlockAndSetHead (blockchain.go:2110) while write happens in SetStateSync (blockchain_reader.go:544) during block finalization. Race between worker's resultLoop and mainLoop goroutines.

https://github.com/0xPolygon/bor/commit/8509ba2

```
INFO [08-28|00:45:08.256] Sealing out-of-turn                      number=1 hash=0x2aa3640      wiggle-in-sec=4,000,000,000 wiggle=4s in-turn-signer=0x97538585a02A3f1B1297EB9979cE1b34ff953f1E
INFO [08-28|00:45:08.256] Sealing successful                       number=1 delay=-255.953371ms headerDifficulty=1
==================
WARNING: DATA RACE
Read at 0x00c0010f23e0 by goroutine 102:
  github.com/ethereum/go-ethereum/core.(*BlockChain).writeBlockAndSetHead()
      /var/lib/bor/core/blockchain.go:2110 +0x4b1
  github.com/ethereum/go-ethereum/core.(*BlockChain).WriteBlockAndSetHead()
      /var/lib/bor/core/blockchain.go:2057 +0x1b9
  github.com/ethereum/go-ethereum/miner.(*worker).resultLoop()
      /var/lib/bor/miner/worker.go:864 +0x1859
  github.com/ethereum/go-ethereum/miner.newWorker.gowrap3()
      /var/lib/bor/miner/worker.go:356 +0x33

Previous write at 0x00c0010f23e0 by goroutine 100:
  github.com/ethereum/go-ethereum/core.(*BlockChain).SetStateSync()
      /var/lib/bor/core/blockchain_reader.go:544 +0x44
  github.com/ethereum/go-ethereum/consensus/bor.(*Bor).FinalizeAndAssemble()
      /var/lib/bor/consensus/bor/bor.go:1110 +0x983
  github.com/ethereum/go-ethereum/miner.(*worker).commit()
      /var/lib/bor/miner/worker.go:1634 +0x2c4
  github.com/ethereum/go-ethereum/miner.(*worker).commitWork()
      /var/lib/bor/miner/worker.go:1584 +0x94f
  github.com/ethereum/go-ethereum/miner.(*worker).mainLoop()
[...]

PR #1748, opened 2025-09-08
fix: blockchain, span_store, finality data races

kamuikatsurgi:
# Description

Fixes 
- https://github.com/0xPolygon/bor/issues/1728
- https://github.com/0xPolygon/bor/issues/1731
- https://github.com/0xPolygon/bor/issues/1735
- https://github.com/0xPolygon/bor/issues/1737

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -312,6 +312,7 @@ type BlockChain struct {
 
 	// Bor related changes
 	borReceiptsCache *lru.Cache[common.Hash, *types.Receipt] // Cache for the most recent bor receipt receipts per block
+	stateSyncMu      sync.RWMutex                            // Mutex to protect the stateSyncData access
 	stateSyncData    []*types.StateSyncData                  // State sync data
 	stateSyncFeed    event.Feed                              // State sync feed
 	chain2HeadFeed   event.Feed                              // Reorg/NewHead/Fork data feed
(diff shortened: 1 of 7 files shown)


core/blockchain.go
```diff
@@ -2707,9 +2710,11 @@ func (bc *BlockChain) insertChainWithWitnesses(chain types.Blocks, setHead bool,
 		}
 
 		// BOR state sync feed related changes
+		bc.stateSyncMu.RLock()
```
lucca30, line 2713: what about using `GetStateSync()` instead of the accessing it directly? So avoid this lock/unlock since the get already have it.

lucca30 (commented):
lgtm, just 1 minor suggestion on statesyncdata access

lucca30 (approved):

core/blockchain.go
kamuikatsurgi: The lock is required while iterating because if another goroutine replaces entries in bc.stateSyncData[i] while we are iterating (write to the same indices), that’s also a data race.

```go
stateSyncData := bc.GetStateSync() // lock released here
for _, data := range stateSyncData { // reading element slots + following pointers
    bc.stateSyncFeed.Send(StateSyncEvent{Data: data})
}
```

PR #1764, opened 2025-09-13
eth, internal/cli: validate correctness of forks without whitelisted entry

manav2401:
# Description

This PR adds an additional safety check to validate if chains being imported belong to correct fork or not. 

Bor whitelists checkpoints and milestones at regular intervals as soon as they're produced by heimdall. This info is later used in validating chains while importing so that wrong forks can be rejected (which do not honour the whitelisted entries). Although for forks which doesn't overlap with the whitelist entries, we blindly accept them as there's no way for us to validate if that fork is correct. Only when next milestone or checkpoint is proposed, we can surely comment on the correctness of a fork. 

This PR adds a way to iterate backwards using the parent hash of the incoming chain until we can validate it against the last whitelisted entry. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/downloader/whitelist/service.go
+++ b/eth/downloader/whitelist/service.go
@@ -40,9 +40,6 @@ type Service struct {
 }
 
 func NewService(db ethdb.Database, disableBlindForkValidation bool, maxBlindForkValidationLimit uint64) *Service {
-	if disableBlindForkValidation {
-		log.Info("Disabling blind fork validation")
-	}
 	if maxBlindForkValidationLimit == 0 {
 		maxBlindForkValidationLimit = DefaultMaxForkCorrectnessLimit
 		log.Info("Invalid max blind fork validation limit, falling back to default", "limit", DefaultMaxForkCorrectnessLimit)
@@ -52,6 +49,11 @@ func NewService(db ethdb.Database, disableBlindForkValidation bool, maxBlindFork
 	var forkValidationCacheSize = maxBlindForkValidationLimit
 	if forkValidationCacheSize > 2*DefaultMaxForkCorrectnessLimit {
 		forkValidationCacheSize = 2 * DefaultMaxForkCorrectnessLimit
+	}
+
+	if disableBlindForkValidation {
+		forkValidationCacheSize = 0
+		log.Info("Disabling blind fork validation")
 	}
 
 	// Fetch last whitelisted checkpoint entry from db. Ignore in case of error or if


eth/downloader/whitelist/finality.go
```diff
@@ -67,18 +67,12 @@ func (f *finality[T]) IsValidChain(currentHeader *types.Header, chain []*types.H
 		return false, nil
 	}
 
-	f.RLock()
```
manav2401, line 70: This was added in https://github.com/0xPolygon/bor/pull/1757 but we already acquire a lock in the respective checkpoint and milestone services so this is redundant and not needed.

eth/downloader/whitelist/service.go
```diff
@@ -145,9 +164,129 @@ func (s *Service) IsValidChain(currentHeader *types.Header, chain []*types.Heade
...
+	}
+
+	// Blind accept the chain if we've to iterate more than `maxForkCorrectnessLimit` blocks
+	if headerNumber-lastKnownValidBlock > s.maxForkCorrectnessLimit {
+		log.Debug("Skipping fork correctness check as block is too far ahead", "block", headerNumber, "last whitelisted", number)
+		return true
+	}
+
+	// Track all blocks iterated for caching
+	var blocksChecked []common.Hash = make([]common.Hash, 0, DefaultMaxForkCorrectnessLimit)
```
lucca30: Probably better using `s.maxForkCorrectnessLimit` on allocation size.

eth/downloader/whitelist/service.go
```diff
@@ -145,9 +164,129 @@ func (s *Service) IsValidChain(currentHeader *types.Header, chain []*types.Heade
...
+		}
+	}
+
+	var lastKnownValidBlock uint64 = number
+	if s.lastValidForkBlock > number {
+		lastKnownValidBlock = s.lastValidForkBlock
+	}
+
+	// Blind accept the chain if we've to iterate more than `maxForkCorrectnessLimit` blocks
+	if headerNumber-lastKnownValidBlock > s.maxForkCorrectnessLimit {
```
lucca30: May potentially face an unsigned underflow since it's a uint and we don't check whether `lastKnownValidBlock < headerNumber`

eth/downloader/whitelist/service.go
```diff
@@ -145,9 +164,129 @@ func (s *Service) IsValidChain(currentHeader *types.Header, chain []*types.Heade
...
+		log.Debug("Skipping fork correctness check as block is too far ahead", "block", headerNumber, "last whitelisted", number)
+		return true
+	}
+
+	// Track all blocks iterated for caching
+	var blocksChecked []common.Hash = make([]common.Hash, 0, DefaultMaxForkCorrectnessLimit)
+
+	// Acquire lock over the cache
+	s.forkValidationCacheMu.Lock()
```
lucca30, line 248: May be woth including this section:

```
	var lastKnownValidBlock uint64 = number
	if s.lastValidForkBlock > number {
		lastKnownValidBlock = s.lastValidForkBlock
	}
```

under the mutex protection

eth/downloader/whitelist/service.go
manav2401: I am planning to parameterise that via a cli flag and hence don't want the allocation to be extra heavy if the size is set to some very high number. Hence, used the default value.

eth/downloader/whitelist/service.go
```diff
@@ -145,7 +164,128 @@ func (s *Service) IsValidChain(currentHeader *types.Header, chain []*types.Heade
...
+	defer s.forkValidationCacheMu.Unlock()
+
+	// Only perform checks if first block being imported is ahead of last whitelisted entry. We can
+	// skip checking for fork correctness otherwise as chain would be already validated before.
+	if headerNumber > number {
+		parentNumber := headerNumber - 1
+		parentHash := firstBlock.ParentHash
+		for {
```
cffls, line 256: Maybe it would be safer to put a loop condition here, e.g. check no more than X number of block?

cffls:
`make test-integration` seems to fail when running local for some reason:

```
--- FAIL: TestAPIs (0.21s)
    bor_api_test.go:95:
        	Error Trace:	/home/ubuntu/bor/tests/bor/bor_api_test.go:95
        	            				/home/ubuntu/bor/tests/bor/bor_api_test.go:251
        	Error:      	Not equal:
        	            	expected: 2
        	            	actual  : 1
        	Test:       	TestAPIs
panic: runtime error: index out of range [1] with length 1 [recovered]
	panic: runtime error: index out of range [1] with length 1

goroutine 13 [running]:
testing.tRunner.func1.2({0x203f500, 0xc00082adc8})
	/home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/testing/testing.go:1734 +0x21c
testing.tRunner.func1()
	/home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/testing/testing.go:1737 +0x35e
panic({0x203f500?, 0xc00082adc8?})
	/home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/runtime/panic.go:792 +0x132
github.com/ethereum/go-ethereum/tests/bor.testGetTransactionReceiptsByBlock(0xc000585500, 0xc000892350)
	/home/ubuntu/bor/tests/bor/bor_api_test.go:106 +0xcbd
github.com/ethereum/go-ethereum/tests/bor.TestAPIs(0xc000585500)
	/home/ubuntu/bor/tests/bor/bor_api_test.go:251 +0x5f4
testing.tRunner(0xc000585500, 0x2465bb0)
	/home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/testing/testing.go:1792 +0xf4
created by testing.(*T).Run in goroutine 1
	/home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/testing/testing.go:1851 +0x413
FAIL	github.com/ethereum/go-ethereum/tests/bor	0.251s
```

eth/downloader/whitelist/service.go
manav2401: I am already checking for how long we have to go back in past to validate the chain [here](https://github.com/0xPolygon/bor/pull/1764/files#diff-7b1971c3789e15f3b09e59dfa774e88a6ac4641fdb75d638fabd6e38d419cb74R213). If it's too long, we return directly.

eth/downloader/whitelist/service.go
manav2401: I checked the code and there's no other place where `s.lastValidForkBlock` is updated apart from the loop below. It's updated along with the cache so the writes are anyways protected. As there's no other writes happening, reads are protected by default IMO.

eth/downloader/whitelist/service.go
manav2401: Fixed in https://github.com/0xPolygon/bor/pull/1764/commits/b73ccdd92beea775abc1ca72ec18deb8d25aa7a8

manav2401:
> make test-integration seems to fail when running local for some reason:

Strange. It passed the first two times locally and then encountered the same panic. Maybe it's non deterministic. But this test even fails on `master` so probably it's unrelated.

PR #1766, opened 2025-09-15
eth: implemented peer consensus based witness verification

pratikspatil024:
# Description

Implemented peer consensus based witness verification

1. Warning Threshold System
   - Triggers verification when a peer reports # pages > allowed (calculated dynamically)
   - Peers under threshold are assumed honest
2. Peer Consensus Verification
   - Query 2 random peers for verification
   - Use collective peer knowledge to detect malicious behaviour
3. Peer Management
   - Drop peers immediately
   - Cache verification results to avoid redundant checks

```
CheckWitnessPageCount() - Entry point for verification
verifyWitnessPageCount() - Core consensus logic
RequestWitnessesWithVerification() - Verification-enabled witness requests
getAllPeers() - Random peer selection support

//configs
const (
    witnessPageWarningThreshold = 10               // Trigger verification threshold
    witnessVerificationPeers    = 2                // Number of peers to query
    witnessVerificationTimeout  = 5 * time.Second  // Query timeout
    witnessVerificationCacheTTL = 10 * time.Minute // Cache duration
)
```

Verification Flow:
```

1. Node A requests witness 0xABC... from Peer B.
2. Downloads Page 0, learns there are 50 total pages.
3. Locks TotalPages = 50.
4. Calculates threshold → ceil(45MB / 15MB) = 3 pages.
5. Since 50 > 3, pauses to verify.
6. Downloads Pages 1, 2, and 3 (up to threshold).Bandwidth used so far: 48MB.
7. Cache check → MISS.
8. Selects two random peers → Peer C and Peer D.
9. Requests metadata (WIT2) from both peers.
10. Both report: TotalPages = 50, Size = 800MB, Block = 12345.
11. Bandwidth for metadata: 180 bytes.
12. Calculates consensus → majority agrees on 50 pages.
13. Peer B deemed honest.
14. Caches result {0xABC...: 50 pages, TTL = 10 min}.
15. Resumes download from Peer B (Pages 4–49).
16. Validates each page (total count, page index, limits).
17. Completes full download (50 pages = 800MB).
18. Reconstructs witness and imports it.
```

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/protocols/wit/handlers_test.go
+++ b/eth/protocols/wit/handlers_test.go
@@ -123,8 +123,8 @@ func TestGetWitnessPacket(t *testing.T) {
 
 // TestProtocolVersionConstants tests that protocol version constants are defined correctly
 func TestProtocolVersionConstants(t *testing.T) {
-	assert.Equal(t, 0, WIT0, "WIT0 should be version 0")
-	assert.Equal(t, 1, WIT1, "WIT1 should be version 1")
+	assert.Equal(t, 1, WIT0, "WIT0 should be version 1")
+	assert.Equal(t, 2, WIT1, "WIT1 should be version 2")
 	assert.True(t, WIT1 > WIT0, "WIT1 should be greater than WIT0")
 }
 
--- a/eth/protocols/wit/protocol.go
+++ b/eth/protocols/wit/protocol.go
@@ -9,8 +9,8 @@ import (
 
 // Constants to match up protocol versions and messages
 const (
-	WIT0 = 0
-	WIT1 = 1
+	WIT0 = 1
+	WIT1 = 2
 )
 
 // ProtocolName is the official short name of the `wit` protocol used during


core/stateless/witness_test.go
```diff
@@ -256,3 +257,189 @@ func TestValidateWitnessPreState_MultipleHeaders(t *testing.T) {
...
+		witnessVerificationPeers    = 2
+	)
+
+	if witnessPageWarningThreshold != 10 {
+		t.Errorf("Expected witnessPageWarningThreshold to be 10, got %d", witnessPageWarningThreshold)
+	}
+	if witnessVerificationPeers != 2 {
+		t.Errorf("Expected witnessVerificationPeers to be 2, got %d", witnessVerificationPeers)
+	}
+}
```
lucca30: Not sure if I understood this test. Is it expected to import the constants from `witness_manager.go` and check them?

eth/handler_eth.go
```diff
@@ -130,8 +132,62 @@ func (h *ethHandler) handleBlockAnnounces(peer *eth.Peer, hashes []common.Hash,
...
+					// Wait for response with timeout
+					select {
+					case res := <-resCh:
+						if res == nil {
+							return 0, fmt.Errorf("no response from peer %s", peerID)
+						}
+						// Extract page count from response
+						if witPacket, ok := res.Res.(*wit.WitnessPacketRLPPacket); ok && len(witPacket.WitnessPacketResponse) > 0 {
+							return witPacket.WitnessPacketResponse[0].TotalPages, nil
+						}
```
lucca30: The `res` from `resCh` is not `*wit.WitnessPacketRLPPacket` but actually `[]*stateless.Witness`.

Reference: https://github.com/0xPolygon/bor/blob/8c7fa0489c145cd318623d28bbf99a673d214459/eth/peer.go#L256-L276


A few things we should consider:

- By requesting that way, we are downloading the whole witness, not a single page. (In this case `>160MB`)
- Even requesting a single page, still a big amount of data (`16MB` for each peer)
- We should consider implementing a new method to just retrieve the page count. (may be useful for future scenarios)

cffls:
Since this isn't a necessary feature in the initial veblop release, let's use `develop` as the base branch.

core/stateless/witness_test.go
```diff
@@ -256,3 +257,189 @@ func TestValidateWitnessPreState_MultipleHeaders(t *testing.T) {
 		t.Errorf("Expected validation to succeed with multiple headers, but got error: %v", err)
 	}
 }
+
+// TestWitnessVerificationConstants tests the verification constants
+func TestWitnessVerificationConstants(t *testing.T) {
+	// These constants should match the ones defined in eth/fetcher/witness_manager.go
+	const (
+		witnessPageWarningThreshold = 10
```
cffls: Might be better to make this a dynamic parameter based on the block gas limit.

pratikspatil024: [done](https://github.com/0xPolygon/bor/pull/1766/commits/e94a428593924b13d0d514078157eb69a49a8789).

core/stateless/witness_test.go
pratikspatil024: this did not makes sense. removed it. Thanks

eth/handler_eth.go
pratikspatil024: Thanks Lucca, makes sense

pratikspatil024: does [this](https://github.com/0xPolygon/bor/pull/1766/commits/439e927b6c480e93f60af716e042cd0ac9eb7250) makes sense?

pratikspatil024: BTW, this does not add a separate wit protocol method, as it would need a protocol upgrade.
Do you think I should do that?
@lucca30 @cffls

lucca30: Hi @pratikspatil024 , good question. Introducing a new method allows us reducing from 16MB in this packet to basically 0. So it seems like an interesting upgrade. We can maybe deliver this first version that way but keep the upgrade on backlog.

eth/handler_eth.go
```diff
@@ -130,8 +131,42 @@ func (h *ethHandler) handleBlockAnnounces(peer *eth.Peer, hashes []common.Hash,
 				return nil, fmt.Errorf("no peer with witness for hash %s is available", hash)
 			}
 
-			// Request witnesses using the wit peer
-			return ethPeer.RequestWitnesses([]common.Hash{hash}, sink)
+			// Create verification callback for page count verification (returns true if honest)
+			verifyPageCount := func(hash common.Hash, pageCount uint64, peer string) bool {
```
cffls: Could this be moved to a standalone function instead of being an embedded function?

eth/peer.go
```diff
@@ -368,11 +493,61 @@ func (p *ethPeer) receiveWitnessPage(
...
+		// Trigger page count verification if callback is provided (only on first page)
+		// If verification fails (peer is dishonest), pause download and mark for cancellation
+		if verifyPageCount != nil && !hasTotalPages {
+			isHonest := verifyPageCount(page.Hash, page.TotalPages, p.ID())
+			if !isHonest {
+				// Peer is dishonest - pause download and discard pages
+				mapsMu.Lock()
+				downloadPaused[page.Hash] = true
+				mapsMu.Unlock()
+				p.witPeer.Peer.Log().Warn("Peer failed verification, pausing witness download", "peer", p.ID(), "hash", page.Hash, "totalPages", page.TotalPages)
```
cffls: Should the node drop a peer here if the peer is dishonest? The block hash is added to `downloadPaused` but no further action is taken it seems.

eth/handler_eth.go
```diff
@@ -130,8 +131,42 @@ func (h *ethHandler) handleBlockAnnounces(peer *eth.Peer, hashes []common.Hash,
...
+					if peer == nil || !peer.SupportsWitness() {
+						return 0, fmt.Errorf("peer %s not available or doesn't support witness", peerID)
+					}
+
+					// Use the new efficient method that only downloads page 0
+					return peer.RequestWitnessPageCount(hash)
+				}
+
+				// Run synchronous verification and return result
+				return h.blockFetcher.GetWitnessManager().CheckWitnessPageCount(hash, pageCount, peer, getRandomPeers, getWitnessPageCount)
```
cffls: Should we call `CheckWitnessPageCount` first before sampling from peers? Because it would avoid sending unnecessary metadata request to peers in most cases.

cffls (commented):
Has this been tested in a devnet? It would be great if we can create some test scenarios for this new logic.

eth/handler_eth.go
pratikspatil024: [Moved](https://github.com/0xPolygon/bor/pull/1766/commits/d9f03718659744ac43c24f94449db94e31293504), thanks!

eth/peer.go
pratikspatil024: Nice catch! [Fixed](https://github.com/0xPolygon/bor/pull/1766/commits/ba40cb6747c69c75e2acd1d7dce10900ac72bb67), thanks!

eth/handler_eth.go
pratikspatil024: This is what's happening right now, the `getRandomPeers` function is used at a later point. [Added comments](https://github.com/0xPolygon/bor/pull/1766/commits/5cb56b47967c663d79e50e59c9b5a9dc93b0ed60).

Let me know if this is not what you meant, thanks!

ssandeep (approved):

adamdossa:
@claude Can you review this PR focussing on safety. Provide a short summary as a comment highlighting any concerns related to these changes.

PR #1767, opened 2025-09-15
core: add witness cleanup to setHead operation in stateless mode

kamuikatsurgi:
# Description

Added witness deletion to the `delFn` callback in `setHead`.


lucca30 (approved):
lgtm

PR #1290, opened 2024-07-16
PIP-35: enforce gas related configs for mainnet

manav2401:
# Description

This PR generalises the 25 gwei gas enforcement (based on [PIP-35](https://github.com/maticnetwork/Polygon-Improvement-Proposals/blob/main/PIPs/PIP-35.md)) to all networks (i.e. specifically polygon mainnet). Earlier, it was only introduced to amoy. This PR removes those checks such that it's applicable to mainnet as well. 

# Changes

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)
- [ ] Changes only for a subset of nodes

# Breaking changes

The values `txpool.pricelimit`, `miner.gasprice` and `gpo.ignoreprice` are now enforced to 25 gwei

# Nodes audience

Applicable to all

# Checklist

- [x] I have added at least 2 reviewer or the whole pos-v1 team
- [x] I have added sufficient documentation in code
- [x] I will be resolving comments - if any - by pushing each fix in a separate commit and linking the commit hash in the comment reply
- [x] Created a task in Jira and informed the team for implementation in Erigon client (if applicable)
- [x] Includes RPC methods changes, and the Notion documentation has been updated

## Testing

- [ ] I have added unit tests
- [ ] I have added tests to CI
- [ ] I have tested this code manually on local environment
- [ ] I have tested this code manually on remote devnet using express-cli
- [ ] I have tested this code manually on mumbai/amoy
- [ ] I have created new e2e tests into express-cli

### Manual tests

Please complete this section with the steps you performed if you ran manual tests for this functionality, otherwise delete it

# Additional comments

Please post additional comments in this section if you have them, otherwise delete it

PR #1773, opened 2025-09-17
Remove minimum gas enforcement while creating Accesslist

lucca30:
# Description

Because our minimum gas enforcement like the one at [PIP-35](https://github.com/0xPolygon/bor/pull/1290/files), the accesslist creation was including this gas when calculating it. By doing that, addresses with no fund were not able to calculate access list via `eth_createAccessList` receiving responses like `insufficient funds for gas * price + value`.

This PR handles this by enforcing 0 gas while calculating the accesslist.

# Changes


cffls:
Look good! Is it possible to add a unit test to verify the fix?

lucca30:
tests done @cffls

cffls (approved):
LGTM, thanks!

PR #1779, opened 2025-09-18
Move fields to the right place

cffls:
# Description

A few fields, e.g. gasLimit, was written in the wrong place in Amoy gensis json, making some toolings fail to generate the genesis hash correctly for Amoy.


lucca30 (approved):
lgtm

PR #1781, opened 2025-09-19
Larger tx announce queue for trusted peers

cffls:
# Description

Allows the node to allocate more buffer (x10) for txn announcements for trusted and static peers. Larger queue reduces the possibility a transaction might get dropped by the node when there is a surge of txns in the network.

# Changes

### DIFF
--- a/eth/protocols/eth/broadcast.go
+++ b/eth/protocols/eth/broadcast.go
@@ -142,6 +142,12 @@ func (p *Peer) announceTransactions() {
 		fail   = make(chan error, 1) // Channel used to receive network error
 		failed bool                  // Flag whether a send failed, discard everything onward
 	)
+
+	queueLimit := maxQueuedTxAnns
+	if p.IsTrusted() || p.IsStatic() {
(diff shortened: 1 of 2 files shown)


cffls:
> LGTM, maybe worth adding a comment stating that this is specific to Polygon, so that it should not be removed accidentally during upstream merge.

Thanks, addressed in the comment.

lucca30 (approved):

eth/protocols/eth/broadcast.go
```diff
@@ -142,6 +142,12 @@ func (p *Peer) announceTransactions() {
 		fail   = make(chan error, 1) // Channel used to receive network error
 		failed bool                  // Flag whether a send failed, discard everything onward
 	)
+
+	queueLimit := maxQueuedTxAnns
+	if p.IsTrusted() || p.IsStatic() {
+		queueLimit = maxQueuedTxAnnsTrusted
```
ssandeep, line 148: Should we have just considered 10x maxQueuedTxAnns rather than defining another const?

PR #1786, opened 2025-09-23
Add support for filtering txs given a list of accounts

Raneet10:
# Description

This PR enables miners to filter out txs sent from certain addresses, which can be configure in a comma-separated text file and passed as `filtered-addresses` in toml config or `txpool.filtered-addresses` in cli. 

# Changes

# Checklist

## Testing

- [x] I have added unit tests


core/txpool/legacypool/legacypool.go
```diff
@@ -270,6 +279,10 @@ type LegacyPool struct {
 	changesSinceReorg int // A counter for how many drops we've performed in-between reorg.
 
 	promoteTxCh chan struct{} // should be used only for tests
+
+	// Censorship fields
+	censoredAddrs map[common.Address]bool // Map of addresses to censor
```
cffls: `map[common.Address]struct{}` will be more efficient than using bool.

core/txpool/legacypool/legacypool.go
```diff
@@ -306,6 +320,13 @@ func New(config Config, chain BlockChain, options ...func(pool *LegacyPool)) *Le
 		fn(pool)
 	}
 
+	// Load censored addresses if configured
+	if config.CensorshipFile != "" {
```
cffls: If possible, it might be better to move file loading to config, as it would fail as early as possible. Also file loading fits more to the scope of config/cli module.

core/txpool/legacypool/legacypool.go
```diff
@@ -2061,3 +2094,57 @@ func (pool *LegacyPool) Clear() {
 func (pool *LegacyPool) HasPendingAuth(addr common.Address) bool {
 	return pool.all.hasAuth(addr)
 }
+
+// loadCensoredAddresses loads comma-separated addresses to censor from the configured TXT file.
```
cffls: Why not separate by new line character? It would make the file more readable.

core/txpool/legacypool/legacypool.go
```diff
@@ -2061,3 +2094,57 @@ func (pool *LegacyPool) Clear() {
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			log.Warn("Censorship file not found", "file", pool.config.CensorshipFile)
+			return nil
+		}
+		return err
+	}
+
+	pool.censorMu.Lock()
+	defer pool.censorMu.Unlock()
```
cffls: If file loading is moved out of the scope of legacypool, we wouldn't need a write lock here.

core/txpool/legacypool/legacypool.go
Raneet10: Ahh yes missed that.

core/txpool/legacypool/legacypool.go
Raneet10: Well I was expecting that the list practically won't be too big so wouldn't make much difference. But sure I can make it separated by newline.

core/txpool/legacypool/legacypool.go
Raneet10: Done.

core/txpool/legacypool/legacypool.go
Raneet10: Done.

core/txpool/legacypool/legacypool.go
Raneet10: Done.

core/txpool/legacypool/legacypool.go
Raneet10: Done now.

PR #1796, opened 2025-09-29
Fix getSnapshotProposerSequence

cffls:
# Description

Use the correct block number for retrieving the snapshot. This is change is required because the behavior  of `snapshot()` in bor consensus had changed. Before, it was expecting the caller to pass the parent block of the target block, but now it is expecting the target block to be directly passed in.

Tested in a devnode. 

# Changes


lucca30 (approved):
lgtm

PR #1817, opened 2025-10-11
Cutomizable block time in block producer

cffls:
# Description

Allow block producers to create blocks at an interval longer than the standard block time.
The motivation of this feature is to 1) enable sub-second granularity of block interval and 2) allow miners to adjust block intervals without a hardfork in consensus.

Notice that this brings no changes to block header - the timestamp in the header is still the standard unix timestamp in seconds. When the miner configures to use a higher block time than the consensus, it would announce the block later than the actual timestamp in the header, which is always the floor of the actual announcement time. Example:

```
Block number:    1, 2,    3,   4,    5,  6
Block announce:  1, 2.75, 4.5, 6.26, 8,  9.75 (1.75s apart)
Block timestamp: 1, 2,    4,   6,    8,  9   (either 1s or 2s apart)
```

# Changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### DIFF
--- a/consensus/bor/bor_test.go
+++ b/consensus/bor/bor_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"math/big"
 	"testing"
+	"time"
 
 	"github.com/holiman/uint256"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


consensus/bor/bor.go
```diff
@@ -996,7 +1020,7 @@ func (c *Bor) Prepare(chain consensus.ChainHeaderReader, header *types.Header) e
 		// need a check for hard fork as it doesn't change any consensus rules, we
 		// still keep it for safety and testing.
 		if c.config.IsBhilai(big.NewInt(int64(number))) && succession == 0 {
-			startTime := time.Unix(int64(header.Time-c.config.CalculatePeriod(number)), 0)
+			startTime := header.GetActualTime().Add(-time.Duration(c.config.CalculatePeriod(number)) * time.Second)
```
lucca30, line 1023: Use sub instead of Add and negative number:

```
header.GetActualTime().Sub(time.Duration(c.config.CalculatePeriod(number)) * time.Second)
```

pratikspatil024: Although true, in this codebase, I have seen a lot of places where we do `.Add(-ve)` to subtract.

core/types/block.go
```diff
@@ -98,6 +98,9 @@ type Header struct {
 	MixDigest   common.Hash    `json:"mixHash"`
 	Nonce       BlockNonce     `json:"nonce"`
 
+	// ActualTime is the actual time of the block. It is internally used by the miner.
```
pratikspatil024, line 101: just wondering, this won't be part of rlp encoding, right?

lucca30: yes, ``json:"-" rlp:"-"`` tells go to ignore encoding

consensus/bor/bor.go
cffls: Instead of calculating a new time, `time.Sub()` returns the duration between two times. https://pkg.go.dev/time#Time.Sub

cffls:
> I'm just curious to better understand why do we want to allow BPs announcing blocks later than expected.

The motivation is to allow BP to produce blocks with more granular block time, e.g. 1.75s. Although the timestamp is not spread evenly, the actual interval between announcement of blocks will be even. Disabling pip-66 is all about making the announcement time more smooth rather than changing the timestamp in the block header.

PR #1818, opened 2025-10-13
[WIP] witness state caching

pratikspatil024:
# Description

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/protocols/wit/handlers_test.go
+++ b/eth/protocols/wit/handlers_test.go
@@ -753,6 +753,8 @@ func TestWit1HandlerMap(t *testing.T) {
 		NewWitnessHashesMsg:   "handleNewWitnessHashes",
 		GetWitnessMetadataMsg: "handleGetWitnessMetadata",
 		WitnessMetadataMsg:    "handleWitnessMetadata",
+		GetCompactWitnessMsg:  "handleGetCompactWitness",
+		CompactWitnessMsg:     "handleCompactWitness",
 	}
 
 	// Check that all expected handlers are present in wit1
@@ -763,13 +765,17 @@ func TestWit1HandlerMap(t *testing.T) {
 	}
 
 	// Verify that wit1 has exactly 6 handlers (WIT1 has 6 message types)
-	assert.Equal(t, 6, len(wit1), "wit1 map should contain exactly 6 handlers")
+	assert.Equal(t, 8, len(wit1), "wit1 map should contain exactly 8 handlers")
 
 	// Verify that wit1 includes the WIT1-specific handlers that are not in wit0
 	_, exists := wit1[GetWitnessMetadataMsg]
 	assert.True(t, exists, "wit1 should contain GetWitnessMetadataMsg handler")
 	_, exists = wit1[WitnessMetadataMsg]
 	assert.True(t, exists, "wit1 should contain WitnessMetadataMsg handler")
+	_, exists = wit1[GetCompactWitnessMsg]
+	assert.True(t, exists, "wit1 should contain GetCompactWitnessMsg handler")
+	_, exists = wit1[CompactWitnessMsg]
+	assert.True(t, exists, "wit1 should contain CompactWitnessMsg handler")
 }
 
 // TestHandlerMapVersionSelection tests the switch statement that selects handlers based on protocol version


core/blockchain.go
```diff
@@ -3885,12 +3896,91 @@ func (bc *BlockChain) SubscribeChain2HeadEvent(ch chan<- Chain2HeadEvent) event.
...
+		if mergedCount > 0 {
+			log.Debug("Merged cached states into witness",
+				"block", blockNum,
+				"witnessStatesBefore", witnessStatesBefore,
+				"mergedFromCache", mergedCount,
+				"witnessStatesAfter", len(witness.State))
+		}
+	}
+
 	// Validate witness.
```
pratikspatil024, line 4460: I think we should move this before the merge (cache -> witness.state). Let me know if you think the same. @cffls @lucca30

lucca30: Not sure if I understood the reason behind it.

pratikspatil024: We validate the witness before the execution. Now once we have the cache, I merge the cache into the witness, so my question was to do this witness validation before the merge or leave it after the merge.

eth/protocols/wit/protocol.go
```diff
@@ -50,8 +52,10 @@ type Packet interface {
 }
 
 // GetWitnessRequest represents a list of witnesses query by witness pages.
+// Also used for reduced witness requests (distinguished by IsReduced flag).
 type GetWitnessRequest struct {
 	WitnessPages []WitnessPageRequest // Request by list of witness pages
+	IsReduced    bool                 // True if requesting reduced witness (omits cached states)
```
cffls: Maybe `Compact` is a bit more descriptive?

core/blockchain.go
```diff
@@ -4139,12 +4163,196 @@ func (bc *BlockChain) SubscribeChain2HeadEvent(ch chan<- Chain2HeadEvent) event.
...
+func (bc *BlockChain) manageSlidingWindow(blockNum uint64) bool {
+	blocksSinceWindowStart := blockNum - bc.cacheWindowStart
+
+	if blocksSinceWindowStart >= bc.cacheWindowSize {
+		// Time to slide the window: discard active, promote next, create new next
+		bc.cacheLock.Lock()
+		oldActiveSize := len(bc.activeCacheMap)
+		bc.activeCacheMap = bc.nextCacheMap
+		bc.nextCacheMap = make(map[string]struct{})
+		bc.cacheWindowStart = blockNum
```
cffls: How do we ensure different nodes can all agree on the same `cacheWindowStart`? From the current state of the code, it seems like this value will be different depending on at which block the node starts syncing.

core/blockchain.go
```diff
@@ -360,6 +361,15 @@ type BlockChain struct {
 	// future blocks are blocks added for later processing
 	futureBlocks *lru.Cache[common.Hash, *types.Block]
 
+	// Sliding window cache for reducing witness bandwidth
+	// Uses two maps with overlap for efficient small-window caching
+	activeCacheMap   map[string]struct{} // Currently active cache map
+	nextCacheMap     map[string]struct{} // Pre-warming map for next window
+	cacheWindowStart uint64              // Start block of current window
+	cacheWindowSize  uint64              // Size of each window (e.g., 20 blocks)
+	cacheOverlapSize uint64              // Overlap between windows (e.g., 10 blocks)
```
cffls: These values are tied to the consensus. Might be a good idea to make them constant.

core/blockchain.go
```diff
@@ -4139,12 +4163,196 @@ func (bc *BlockChain) SubscribeChain2HeadEvent(ch chan<- Chain2HeadEvent) event.
...
+	}
+
+	filtered.State = filteredStates
+
+	return filtered, originalStates, removed
+}
+
+// updateSlidingWindowCache updates the sliding window cache after processing a block.
+// This is used by both full nodes and witness-receiving nodes to maintain synchronized caches.
+func (bc *BlockChain) updateSlidingWindowCache(blockNum uint64, witness *stateless.Witness, statedb *state.StateDB) {
```
cffls: Would be good to have a timer metric here to understand the performance impact of constructing the caches.

eth/handler_wit.go
```diff
@@ -182,6 +197,96 @@ func (h *witHandler) handleGetWitness(peer *wit.Peer, req *wit.GetWitnessPacket)
 	return response, nil
 }
 
+// handleGetReducedWitness retrieves witnesses and filters them using the sliding window cache.
+// This reduces bandwidth by omitting state nodes that the receiver should already have cached.
+func (h *witHandler) handleGetReducedWitness(peer *wit.Peer, req *wit.GetWitnessPacket) (wit.WitnessPacketResponse, error) {
```
cffls: The result could be cached to reduce computation.

core/blockchain.go
```diff
@@ -4139,12 +4163,196 @@ func (bc *BlockChain) SubscribeChain2HeadEvent(ch chan<- Chain2HeadEvent) event.
...
+				"block", blockNum,
+				"witnessStatesBefore", witnessStatesBefore,
+				"mergedFromCache", mergedCount,
+				"witnessStatesAfter", len(witness.State))
+		}
+	}
+
 	// Validate witness.
 	// During parallel import, defer pre-state validation to the end of the batch.
 	if !bc.parallelStatelessImportEnabled.Load() {
```
cffls, line 4462: In parallelStatelessImport mode, witness cache won't work. Need to make sure parallel block import and witness cache are used in a mutually exclusive way.

core/blockchain.go
```diff
@@ -4139,12 +4163,196 @@ func (bc *BlockChain) SubscribeChain2HeadEvent(ch chan<- Chain2HeadEvent) event.
...
+			// Add to active map
+			if _, exists := bc.activeCacheMap[stateNode]; !exists {
+				bc.activeCacheMap[stateNode] = struct{}{}
+				cachedToActive++
+			}
+			// Also add to next map if in overlap period
+			if inOverlapPeriod {
+				if _, exists := bc.nextCacheMap[stateNode]; !exists {
+					bc.nextCacheMap[stateNode] = struct{}{}
+					cachedToNext++
```
cffls: It seems like the witness here is a super set of all trie nodes from active window start block to the current block (because of `mergeSpanCacheIntoWitness`) in stateless nodes. If this is true, `bc.nextCacheMap` will have the same size as `bc.activeCacheMap` and hence lose its original purpose - trim off unnecessary nodes. The correct implementation * I think * is to make sure the witness passed to `updateSlidingWindowCache` contains only the trie nodes that are used by the current block.

core/blockchain.go
pratikspatil024: Good point.
Just to double check - 
1. If the sender has parallel mode enabled, it will only send full witness (irrespective of the request)?
2. Also, if the receiver has parallel mode enabled, it will only be able to sync with full witness?

core/blockchain.go
pratikspatil024: Nice catch! Fixed [here](https://github.com/0xPolygon/bor/pull/1818/commits/61b1a40cedcb9e719aece0a5bdb1e82d86130c9a).

core/blockchain.go
pratikspatil024: Fixed [here](https://github.com/0xPolygon/bor/pull/1818/commits/ee7342f6610eb94cdd70140521a4749508dd267f), thanks!

eth/handler_wit.go
pratikspatil024: Does [this](https://github.com/0xPolygon/bor/pull/1818/commits/bee613647511c1778d25960cf4361e1bab510f50) look good? Unsure about the size of the cache though, it will be ok to cache blocks > length of the window, right?

core/blockchain.go
pratikspatil024: Added [here](https://github.com/0xPolygon/bor/pull/1818/commits/40a905135b4bdf467d53e9d24c73fa65166b9010).

core/blockchain.go
pratikspatil024: Updated, [thanks](https://github.com/0xPolygon/bor/pull/1818/commits/f05239e56c2a775359eb1bd62bba46391fd1c027)!

eth/protocols/wit/protocol.go
pratikspatil024: [Done](https://github.com/0xPolygon/bor/pull/1818/commits/737856fb641ee12e1b4b738a1221883387d38f96).

core/blockchain.go
pratikspatil024: Thanks! Fixed [here](https://github.com/0xPolygon/bor/pull/1818/commits/eddf23582c06031e9472efe8cf965f1cd4c104c5). Right now, the window size is 20, but we can change later if needed.

core/blockchain.go
```diff
@@ -4198,6 +4447,14 @@ func (bc *BlockChain) ProcessBlockWithWitnesses(block *types.Block, witness *sta
 		err = fmt.Errorf("stateless self-validation receipt root mismatch: remote %x != local %x", block.ReceiptHash(), crossReceiptRoot)
 		return nil, nil, err
 	}
+
+	// Update sliding window cache with ORIGINAL witness states and execution data
+	// Only in sequential mode - parallel import is incompatible with cache
+	// Pass originalWitnessStates (not full witness) to avoid re-caching merged states
+	if !bc.parallelStatelessImportEnabled.Load() {
+		bc.updateSlidingWindowCache(blockNum, originalWitnessStates, statedb)
```
cffls, line 4507: If I understand correctly, `originalWitnessStates` will be full if the block is at window starts (even the overlapped ones), and only compact in other blocks? Otherwise, if the witness is compact, the stateless node might not be able to execute correctly in a later block in the window. 

## Example

### Trie node usage
Trie nodes used by block [1-20]:
```
{
    "0x1", # trie node used by block 1
    "0x2", # trie node used by block 2
}
```

Trie node used by block [11-30]:
```
{
    "0x1", # trie node used by block 11 and 30
    "0x3", # trie node used by block 11
}
```

### Witness

The compact witness for block 11 would be:
```
{
    "0x3", # new trie node used by block 11
}
```

because `0x1` is trimmed off due to its presence in the first window.

The full witness for block 11 would be:
```
{
    "0x1", # trie node used by block 11 and 30
    "0x3", # trie node used by block 11
}
```

if `0x1` isn't included in the second cache window [11-30], which is the compact witness of block 11, execution of block 30 will fail. 

## Alternative
I think an alternative solution is to force the stateless node to compute the full witness during the block execution, and use that as the original full witness. This will allow nodes to receive only compact witness for any block (except the first one) but still allow the cache window to shift at the same time.

eth/handler_eth.go
```diff
@@ -118,26 +118,62 @@ func (h *ethHandler) handleBlockAnnounces(peer *eth.Peer, hashes []common.Hash,
...
+
+	// If block is at window start, need full witness (this is first block of window)
+	// The sender will have just cleared/slid their cache at this block
+	if blockNum == expectedWindowStart {
+		return false
+	}
+
+	// Block is within window but not at start, can use compact witness
+	// The sender should have cached states from earlier blocks in this window
+	return true
```
cffls, line 207: The first block of the overlapped window is also a start block. If [this](https://github.com/0xPolygon/bor/pull/1818/files#r2487963279) is true, the node should request full witness every 10 blocks instead of 20 (assuming 20-block cache window and 10-block overlap).

eth/handler_eth.go
```diff
@@ -118,26 +118,71 @@ func (h *ethHandler) handleBlockAnnounces(peer *eth.Peer, hashes []common.Hash,
...
+	expectedWindowStart := (blockNum / windowSize) * windowSize
+
+	// Calculate position within the window
+	blocksSinceWindowStart := blockNum - expectedWindowStart
+
+	// Need full witness at:
+	// 1. Window start (block 0, 20, 40...) - new window begins
+	// 2. Overlap start (block 10, 30, 50...) - pre-warming next window
+	// This ensures nextCacheMap has complete state coverage for the next window
+	if blockNum == expectedWindowStart || blocksSinceWindowStart == overlapSize {
```
cffls: Regarding this check `blocksSinceWindowStart == overlapSize`, shouldn't it be `blocksUntilNextWindowStart == overlapSize`, where `blocksUntilNextWindowStart := expectedWindowStart+windowSize-overlapSize`?

It would be great to create a unit test that will fail otherwise without this check.

eth/handler_wit.go
```diff
@@ -182,6 +203,161 @@ func (h *witHandler) handleGetWitness(peer *wit.Peer, req *wit.GetWitnessPacket)
...
+	// Return full witness instead of trying to filter
+	if h.chain.IsParallelStatelessImportEnabled() {
+		log.Debug("Parallel import enabled, returning full witness instead of compact",
+			"peer", peer.ID(),
+			"reqID", req.RequestId)
+		return fullResponse, nil
+	}
+
+	// Get current cache window start for cache key
+	cacheWindowStart := h.chain.GetCacheWindowStart()
```
cffls: It seems to me this should be a deterministic number based on the requested block number?

core/blockchain.go
```diff
@@ -4151,12 +4184,248 @@ func (bc *BlockChain) SubscribeChain2HeadEvent(ch chan<- Chain2HeadEvent) event.
...
+	bc.cacheLock.RLock()
+	defer bc.cacheLock.RUnlock()
+
+	originalStates := len(filtered.State)
+	filteredStates := make(map[string]struct{})
+	removed := 0
+
+	for stateNode := range filtered.State {
+		// Only include states NOT in active cache
+		if _, cached := bc.activeCacheMap[stateNode]; !cached {
```
cffls, line 4314: This logic doesn't seem to be deterministic. The activeCacheMap will change overtime. We might need to directly save compact witness in db rather than computing it on the fly.

eth/handler_eth.go
pratikspatil024: Nice catch, thanks! Fixed [here](https://github.com/0xPolygon/bor/pull/1818/commits/dade99746daac29ecd34f314c16fe4c0502b399f).

core/blockchain.go
pratikspatil024: [Done](https://github.com/0xPolygon/bor/pull/1818/commits/598a9dd7a41f08be58d21618540d4ec15687bd12)

eth/handler_wit.go
```diff
@@ -182,6 +203,222 @@ func (h *witHandler) handleGetWitness(peer *wit.Peer, req *wit.GetWitnessPacket)
...
+		}
+
+		witness, err := stateless.GetWitnessFromRlp(witnessPage.Data)
+		if err != nil {
+			log.Warn("Failed to decode witness for filtering", "hash", witnessPage.Hash, "err", err)
+			filteredResponse = append(filteredResponse, witnessPage)
+			continue
+		}
+
+		filteredWitness := h.filterWitnessWithCache(witness)
```
cffls: We don't want to introduce any wrong compact witness here. If the witness should be compact but we only have full in db, just return full. wdyt?

pratikspatil024: nice idea, thanks. updated [here](https://github.com/0xPolygon/bor/pull/1818/commits/8b0aaeb9eff3ffab110ccb725d01a0a929990750).

PR #1819, opened 2025-10-13
feat: TestDecodeErigonStateSyncReceipt

kamuikatsurgi:
# Description

`TestDecodeErigonStateSyncReceipt` verifies that Bor can decode state sync receipts sent by Erigon nodes over P2P using `eth/69` protocol. The RLP data is captured from actual Erigon state sync receipt generation.

### DIFF
--- a/core/types/receipt_test.go
+++ b/core/types/receipt_test.go
@@ -575,3 +575,59 @@ func clearComputedFieldsOnLogs(logs []*Log) []*Log {
 
 	return l
 }
+
+// TestDecodeErigonStateSyncReceipt verifies that Bor can decode state sync receipts
+// sent by Erigon nodes over P2P using eth/69 protocol.
+// The RLP data is captured from actual Erigon state sync receipt generation.
+func TestDecodeErigonStateSyncReceipt(t *testing.T) {
+	// This is the eth/69 RLP encoding of a state sync receipt from Erigon.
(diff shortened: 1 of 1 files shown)


core/types/receipt_test.go
```diff
@@ -575,3 +575,59 @@ func clearComputedFieldsOnLogs(logs []*Log) []*Log {
...
+
+	// Decode the hex string.
+	rlpBytes := common.FromHex(eth69RLP)
+
+	if len(rlpBytes) == 0 {
+		t.Fatal("Empty RLP bytes")
+	}
+
+	var data storedReceiptRLP
```
manav2401, line 593: It would be good if we can decode based on how it's done [here](https://github.com/0xPolygon/bor/blob/develop/eth/protocols/eth/handlers.go#L530) using ReceiptList where you can explicitly use `ReceiptList69` struct instead of the generic parameter `L`.

core/types/receipt_test.go
```diff
@@ -575,3 +575,59 @@ func clearComputedFieldsOnLogs(logs []*Log) []*Log {
...
+	err := rlp.DecodeBytes(rlpBytes, &data)
+	if err != nil {
+		t.Fatalf("Failed to decode eth/69 receipt: %v", err)
+	}
+
+	receiptType := LegacyTxType
+
+	// Verify it's a state sync receipt (CumulativeGasUsed = 0).
+	if data.CumulativeGasUsed != 0 {
```
manav2401, line 602: This is only the case if block only contained state-sync receipt and nothing else. It would be good if we can include a case which has normal receipts + state-sync receipts. In that case, the cumulative gas field in state-sync receipt would be equal to the value in last normal receipt.

PR #1821, opened 2025-10-14
Fix self destruct in parallel evm

avalkov:
# Description

When destructing contract, check if it exists in the actual state db you want to destroy it in.


avalkov:
> If possible, could you add a unit test which will fail without this fix but pass with the fix?

Added unit test.

lucca30 (approved):

PR #1828, opened 2025-10-16
Stateless Pruning

lucca30:
# Description

- Refactor Witness Pruning to be on `rawdb` package (same place as freezer)
- Generic Prune Solution both for Witness and Blocks
- Disabled Freezing for Stateless clients
- Improved a few scenarios
  - Chunk delete
  - find earliest block to delete on cursor

# Changes

### DIFF
--- a/core/rawdb/schema.go
+++ b/core/rawdb/schema.go
@@ -120,7 +120,7 @@ var (
 
 	blockBodyPrefix     = []byte("b") // blockBodyPrefix + num (uint64 big endian) + hash -> block body
 	blockReceiptsPrefix = []byte("r") // blockReceiptsPrefix + num (uint64 big endian) + hash -> block receipts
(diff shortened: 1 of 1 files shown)


core/rawdb/pruner_strategy_wit.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethdb"
+)
+
+const (
+	WitnessRetentionBlocks = 6400              // Minimum necessary distance between local header and latest non pruned witness
+	WitnessPruneInterval   = 120 * time.Second // The time interval between each witness prune routine
```
cffls: Should we make these two parameter configurable?

cffls (approved):

PR #1838, opened 2025-10-24
chore: nits

kamuikatsurgi:


kamuikatsurgi:
> whats the reasoning behind this?
> 
> If `len(stateSyncData) > 0`, does this not imply `len(body.Transactions) > 0`?

Just an additional safety check, doesn't hurt to keep it there. Just in case a block received by a peer doesn't have state-sync tx appended, even though there is some state sync data, we need to have this check.

pratikspatil024:
> Just in case a block received by a peer doesn't have state-sync tx appended, even though there is some state sync data, we need to have this check.

Right, but the txn you are using is extracted from the body. And `stateSyncData` is not used. You can have that additional check, but better to use `stateSyncData` to get the `stateSyncTx` for uniformity.

kamuikatsurgi:
> > Just in case a block received by a peer doesn't have state-sync tx appended, even though there is some state sync data, we need to have this check.
> 
> Right, but the txn you are using is extracted from the body. And `stateSyncData` is not used. You can have that additional check, but better to use `stateSyncData` to get the `stateSyncTx` for uniformity.

So here we are getting a block from the peer, and hence it's better to get state sync tx from the body instead of data because we want to catch invalid receiptHash and bloom errors early on and reject the block and drop the peer.

PR #1839, opened 2025-10-26
geth upstream v1.16.4

marcello33:
# Description

Merge geth upstream v1.16.4.

Block height for Osaka related code to be enabled and activated (as previously done for Madhugiri), in case we decide to support the remaining EIPs. The only one left requiring a HF is EIP-7939 

# Changes

# Checklist

## Testing

- [x] I have added unit tests


core/blockchain.go
```diff
@@ -2245,18 +2265,17 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
...
 	// Commit all cached state changes into underlying memory database.
-
-	var root common.Hash
-	var err error
-	if statedb.Database() != nil {
-		root, err = statedb.Commit(block.NumberU64(), bc.chainConfig.IsEIP158(block.Number()), bc.chainConfig.IsCancun(block.Number()))
-		if err != nil {
-			return []*types.Log{}, err
-		}
+	root, stateUpdate, err := statedb.CommitWithUpdate(block.NumberU64(), bc.chainConfig.IsEIP158(block.Number()), bc.chainConfig.IsCancun(block.Number()))
```
lucca30, line 2274: cc @pratikspatil024 just to show they implemented the method we did for stateless caching

core/stateless/witness.go
```diff
@@ -126,14 +126,20 @@ func (w *Witness) AddCode(code []byte) {
 }
 
 // AddState inserts a batch of MPT trie nodes into the witness.
-func (w *Witness) AddState(nodes map[string]struct{}) {
+func (w *Witness) AddState(nodes map[string][]byte) {
```
lucca30, line 129: Any major reason to include value on the map? The key itself is already the `string(value)`. Which basically makes the map doubles the size: storing the bytes in both `string` (as key) and `[]byte` (as value) format.

You see the only time we use the value of map is the for loop to convert back to `string`.

```
	for _, value := range nodes {
		w.State[string(value)] = struct{}{}
	}
```
- `string(value)` is the same thing as the key

The original code is just a copy, faster and more straightforward.

core/stateless/encoding.go
```diff
@@ -19,30 +19,40 @@ package stateless
...
 // fromExtWitness converts the consensus witness format into our internal one.
-func (w *Witness) fromExtWitness(ext *extWitness) error {
+func (w *Witness) fromExtWitness(ext *ExtWitness) error {
 	w.context = ext.Context
 	w.Headers = ext.Headers
+
+	w.Codes = make(map[string]struct{}, len(ext.Codes))
+	for _, code := range ext.Codes {
+		w.Codes[string(code)] = struct{}{}
+	}
```
lucca30, line 55: We don't use code on our witness. Is ok to remove.

core/stateless/encoding.go
```diff
@@ -52,21 +62,23 @@ func (w *Witness) fromExtWitness(ext *extWitness) error {
...
 
-// extWitness is a witness RLP encoding for transferring across clients.
-type extWitness struct {
+// ExtWitness is a witness RLP encoding for transferring across clients.
+type ExtWitness struct {
 	Context *types.Header
-	Headers []*types.Header
-	State   [][]byte
+	Headers []*types.Header `json:"headers"`
+	Codes   []hexutil.Bytes `json:"codes"`
```
lucca30, line 81: Ok to remove Codes

core/stateless/encoding.go
```diff
@@ -52,21 +62,23 @@ func (w *Witness) fromExtWitness(ext *extWitness) error {
...
+	State   []hexutil.Bytes `json:"state"`
+	Keys    []hexutil.Bytes `json:"keys"`
```
lucca30, line 83: Keys is a new param which is not used anywhere. The only reference goes to a [`panic("not implemented")`](https://github.com/0xPolygon/bor/blob/8fb990fb215df402740e7073234263563eb70293/core/stateless/witness.go#L141-L143)

core/stateless/stats.go
```diff
@@ -0,0 +1,92 @@
...
+		// The last path is always a leaf.
+		if i == len(paths)-1 || !strings.HasPrefix(paths[i+1], paths[i]) {
+			if owner == (common.Hash{}) {
+				s.accountTrieLeaves[len(path)] += 1
+			} else {
+				s.storageTrieLeaves[len(path)] += 1
+			}
+		}
+	}
+}
```
lucca30, line 72: This is one of the places you will need to change to keep `map[string]struct{}`

```
	for i, _ := range paths {
        path := []byte(i)
		// If current path is a prefix of the next path, it's not a leaf.
		// The last path is always a leaf.
		if i == len(paths)-1 || !strings.HasPrefix(paths[i+1], paths[i]) {
			if owner == (common.Hash{}) {
				s.accountTrieLeaves[len(path)] += 1
			} else {
				s.storageTrieLeaves[len(path)] += 1
			}
		}
	}

```

core/stateless/witness.go
marcello33: So this comes from [here](https://github.com/ethereum/go-ethereum/commit/e9656238a7f20d8d797bb957e9c58f3be73993a4). The original code was also from upstream, so I decided to keep the changes for compatibility, mostly. I see the biggest simplification this introduces is due to `WitnessStats`, which I don't explicitly use in this PR (passing `nil` whenever stats are required), since I wanted to keep our own implementation of witnesses, generally speaking.
You think it's worth reverting to `struct{}` and diverge from upstream, although the result is the same - except for the map's size?

core/stateless/encoding.go
marcello33: I agree, wanted to remove this in the first place, then I checked the upstream PR in which it was added: https://github.com/ethereum/go-ethereum/pull/32216.
They mention "the `keys` field is present but is always empty - this is an agreement with Kev, that will be done in a later PR. The role of this PR is to find the disagreements with reth."
So there's apparently a plan of using `keys` later on. However, you guys have implemented the witness feature, so if you consider it's better to remove it compromising future compatibility with upstream, I can delete that extra field ofc

core/stateless/encoding.go
marcello33: But it's harmless to keep, right? If we don't use it, the allocation would be empty, and the `range` of `ext.Codes` would be zero, leading to skip the `for` loop. If you confirm that, I'd keep the code for upstream compatibility and avoid breaking potential tests depending on this (from upstream).

core/stateless/stats.go
marcello33: Yes, thanks for the reminder. Let's decide [here](https://github.com/0xPolygon/bor/pull/1839#discussion_r2502076755) about reverting it or not, and I can implement the "cascade effects" everywhere based on that.

core/stateless/encoding.go
marcello33: Ok waiting on [this](https://github.com/0xPolygon/bor/pull/1839#discussion_r2502108553) to decide 🙏

miner/worker.go
```diff
@@ -162,13 +163,23 @@ type task struct {
 	createdAt time.Time
 }
 
+// txFits reports whether the transaction fits into the block size limit.
+func (env *environment) txFitsSize(tx *types.Transaction) bool {
+	return env.size+tx.Size() < params.MaxBlockSize-maxBlockSizeBufferZone
```
kamuikatsurgi, line 197: we don' have `maxBlockSizeBufferZone`. Make `maxBlockSizeBufferZone = 0` in `const maxBlockSizeBufferZone = 1_000_000`

marcello33: Done, thanks!

core/stateless/witness.go
lucca30: My concern on changing the way we store it is the risk of incompatibility on the WIT protocol. Ethereum has no protocol associated with witness, so they probably are not cautious on changing the struct. In our solution, will it break compatibility? If so would be necessary creating a new protocol version

cmd/utils/flags.go
```diff
@@ -624,6 +654,12 @@ var (
 		Value:    ethconfig.Defaults.RPCTxFeeCap,
 		Category: flags.APICategory,
 	}
+	RPCGlobalLogQueryLimit = &cli.IntFlag{
+		Name:     "rpc.logquerylimit",
```
manav2401, line 662: We'll have to port this flag to `internal/cli`.

marcello33: Right, thanks. Addressed [here](https://github.com/0xPolygon/bor/pull/1839/changes/26f83d304507e3b148c411a58efdfeb9d3f6af8d)

marcello33:
> ## Code Review
> ### Issue Found: Incorrect fork ordering in `LatestFork` function
> **File:** [params/config.go](https://github.com/0xPolygon/bor/blob/0e0c78101357c5e9ff85e5545bf5808d19de7338/params/config.go#L1552-L1564)
> 
> **Lines:** 1552-1564
> 
> **Description:** The switch statement checks forks in the wrong order. Since Osaka (fork 42) is activated after Prague (fork 41), when both forks are active, the `IsPrague` check will match first and return `forks.Prague` instead of `forks.Osaka`.
> 
> According to [params/forks/forks.go](https://github.com/0xPolygon/bor/blob/0e0c78101357c5e9ff85e5545bf5808d19de7338/params/forks/forks.go#L41-L42):
> 
> ```go
> Prague  // 41
> Osaka   // 42
> ```
> 
> **Current code (incorrect):**
> 
> ```go
> switch {
> case c.IsPrague(london):
>     return forks.Prague
> case c.IsOsaka(london):
>     return forks.Osaka
> ```
> 
> **Suggested fix:** Check the newest fork first (Osaka), then progressively older ones:
> 
> ```go
> switch {
> case c.IsOsaka(london):
>     return forks.Osaka
> case c.IsPrague(london):
>     return forks.Prague
> ```
> 
> **Impact:** This bug would affect any code that relies on `LatestFork` to correctly identify when Osaka is active, such as `ActiveSystemContracts` and engine API versioning.
> 
> **Note:** Since production configs (BorMainnet, Amoy, Mumbai) have `OsakaBlock = nil`, this bug currently doesn't manifest in production. However, it affects test configurations (like `AllDevChainConfig`) where Osaka is enabled.
> 
> Checked for bugs and CLAUDE.md compliance.

fixed

marcello33:
@claude Can you review this PR, focusing on safety and performance? Provide a summary as a comment highlighting any concerns related to these changes.

marcello33:
@claude
1. Binary Trie Implementation (trie/bintrie/) > this comes from upstream and will be tested before enabling it
2. Blob Transaction V0→V1 Conversion (Osaka Fork) > we don't use blobs in bor, hence we should be good
3. PathDB History Refactoring (triedb/pathdb/history.go) > ported from upstream
4. Witness Format Change > ported from upstream. Received green light from @lucca30 
5. Future Blocks Removal (core/blockchain.go) > this is ok
6. Precompile Interface Change > this is handled in the PR
7. Missing CLI Flag Port > Ported also `TrieJournalDirectory` and `EnableStateSizeTracking`

consensus/bor/bor.go
```diff
@@ -1195,11 +1196,10 @@ func (c *Bor) changeContractCodeIfNeeded(headerNumber uint64, state vm.StateDB)
 
 			for addr, account := range allocs {
 				log.Info("change contract code", "address", addr)
-				state.SetCode(addr, account.Code)
+				state.SetCode(addr, account.Code, tracing.CodeChangeContractCreation)
```
kamuikatsurgi: In `SetBalance`, we are using the `BalanceChangeUnspecified`, but here we are using `CodeChangeContractCreation`. This function could change the contract code at any time, not just during deployment. Should we use `CodeChangeUnspecified`? Just asking, don't have much info, but there seems to be a mismatch in tracing reasons.

kamuikatsurgi (approved):
LGTM, left a que, and also would be great if we could enable witness stats, seems like something we could benefit in the long run, wdyt?

consensus/bor/bor.go
manav2401: Agree, we should keep it unspecified. The `CodeChangeContractCreation` is only for deployment via CREATE/CREATE2 which is not the case here.

manav2401: If `CodeChangeGenesis` fits, we could also use that actually. 

Comment says
```
// CodeChangeGenesis is when contract code is set during blockchain genesis or initial setup.
```

kamuikatsurgi: But looking at other places in the code, `CodeChangeUnspecified` is used. If we go ahead with `CodeChangeGenesis`, then we should do it for balance as well: `BalanceChangeGenesis`. But I think `CodeChangeUnspecified` would be better.

marcello33: Addressed, thanks.

marcello33:
> LGTM, left a que, and also would be great if we could enable witness stats, seems like something we could benefit in the long run, wdyt?

I agree. I'd just won't do it straight away. Maybe we can release this and enable it on some nodes in amoy to test it out.

PR #1842, opened 2025-10-27
core, miner, params, cmd: implement EIP-7823, EIP-7825 and EIP-7883

marcello33:
# Description

Implement `EIP-7823`, `EIP-7825` and `EIP-7883` from geth upstream into `Madhugiri` HF

# Changes

# Checklist

### Manual tests

Yet to be tested when triggering the `Madhugiri` HF (e.g. on a `kurtosis` env)


core/txpool/legacypool/legacypool.go
```diff
@@ -589,11 +589,19 @@ func (pool *LegacyPool) Pending(filter txpool.PendingFilter, interrupt *atomic.B
...
-					txs = txs[:i]
-					break
+				if minTipBig != nil {
+					if tx.EffectiveGasTipIntCmp(minTipBig, baseFeeBig) < 0 {
+						txs = txs[:i]
+						break
+					}
+				}
+				if filter.GasLimitCap != 0 {
+					if tx.Gas() > filter.GasLimitCap {
```
cffls, line 601: Is it possible to prevent such transaction from being added to the pool in the first place?

marcello33: I believe so, if not done yet, but I need to double check. However, with additional changes we would diverge from what geth is doing upstream

cffls: Thanks for the context. It seems strange why geth would allow such transactions to be included in txpool. Isn't legacy pool deprecated in geth btw?

marcello33: As per my understanding, the legacypool refers to the transaction types it accepts (legacy types 0/1/2 vs. blobs), so it should not be deprecated.
These changes reflect the upstream PRs for such EIP (see https://github.com/ethereum/go-ethereum/pull/31824 and https://github.com/ethereum/go-ethereum/pull/32230).
In the relative EIP they mention "Txpool Validation: During transaction validation, if the gasLimit specified by the sender exceeds 16,777,216, the transaction is invalidated (not included in the txpool).
Block Validation: As part of block validation before processing, any block having a transaction with gasLimit > 16,777,216 is deemed invalid and rejected." (see https://eips.ethereum.org/EIPS/eip-7825).
So bor should behave exactly the same with these changes, no?
Please, do let me know if you foresee any issue.

cmd/evm/internal/t8ntool/transaction.go
```diff
@@ -202,6 +202,15 @@ func Transaction(ctx *cli.Context) error {
 			r.Error = errors.New("max initcode size exceeded")
 		}
 
+		// bor: apply EIP-7825 at Madhugiri block
+		isOsaka := chainConfig.IsOsaka(new(big.Int))
+		isMadhugiri := chainConfig.Bor != nil && chainConfig.Bor.IsMadhugiri(new(big.Int))
+
+		// Verify tx gas limit does not exceed EIP-7825 cap.
+		if (isOsaka || isMadhugiri) && tx.Gas() > params.MaxTxGas {
+			r.Error = errors.New("gas limit exceeds maximum")
```
kamuikatsurgi, line 211: `r.Error = errors.New("gas limit exceeds maximum")` -> `r.Error = errors.New("tx gas limit exceeds maximum")`?

cmd/evm/internal/t8ntool/transaction.go
```diff
@@ -202,6 +202,15 @@ func Transaction(ctx *cli.Context) error {
 			r.Error = errors.New("max initcode size exceeded")
 		}
 
+		// bor: apply EIP-7825 at Madhugiri block
```
kamuikatsurgi:
```go
// Bor: EIP-7825 at Madhugiri HF block
```

core/txpool/legacypool/legacypool.go
```diff
@@ -1377,6 +1385,24 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest,
 	lockTime := time.Now()
 	pool.mu.Lock()
 	if reset != nil {
+		if reset.newHead != nil && reset.oldHead != nil {
+			// bor: apply EIP-7825 at Madhugiri block
```
kamuikatsurgi:
```go
// Bor: EIP-7825 at Madhugiri HF block
```

core/txpool/validation.go
```diff
@@ -92,6 +92,10 @@ func ValidateTransaction(tx *types.Transaction, head *types.Header, signer types
 	if rules.IsShanghai && tx.To() == nil && len(tx.Data()) > params.MaxInitCodeSize {
 		return fmt.Errorf("%w: code size %v, limit %v", core.ErrMaxInitCodeSizeExceeded, len(tx.Data()), params.MaxInitCodeSize)
 	}
+	// bor: apply EIP-7825 at Madhugiri block
```
kamuikatsurgi:
```go
// Bor: EIP-7825 at Madhugiri HF block
```

core/txpool/legacypool/legacypool.go
kamuikatsurgi: You can see in the txpool code that the tx hashes that have a greater than max tx gas limit, we store their hashes:
```go
if tx.Gas() > params.MaxTxGas {
    hashes = append(hashes, hash)
}
```

After that, we remove those txs from the pool like this:
```go
for _, hash := range hashes {
    pool.removeTx(hash, true, true)
}
```

cmd/evm/internal/t8ntool/transaction.go
marcello33: Applied everywhere, thanks

core/txpool/validation.go
marcello33: Applied everywhere, thanks

core/txpool/legacypool/legacypool.go
marcello33: Applied everywhere, thanks

cmd/evm/internal/t8ntool/transaction.go
marcello33: Done, thanks

consensus/bor/contract/client.go
```diff
@@ -118,7 +118,15 @@ func (gc *GenesisContractsClient) LastStateId(state *state.StateDB, number uint6
...
-	gas := (hexutil.Uint64)(uint64(math.MaxUint64 / 2))
+	var gas hexutil.Uint64
+
+	// Bor: EIP-7825 at Madhugiri HF block
+	IsMadhugiri := gc.chainConfig.Bor != nil && gc.chainConfig.Bor.IsMadhugiri(big.NewInt(int64(number)))
+	if IsMadhugiri {
+		gas = (hexutil.Uint64)(params.MaxTxGas / 2)
+	} else {
+		gas = (hexutil.Uint64)(math.MaxUint64 / 2)
+	}
```
kamuikatsurgi, line 129: Do you think this is the correct approach here?

MaxUint64 / 2 == (2^64 - 1) / 2, which is almost equal to 2^63

But MaxTxGas / 2 == 2^24 / 2, which is equal to 2^23.

So we'd be reducing the gas of state sync txs by 2^40.

cc @cffls

marcello33: Thanks for bringing this up, it is smth I wanted to discuss w/ the team anyway.
Basically, we used to have - as you said - `math.MaxUint64 / 2`
Which used to be the old "cap" divided by two.
That will no longer work. 
Hence I decided to keep the "half the cap" approach, by setting it to `params.MaxTxGas / 2`
Alternatively, it could be "as big as `params.MaxTxGas`", but surely not anything near `math.MaxUint64`

This was brought to my attention by the failure of the integration test `TestFetchStateSyncEvents_PostMadhugiriHF `

core/txpool/legacypool/legacypool.go
cffls: > You can see in the txpool code that the tx hashes that have a greater than max tx gas limit, we store their hashes

That's true. This shows the tx hash was added to the pool in the first place? 

> Txpool Validation: During transaction validation, if the gasLimit specified by the sender exceeds 16,777,216, the transaction is invalidated (not included in the txpool).

It looks like the correct way to implement this is to have this check in `add` https://github.com/0xPolygon/bor/blob/develop/core/txpool/legacypool/legacypool.go#L750?.

kamuikatsurgi: >That's true. This shows the tx hash was added to the pool in the first place?

Added and then removed, also no. If we straight-up reject its addition in mempool, it wouldn't make sense, right?

marcello33: Uhm not sure @cffls 
I can also duplicate that check ofc in the `add` method, but I think the upstream implementation (which we are currently mirroring in bor) is sufficient.
I'd tend to agree w/ @kamuikatsurgi

cmd/evm/internal/t8ntool/transaction.go
```diff
@@ -202,6 +202,15 @@ func Transaction(ctx *cli.Context) error {
 			r.Error = errors.New("max initcode size exceeded")
 		}
 
+		// Bor: EIP-7825 at Madhugiri HF block
+		isOsaka := chainConfig.IsOsaka(new(big.Int))
+		isMadhugiri := chainConfig.Bor != nil && chainConfig.Bor.IsMadhugiri(new(big.Int))
+
+		// Verify tx gas limit does not exceed EIP-7825 cap.
+		if (isOsaka || isMadhugiri) && tx.Gas() > params.MaxTxGas {
+			r.Error = errors.New("tx gas limit exceeds maximum")
```
kamuikatsurgi: `r.Error = errors.New("gas limit exceeds maximum")`

Sorry, we should not divert from L1 here. My bad.

marcello33: Reverted ;)

core/txpool/legacypool/legacypool.go
cffls: I think if we check it in `add`, we wouldn't need to check in pending and reorg. but I am fine with keeping this change as it will be the same as geth implementation.

consensus/bor/contract/client.go
manav2401: Addressed in https://github.com/0xPolygon/bor/pull/1842/commits/3cec7835103ae1a04281d5d169724eda0cd77469

consensus/bor/statefull/processor.go
```diff
@@ -55,7 +54,7 @@ func GetSystemMessage(toAddress common.Address, data []byte) Callmsg {
 	return Callmsg{
 		ethereum.CallMsg{
 			From:     systemAddress,
-			Gas:      math.MaxUint64 / 2,
+			Gas:      params.MaxTxGas, // should be more than enough for state-sync related syscalls
```
kamuikatsurgi, line 57: The gas should be sufficient; I'm just concerned about potential sync issues for historical blocks. Should we determine gas based on the block number?

manav2401: I think it should be fine because of the 5M cap in contract.

PR #1846, opened 2025-10-28
Fix bad block when tx is reverted in parallel EVM

avalkov:
# Description

When tx is reverted in parallel EVM execution, in the final state db we should write only sender new balance and nonce.

### DIFF
--- a/core/parallel_state_processor_test.go
+++ b/core/parallel_state_processor_test.go
@@ -1,9 +1,21 @@
 package core
 
 import (
+	"math/big"
 	"testing"
 
+	"github.com/holiman/uint256"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/blockstm"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/tracing"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/triedb"
 )
 
 func TestMetadata(t *testing.T) {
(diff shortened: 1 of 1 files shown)


core/state/statedb.go
```diff
@@ -400,6 +400,35 @@ func (s *StateDB) FlushMVWriteSet() {
...
+					s.SetBalance(addr, sr.GetBalance(addr), tracing.BalanceChangeUnspecified)
+				}
+			case NoncePath:
+				if addr == sender {
+					s.SetNonce(addr, sr.GetNonce(addr), tracing.NonceChangeUnspecified)
+				}
+			case CodePath:
+			case SuicidePath:
+			default:
+				panic(fmt.Errorf("unknown key type: %d", path.GetSubpath()))
```
kamuikatsurgi, line 426: good idea to panic here?

kamuikatsurgi (approved):

core/state/statedb.go
kamuikatsurgi: I know it's not possible but just a q

avalkov: Depends how you look at it, I think @cffls said that we gonna keep running serial and parallel always together so in such case, we can remove the panic. But if we ever plan to force just parallel, better to panic in my opinion because you reached some unexpected state.

cffls: Panic will be recovered in parallel_state_processor. I am fine with this.

core/parallel_state_processor.go
```diff
@@ -182,10 +182,14 @@ func (task *ExecutionTask) Settle() {
 
 	coinbaseBalance := task.finalStateDB.GetBalance(task.coinbase)
 
-	task.finalStateDB.ApplyMVWriteSet(task.statedb.MVFullWriteList())
+	if task.result.Failed() {
+		task.finalStateDB.ApplyFailedTxMVWriteSet(task.statedb.MVFullWriteList(), task.sender)
```
cffls, line 186: Special handling for failed transaction wasn't there before because the expectation was that only nonce and balance change will have different values that need to be applied. It might be a good idea to understand why other values, e.g. code, storage, are written to task.statedb even when the transaction failed.

PR #1867, opened 2025-11-04
chore: set madhugiri block for amoy and consensus block time

marcello33:
# Description

Set Madhugiri block for amoy


marcello33:
`govulncheck` error can be ignored. A fix is already started in other branches.

PR #1871, opened 2025-11-06
Add prefetch trie nodes

Raneet10:
# Description

This PR enables prefetching of trie nodes before execution of block. It adds a  `waitforwarm` config which if enabled will block execution of block until the warming finishes. 

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1698,9 +1698,9 @@ func (w *worker) fillTransactions(interrupt *atomic.Int32, env *environment) err
 			}
 
 			// start warm reader cache (group txs to avoid duplicate contract per group)
-			if len(selected) > 0 {
-				w.groupAndWarm(selected, env, w.chain.WaitForWarmEnabled(), warmedAttempted, warmedCompleted)
-			}
+			// if len(selected) > 0 {
+			// 	w.groupAndWarm(selected, env, w.chain.WaitForWarmEnabled(), warmedAttempted, warmedCompleted)
+			// }
 
 			// warming for remaining cold contract txs
 			if len(selectedSet) > 0 && len(txs) > 0 {
@@ -1820,6 +1820,7 @@ func (w *worker) groupAndWarm(txs []*types.Transaction, env *environment, wait b
 	numProcs := runtime.GOMAXPROCS(8)
 	sem := make(chan struct{}, numProcs)
 	var wg sync.WaitGroup
+	var completedMu sync.Mutex
 
 	for _, group := range groups {
 		if w.warmInterrupted() {
@@ -1833,9 +1834,11 @@ func (w *worker) groupAndWarm(txs []*types.Transaction, env *environment, wait b
 			defer wg.Done()
 			defer func() { <-sem }()
 			if _, err := blockstm.ExecuteParallel(tasks, false, false, numProcs, ctx); err == nil || ctx.Err() == nil {
+				completedMu.Lock()
 				for _, tx := range groupTxs {
 					warmedCompleted[tx.Hash()] = struct{}{}
 				}
+				completedMu.Unlock()
 
 			}
 		}(group, tasks)


core/txpool/legacypool/legacypool.go
```diff
@@ -2116,3 +2119,17 @@ func (pool *LegacyPool) isFiltered(addr common.Address) bool {
...
 }
+
+// PreLoadTrieNodes warms relevant trie nodes and code for transactions concurrently
+// then sets a single global prefetch reader.
+func (pool *LegacyPool) PreLoadTrieNodes(txs ...*types.Transaction) {
+	pool.mu.RLock()
+	header := pool.currentHead.Load()
+	pool.mu.RUnlock()
+
+	if header == nil || len(txs) == 0 {
```
marcello33: Shall we move the txs len check on top of the function?
Just an optimization to reduce lock contention. Since txs is an input and doesn’t depend on pool state, you can fast-return before touching any shared data.

core/blockchain.go
marcello33: Shall we remove all the unused metrics here? Lint's complaining

marcello33, line 687: is linter right here?

core/blockchain.go
```diff
@@ -708,29 +737,28 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header, wit
...
+		bc.lastProcStorHit += stats.StorageHit
+		bc.lastProcStorMiss += stats.StorageMiss
 	}()
 
-	go func(start time.Time, throwaway *state.StateDB, block *types.Block) {
-		// Disable tracing for prefetcher executions.
-		vmCfg := bc.cfg.VmConfig
-		vmCfg.Tracer = nil
-		bc.prefetcher.Prefetch(block, throwaway, vmCfg, followupInterrupt)
+	// go func(start time.Time, throwaway *state.StateDB, block *types.Block) {
```
marcello33, line 755: Is this worth deleting?
Or - alternatively - have a comment to explain why it's disabled

core/blockchain.go
```diff
@@ -822,6 +850,23 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header, wit
 	return result.receipts, result.logs, result.usedGas, result.statedb, vtime, result.err
 }
 
+// warmReaderCache preloads the supplied readerWithCache with accounts/storage and trie nodes.
+func (bc *BlockChain) warmReaderCache(parentRoot common.Hash, reader state.ReaderWithStats, b *types.Block, followupInterrupt *atomic.Bool) (state.ReaderWithStats, time.Duration) {
+	if reader == nil || b == nil || len(b.Transactions()) == 0 {
+		return nil, 0
+	}
+	vmCfg := bc.cfg.VmConfig
+	vmCfg.Tracer = nil
```
marcello33, line 870: why is this required? Same for `PrefetchFromTxpool`

core/blockchain_insert.go
```diff
@@ -70,6 +98,35 @@ func (st *insertStats) report(chain []*types.Block, index int, snapDiffItems, sn
 				context = append(context, []interface{}{"snapdirty", snapBufItems}...)
 			}
 		}
+		// Append cache hit/miss percentages
+		if bc != nil {
+			if total := bc.lastProcAccHit + bc.lastProcAccMiss; total > 0 {
+				pct := int64(100 * bc.lastProcAccHit / total)
```
marcello33, line 104: seems like this (and others in the same func) conversion is pleonastic

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/ethereum/go-ethereum
 
 // Note: Change the go image version in Dockerfile if you change this.
-go 1.24.6
+go 1.24.8
```
marcello33: worth updating in the CI yml fles as well.
Also, since we're bumping, maybe we can use 1.24.9 to solve the govulncheck issues (I have that on my upstream PR as well)

core/analyzer.sh
```diff
@@ -0,0 +1,102 @@
+#   ./log_analyzer.sh "3 days ago" # Scans the last 3 days
```
cffls: This should be excluded from this PR I guess?

core/state_prefetcher.go
```diff
@@ -111,18 +111,21 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
+			// Aggregate touched pre-state from stateCpy and populate the process reader cache directly
+			acc, stor := stateCpy.PrefetchTouched()
+			for a := range acc {
+				reader.Account(a)
+			}
+			for a, bucket := range stor {
+				for k := range bucket {
+					reader.Storage(a, k)
+				}
+			}
```
cffls: What's the rationale of this change? `IntermediateRoot` also fetches touched account/storages.

core/blockchain.go
```diff
@@ -4297,3 +4359,61 @@ func (bc *BlockChain) verifyPendingHeaders() {
...
+	go func() {
+		warmdb, err := state.New(header.Root, bc.statedb)
+		if err != nil {
+			return
+		}
+		vmCfg := bc.cfg.VmConfig
+		vmCfg.Tracer = nil
+		synthetic := types.NewBlockWithHeader(header).WithBody(types.Body{Transactions: txs})
+		var noop atomic.Bool
+		bc.prefetcher.Prefetch(synthetic, warmdb, vmCfg, &noop)
```
cffls: This call execute transactions without taking fees from transactions, which introduces a DoS vector. Malicious users can submit resource-expensive transaction and replace them with a bit higher gas continuously without paying the gas fee. 

For example, if the current base gas price is 50Gwei, I can submit a transaction with 32M gas of SLOAD, with a initial gas price of 1Gwei, and continuously replace the existing tx by bumping up the gas by 10% for 41 times before reaching the 50Gwei base price (1Gwei * 1.1^41 = 49Gwei). This will result in the node to fully execute this transaction 41 times before the transaction is actually included in a block.

core/state_prefetcher.go
Raneet10: Yes but it also calculates state root which we strictly don't need. Pre-execution should fill up the accounts/storages. Do you think it still should be there ?

go.mod
Raneet10: Could do yeah but anyway this isn't getting merged to develop just yet.

core/blockchain_insert.go
Raneet10: Yeah right. I don't think these will make it to develop anyway.

core/blockchain.go
Raneet10: I commented it out so that we don't run multiple prefetches together. I'll probably remove this once we finalize all the optimizations.

core/blockchain.go
Raneet10: Not sure. I didn't make this change.

core/state_prefetcher.go
cffls: It should be there. Simply loading account/storage isn't sufficient, because in order to calculate the final state root hash, some sibling nodes will need to be retrieved, even though they aren't related to touched accounts.

core/blockchain.go
Raneet10: Good catch. Though I am planning to remove `PrefetchFromTxpool` altogether as I didn't see any impact of prefetching from txpool. Perhaps running it in worker (which is present currently) is more deterministic ?

core/state/reader.go
```diff
@@ -566,11 +564,56 @@ func (r *readerWithCacheStats) Storage(addr common.Address, slot common.Hash) (c
...
 		AccountHit:  r.accountHit.Load(),
 		AccountMiss: r.accountMiss.Load(),
 		StorageHit:  r.storageHit.Load(),
 		StorageMiss: r.storageMiss.Load(),
 	}
 }
+
+// SeedCaches inserts the supplied accounts and storage slots directly into the
+// local caches.
+func (r *ReaderWithCacheStats) SeedCaches(accounts map[common.Address]*types.StateAccount, storage map[common.Address]map[common.Hash]common.Hash) {
```
cffls, line 578: How do we know/ensure the data provided by seeded cache is always correct?

miner/worker.go
```diff
@@ -1451,22 +1495,74 @@ func (w *worker) fillTransactions(interrupt *atomic.Int32, env *environment) err
...
+					filteredTxs = append(filteredTxs, tx)
+				}
+			}
+			txs = filteredTxs
 
-		if err := w.commitTransactions(env, plainTxs, blobTxs, interrupt, new(uint256.Int)); err != nil {
-			return err
+			// start warm reader cache
+			tasks, _ := core.NewWarmExecTasks(w.chain, w.chainConfig, env.header, env.state, env.coinbase, txs, vm.Config{})
+			numProcs := max(1, 4*runtime.NumCPU()/5)
```
cffls: How is this calculation derived?

core/parallel_state_processor.go
```diff
@@ -95,6 +95,66 @@ type ExecutionTask struct {
...
+		usedGas  = new(uint64)
+	)
+	// Create block and EVM block context
+	blockNumber := header.Number
+	blockHash := header.Hash()
+	blockTime := header.Time
+	blockCtx := NewEVMBlockContext(header, bc, &coinbase)
+	shouldDelayFeeCal := false
+
+	for i, tx := range txs {
```
cffls, line 118: There could be large amount of transactions in `txs`, which may cause unnecessary warm up. For example, if there are 10k pending txs, but only 20% of them will be included in the next block, 80% of compute would be wasted. One way of limiting is to include transactions until the sum of their gas limit reaches the block gas limit.

core/state/reader.go
Raneet10: Well the aim is to warm the cache with _unmodified_ account/storage slots that are touched during the tx. I presume the data should be correct for most cases except maybe if there's a tx dependency. 
During testing on mainnet ndoe, however, I never saw a failure due to this.

miner/worker.go
Raneet10: I think this was the same value being used in state prefetcher from upstream.

core/parallel_state_processor.go
Raneet10: True but before the callsite we are [filtering the txs](https://github.com/0xPolygon/bor/pull/1871/changes#diff-689874b30f6f905ce6c47cdefeddcf052b467a7936a981f4b6cf38939e10d924R1529-R1543).

cffls: I see. On the other hand, with this filter, we might not be able to fill the block. For example, if a malicious user submit 2 transactions with high gas price and each transaction has gas limit of 30M, we will only mine these two transactions. However, if each transactions actually consumes only 1M gas, the block will be only filled with 2M gas in total.

core/state/reader.go
cffls: Could you help me understand the lifecycle of the cache? For example, when is value written, read, or deleted? Do they all happen for the same block? What if a new block read into an old value that should've been updated?

Raneet10: Sure, sorry for not clarifying earlier. So basically: 

- We start warming process for the individual block [here](https://github.com/0xPolygon/bor/pull/1871/changes#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215R3160). 
- This would create a [cached state reader and calls warmReaderCache](https://github.com/0xPolygon/bor/pull/1871/changes#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215R4376-R4378). 
- warmReaderCache calls [the Prefetch routine](https://github.com/0xPolygon/bor/pull/1871/changes#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215R865). 
- The Prefetch method does [the seeding](https://github.com/0xPolygon/bor/pull/1871/changes#diff-cf3fbfff4358b57b0d24c889c0c442592814cab289ca416c78412379b80a66e3R134). 
- The warmed reader is used for [actual processing of that block](https://github.com/0xPolygon/bor/pull/1871/changes#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215R3221-R3225). 
- The statedb is updated and we go again for the next block.

core/parallel_state_processor.go
Raneet10: Correct but I think the only way to accurately determine the gas used is to execute the txs and this is a time consuming process itself.

core/state/reader.go
cffls: Thanks for the explanation!

PR #1877, opened 2025-11-13
Include missing check of P256 on Amoy

lucca30:
# Description

Including check of p256 on precompiles when enabling a new HF on Amoy. The test already covers Mainnet, but misses amoy

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/core/vm/contracts_test.go
+++ b/core/vm/contracts_test.go
@@ -439,7 +439,6 @@ func TestPrecompiledP256Verify(t *testing.T) {
 // BOR: if this test failed, it means you should include PrecompiledP256Verify in the PrecompiledContracts
 // TODO: handle when common.BytesToAddress([]byte{0x01, 0x00}) will colide a new Ethereum's precompile
 func TestPrecompiledP256VerifyAlwaysAvailableInHFs(t *testing.T) {
-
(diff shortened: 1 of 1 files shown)


core/vm/contracts_test.go
```diff
@@ -439,19 +439,22 @@ func TestPrecompiledP256Verify(t *testing.T) {
 // BOR: if this test failed, it means you should include PrecompiledP256Verify in the PrecompiledContracts
 // TODO: handle when common.BytesToAddress([]byte{0x01, 0x00}) will colide a new Ethereum's precompile
 func TestPrecompiledP256VerifyAlwaysAvailableInHFs(t *testing.T) {
```
kamuikatsurgi, line 441: could we make a generic test for all the precompile contracts for mainnet and amoy?

kamuikatsurgi (approved):

core/vm/contracts_test.go
lucca30: We just want to make sure any new inserted hf has this precompile. The whole point is that we inserted it before Osaka was launched. So even without this check on future we are not going to see this issue again.

PR #1889, opened 2025-11-18
(feat): allow miner gas tip override

manav2401:
# Description

Adds a new flag `--allow-gas-tip-override` to allow block producers to use a different value instead of enforcing the default 25 gwei. The aim is to allow block producers to set a slightly lower value than default to handle some edge cases where transactions priced exactly at 25 gwei (or slightly more) are mineable and they don't bloat the pool. 

Note that rpc providers should fallback to default so any transaction with a gas tip lower than 25 gwei will be rejected. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/docs/cli/server.md
+++ b/docs/cli/server.md
@@ -288,8 +288,6 @@ The ```bor server``` command runs the Bor client.
 
 - ```allow-gas-tip-override```: Allows block producers to override the mining gas tip (default: false)
 
-- ```is-active-block-producer```: Indicates whether this validator is an active block producer post rio HF (to allow config modifications) (default: false)
-
 - ```miner.etherbase```: Public address for block mining rewards
 
 - ```miner.extradata```: Block extra data set by the miner (default = client version)


docs/cli/server.md
```diff
@@ -286,6 +286,8 @@ The ```bor server``` command runs the Bor client.
 
 - ```mine```: Enable mining (default: false)
 
+- ```is-active-block-producer```: Indicates whether this validator is an active block producer post rio HF (to allow config modifications) (default: false)
```
cffls: Would it be more clear to name this flag directly as something like `allow-gas-tip-override`? `is-active-block-producer` seems ambiguous. In fact, it isn't a value that could be controlled by the BP itself. For example, if producer vote changes and this BP is ousted, this flag will be conflicting with the fact, although it can still work as it is.

manav2401: My initial thought was to have a generic flag so that if we have to do a change like this in future, we could control it with the same feature flag instead of adding a new one. 

But, agree with your point about it being contradicting in some cases. Will update the name.

manav2401: Also now that we have a flag for gas tip override, do we want to have a constant defined which overrides it to 24 gwei or do we skip any enforcement and make sure the flag `miner.gasprice` is set to 24 gwei?

cffls: Making this value configurable seems like a good idea to me. For example, we can set the value to 24.9Gwei instead of 24Gwei if needed.

manav2401: Done, thanks!

kamuikatsurgi (approved):

PR #1903, opened 2025-11-24
core(txpool): align with geth

kamuikatsurgi:
# Description

txpool chores.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -689,7 +689,7 @@ func (w *worker) mainLoop() {
 
 				tcount := w.current.tcount
 
-				w.commitTransactions(w.current, plainTxs, blobTxs, nil, new(uint256.Int))
+				w.commitTransactions(w.current, plainTxs, blobTxs, nil)
 				stopFn()
 
 				// Only update the snapshot if any new transactons were added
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -1072,11 +1073,6 @@ mainloop:
 			continue
 		}
 
-		// If we don't receive enough tip for the next transaction, skip the account
-		if ptip.Cmp(minTip) < 0 {
```
kamuikatsurgi, line 1076: This has already been done as:
```go
filter := txpool.PendingFilter{
	MinTip: uint256.MustFromBig(tip.ToBig()),
}
```

Additionally, the `minTip` is always `new(uint256.Int)` (which is 0), so the check is redundant anyway.

miner/worker.go
```diff
@@ -959,11 +959,12 @@ func (w *worker) commitTransaction(env *environment, tx *types.Transaction) ([]*
 	}
 	env.txs = append(env.txs, tx)
 	env.receipts = append(env.receipts, receipt)
+	env.tcount++
```
kamuikatsurgi, line 962: geth has moved this line into this func.

marcello33 (approved):

PR #1938, opened 2025-12-16
Remove Recommit and Fix Parent Actual Time track on Prepare

lucca30:
# Description

This PR disables the recommit mechanism and improves how the parent header ActualTime is tracked. Recommit was removed because our current block production/plotting time already consumes most of the block interval, making reprocessing the same block wasteful and counter-productive, as those resources are better reserved for the next block. In addition, the parent ActualTime is now derived from the correct parent block (by hash) when available, instead of assuming the last locally processed block was always the parent. This fixes incorrect timestamp progression and avoids longer-than-expected waiting slots, especially when Prepare is invoked multiple times for the same block (e.g. during recommits).

# Changes

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -1252,10 +1252,6 @@ func (c *Bor) FinalizeAndAssemble(chain consensus.ChainHeaderReader, header *typ
 	// Assemble block
 	block := types.NewBlock(header, body, receipts, trie.NewStackTrie(nil))
 
-	if c.parentActualTimeCache != nil && !block.Header().ActualTime.IsZero() {
-		c.parentActualTimeCache.Add(block.Hash(), block.Header().ActualTime)
-	}
-
 	// return the final block for sealing
 	return block, receipts, nil
 }
@@ -1325,6 +1321,10 @@ func (c *Bor) Seal(chain consensus.ChainHeaderReader, block *types.Block, witnes
 	err = Sign(currentSigner.signFn, currentSigner.signer, header, c.config)
 	if err != nil {
 		return err
+	}
+
+	if c.parentActualTimeCache != nil && !header.ActualTime.IsZero() {
+		c.parentActualTimeCache.Add(header.Hash(), header.ActualTime)
 	}
 
 	// Wait until sealing is terminated or delay timeout.


consensus/bor/bor.go
```diff
@@ -1017,12 +1019,19 @@ func (c *Bor) Prepare(chain consensus.ChainHeaderReader, header *types.Header) e
 
 	if c.blockTime > 0 && c.config.IsRio(header.Number) {
 		// Only enable custom block time for Rio and later
-		parentActualTime := c.lastMinedBlockTime
-		if parentActualTime.IsZero() || parentActualTime.Before(time.Unix(int64(parent.Time), 0)) {
-			parentActualTime = time.Unix(int64(parent.Time), 0)
+
+		parentChainTime := time.Unix(int64(parent.Time), 0)
```
kamuikatsurgi: `parentChainTime` -> `parentBlockTime`?

consensus/bor/bor.go
```diff
@@ -1017,12 +1019,19 @@ func (c *Bor) Prepare(chain consensus.ChainHeaderReader, header *types.Header) e
...
+		// Default to parent chain timestamp
```
kamuikatsurgi: `// Default to parent chain timestamp` -> `// Default to parent block timestamp`?

consensus/bor/bor.go
```diff
@@ -326,6 +326,8 @@ func New(
 		},
 	})
 
+	c.parentActualTimeCache, _ = lru.New(100)
```
kamuikatsurgi: imo, `lru.New(10)` should suffice, wdyt?

miner/worker.go
```diff
@@ -562,16 +562,18 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+			// // If sealing is running resubmit a new work cycle periodically to pull in
+			// // higher priced transactions. Disable this overhead for pending blocks.
+			// if w.IsRunning() && (w.chainConfig.Clique == nil || w.chainConfig.Clique.Period > 0) {
+			// 	// Short circuit if no new transaction arrives.
+			// 	if w.newTxs.Load() == 0 {
+			// 		timer.Reset(recommit)
+			// 		continue
+			// 	}
+			// 	commit(true, commitInterruptResubmit)
+			// }
```
kamuikatsurgi: Can we remove the dead code if we're not going to use it?

kamuikatsurgi (approved):
smol nits

consensus/bor/bor.go
```diff
@@ -1017,12 +1019,19 @@ func (c *Bor) Prepare(chain consensus.ChainHeaderReader, header *types.Header) e
 
 	if c.blockTime > 0 && c.config.IsRio(header.Number) {
 		// Only enable custom block time for Rio and later
-		parentActualTime := c.lastMinedBlockTime
-		if parentActualTime.IsZero() || parentActualTime.Before(time.Unix(int64(parent.Time), 0)) {
-			parentActualTime = time.Unix(int64(parent.Time), 0)
+
+		parentChainTime := time.Unix(int64(parent.Time), 0)
+		// Default to parent chain timestamp
+		parentActualTime := parentChainTime
```
kamuikatsurgi: `parentActualTime` -> `parentActualBlockTime`

consensus/bor/bor.go
```diff
@@ -252,7 +252,7 @@ type Bor struct {
 	// The block time defined by the miner. Needs to be larger or equal to the consensus block time. If not set (default = 0), the miner will use the consensus block time.
 	blockTime time.Duration
 
-	lastMinedBlockTime time.Time
+	parentActualTimeCache *lru.Cache
```
kamuikatsurgi, line 256: Just a small comment we can add: `// Cache to store the actual times of the parent blocks`.

marcello33:
There seems to be some compilation errors for tests due to these changes. Other than that - and the suggestions from @kamuikatsurgi - LGTM, thanks

consensus/bor/bor.go
kamuikatsurgi: One q: if the calculated `header.Time` is in the past, we bump it here (could be because of certain delays and stuff), but we are not updating `header.ActualTime` here. Meaning it would be in the past and stale, both in the `header.ActualTime` field and in the `parentActualTimeCache` cache as well.

kamuikatsurgi: We should update `header.ActualTime` here as well.

kamuikatsurgi: wdyt?

kamuikatsurgi:
```go
now := time.Now()
if header.Time < uint64(now.Unix()) {
    header.Time = uint64(now.Unix())

    if !header.ActualTime.IsZero() {
        header.ActualTime = now
    }
}
```

lucca30: Yes, I agree. That makes sense. Header.ActualTime is always the represantion of time in a more granular way. So whenever we update time we must also update actual time. It could also be one of the reasons we faced some not expected behaviors. Nice catch

lucca30: The only change I would do is on the condition

```
	now := time.Now()
	if header.Time < uint64(now.Unix()) {
		header.Time = uint64(now.Unix())
		if c.blockTime > 0 && c.config.IsRio(header.Number) {
			header.ActualTime = now
		}
	}
```

wdyt?

consensus/bor/bor.go
lucca30: done

miner/worker.go
lucca30: done

consensus/bor/bor.go
lucca30: done

consensus/bor/bor.go
lucca30: done

consensus/bor/bor.go
lucca30: done

consensus/bor/bor.go
lucca30: done

PR #1943, opened 2025-12-17
fix(miner): track pending work using block number to avoid duplicate block building work

kamuikatsurgi:
# Description

Fixes the race condition with veblopTimer, which causes duplicate sealing of blocks.

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -853,7 +853,7 @@ func (c *Bor) verifySeal(chain consensus.ChainHeaderReader, header *types.Header
 		return err
 	}
 
-	if !snap.ValidatorSet.HasAddress(signer) {
+	if !snap.ValidatorSet.HasAddress(signer) && !isPartOfVeBlopSet(signer, header.Number.Uint64()) {
 		// Check the UnauthorizedSignerError.Error() msg to see why we pass number-1
 		return &UnauthorizedSignerError{number, signer.Bytes(), snap.ValidatorSet.Validators}
 	}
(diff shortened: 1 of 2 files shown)


miner/worker.go
```diff
@@ -553,14 +559,23 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+				continue
+			}
+
 			if !hasPendingTasks && time.Now().Unix()-int64(currentBlock.Time) >= int64(veblopTimeout.Seconds()) {
 				timestamp = time.Now().Unix()
+				w.pendingWorkBlock.Store(currentBlock.Number.Uint64() + 1)
 				commit(false, commitInterruptNewHead)
+				// veblopTimer is already reset by commit() so we don't need to reset it here.
+			} else {
+				veblopTimeout = time.Duration(w.chainConfig.Bor.CalculatePeriod(currentBlock.Number.Uint64())) * time.Second
```
cffls: Might be a good idea to verify the timeout is not greater than the actual block time set by the miner after this line:

```
			if veblopTimeout < w.blockTime {
				veblopTimeout = w.blockTime
			}
```

cffls:
Please make sure all E2E tests are passing

miner/worker.go
```diff
@@ -534,12 +538,14 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
 
 			timestamp = time.Now().Unix()
+			w.pendingWorkBlock.Store(w.chain.CurrentBlock().Number.Uint64() + 1)
 			commit(false, commitInterruptNewHead)
 
 		case head := <-w.chainHeadCh:
 			w.clearPending(head.Header.Number.Uint64())
 
 			timestamp = time.Now().Unix()
+			w.pendingWorkBlock.Store(head.Header.Number.Uint64() + 1)
```
cffls, line 560: We might need to check `pendingWorkBlock` here as well? It is possible that veblop timer creates a work task before entering here.

kamuikatsurgi: Added, thanks!

miner/worker.go
kamuikatsurgi: Added, thanks!

kamuikatsurgi:
> Please make sure all E2E tests are passing

All the tests are passing, except for the fast-forward test, which fails at the end of the stateless sync E2E test. It's also failing in other PRs. We need to investigate what happened there, but this PR doesn't change anything related to that issue.

PR #1944, opened 2025-12-17
fix: ignore invalid signer error for specific block range

avalkov:
# Description

Unstuck bor


adamdossa (approved):

web3nodes:
tbh, i wouldn't put the block range check until after the underlying issue is fixed. or else it might get stuck again

kamuikatsurgi:
Bump the version.

ssandeep:
@web3nodes , there could be some bad consequences of not having the block range check like introducing non determinism in whose block should get into the canonical chain among the block producers (without honouring the span). Want to be safe for now, we can remove it if not required after we verify it passes all corner cases.

PR #1947, opened 2025-12-18
Revert "V2.5.6 candidate"

pratikspatil024:
Reverts 0xPolygon/bor#1941


kamuikatsurgi:
We don't want to revert this candidate branch since we already have pushed out all the changes in [v2.5.6-beta3](https://github.com/0xPolygon/bor/releases/tag/v2.5.6-beta3), only the version is different.

manav2401 (approved):

PR #1958, opened 2025-12-26
Validator Set Overrides on Config

lucca30:
# Description

This PR finalizes and hardens the validator override mechanism introduced in response to a recent network outage.

The override is now implemented as an explicit, configuration-driven mechanism, strictly scoped by block range and validator set. This ensures the behavior is deterministic, auditable, and tightly bounded to the affected chain and timeframe, without impacting normal validator verification outside the override window.

# Changes

### DIFF
--- a/consensus/bor/snapshot_test.go
+++ b/consensus/bor/snapshot_test.go
@@ -1,6 +1,7 @@
 package bor
 
 import (
+	"crypto/ecdsa"
 	"math/big"
 	"sort"
 	"testing"
@@ -109,11 +110,7 @@ func TestGetSignerSuccessionNumber_ProposerNotFound(t *testing.T) {
 	signerTest := snap.ValidatorSet.Validators[3].Address
 
 	_, err := snap.GetSignerSuccessionNumber(signerTest)
-	require.NotNil(t, err)
-
-	e, ok := err.(*UnauthorizedProposerError)
-	require.True(t, ok)
-	require.Equal(t, dummyProposerAddress.Bytes(), e.Proposer)
+	assertUnauthorizedProposerError(t, err, dummyProposerAddress)
 }
 
 func TestGetSignerSuccessionNumber_SignerNotFound(t *testing.T) {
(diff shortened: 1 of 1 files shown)


kamuikatsurgi (approved):

params/version.go
```diff
@@ -23,10 +23,10 @@ import (
...
 
 const (
-	VersionMajor = 2  // Major version component of the current release
-	VersionMinor = 5  // Minor version component of the current release
-	VersionPatch = 5  // Patch version component of the current release
-	VersionMeta  = "" // Version metadata to append to the version string
+	VersionMajor = 2       // Major version component of the current release
+	VersionMinor = 5       // Minor version component of the current release
+	VersionPatch = 6       // Patch version component of the current release
+	VersionMeta  = "beta4" // Version metadata to append to the version string
```
ssandeep: let's remove this version change as we won't be releasing now and it's stale.

internal/cli/server/chains/mainnet.go
```diff
@@ -74,6 +74,11 @@ var mainnetBor = &Chain{
 				OverrideStateSyncRecordsInRange: []params.BlockRangeOverride{
 					{StartBlock: 73812433, EndBlock: 73826700, Value: 0},
 				},
+				OverrideValidatorSetInRange: []params.BlockRangeOverrideValidatorSet{
+					{StartBlock: 80440819, EndBlock: 80440834, Validators: []common.Address{
```
cffls, line 78: This change should be added to `params/config.go` and `genesis-mainnet-v1.json` as well.

manav2401 (approved):
Changes wise, LGTM. 

Small nits for testing. I saw that there are more tests in erigon for the same change. Is it possible to port all of them in bor as well? 

Also, is it possible to have an e2e test which confirms that blocks built by a signer which is added in override set but is not in validator set is accepted?

internal/cli/server/chains/mainnet.go
lucca30: done, thanks for reporting

lucca30:
> Changes wise, LGTM.
> 
> Small nits for testing. I saw that there are more tests in erigon for the same change. Is it possible to port all of them in bor as well?
> 
> Also, is it possible to have an e2e test which confirms that blocks built by a signer which is added in override set but is not in validator set is accepted?

Thanks for feedback. Included these extra tests and covered a few more scenarios, checking the succession for example

PR #1959, opened 2025-12-29
fix(api): traceBlock

kamuikatsurgi:
# Description

Fixes traceBlock api.


manav2401 (approved):

manav2401:
I wonder why did the tests not catch this. Can we check if we have a tracing related test pre and post this HF?

kamuikatsurgi:
> I wonder why did the tests not catch this. Can we check if we have a tracing related test pre and post this HF?

It doesn't have a specific test related to this HF.

PR #1960, opened 2025-12-29
Test validator's ability to rewind upon span rotation

cffls:
# Description

Add an integration test that verifies the block producer is able to rewind to the correct block when a span rotation happens after some blocks were built on a wrong fork.
This test should pass after https://github.com/0xPolygon/bor/pull/1957 is merged.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments


lucca30:
lgtm. The tests seems not to be parallel, which means an increase at the test time. How many time do they take? If its big, would be good considering parallelization if possible

consensus/bor/bor.go
```diff
@@ -1724,6 +1724,17 @@ func (c *Bor) SetHeimdallClient(h IHeimdallClient) {
...
 
+// PurgeCache clears all cached snapshots and span data. This is useful in tests
+// when the mock heimdall client is changed and old cached data needs to be invalidated.
+func (c *Bor) PurgeCache() {
+	// Clear the recents cache (snapshots)
+	c.recents.DeleteAll()
+	// Clear the recent verified headers cache
+	c.recentVerifiedHeaders.DeleteAll()
+	// Clear the span store cache
+	c.spanStore.PurgeCache()
```
kamuikatsurgi, line 1750: Just a nit: we can rename `c.spanStore.PurgeCache()` to `c.spanStore.DeleteAll()` to match the naming of the others.

cffls: PurgeCache seems closer to what it is doing I think. DeleteAll sounds like we are deleting the span and a bit more destructive.

PR #1968, opened 2026-01-06
Dandeli HF: modify block gas target calculations

manav2401:
# Description

Instead of deriving block gas target to calculate base fee for eip 1559 using elasticity multiplier, this HF introduces a percentage denoting the percent of gas target to choose from total block gas limit. It acts as an inverse of elasticity multiplier in percent terms. E.g. current default elasticity multiplier is 2 which translates to 50% of total block limit as gas target. To address the gas spikes, this is modified to 66% which means gas target would be 66% of block gas limit. This makes the base fee increase bit slower that earlier during surge in chain activity. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/builder/files/genesis-amoy.json
+++ b/builder/files/genesis-amoy.json
@@ -24,6 +24,7 @@
       "bhilaiBlock": 22765056,
       "rioBlock": 26272256,
       "madhugiriBlock": 28899616,
+      "madhugiriProBlock": 29287400,
       "dandeliBlock": 31890000,
       "skipValidatorByteCheck": [26160367, 26161087, 26171567, 26173743, 26175647],
       "stateSyncConfirmationDelay": {


params/protocol_params.go
```diff
@@ -140,6 +140,9 @@ const (
 	DefaultBaseFeeChangeDenominator = 8 // Bounds the amount the base fee can change between blocks.
 	DefaultElasticityMultiplier     = 2 // Bounds the maximum gas limit an EIP-1559 block may have.
 
+	DefaultTargetGasPercentage     = 50 // Specifies target block gas as percentage of block gas limit for EIP-1559
+	TargetGasPercentagePostDandeli = 60 // Specifies target block gas as percentage of block gas limit for EIP-1559 after Dandeli hard fork
```
ssandeep: Let's increase this to 65%

params/protocol_params.go
```diff
@@ -140,6 +140,9 @@ const (
 	DefaultBaseFeeChangeDenominator = 8 // Bounds the amount the base fee can change between blocks.
 	DefaultElasticityMultiplier     = 2 // Bounds the maximum gas limit an EIP-1559 block may have.
 
+	DefaultTargetGasPercentage     = 50 // Specifies target block gas as percentage of block gas limit for EIP-1559
```
pratikspatil024, line 143: just a minor comment saying `only used in testing` would be good

consensus/misc/eip1559/eip1559.go
```diff
@@ -98,3 +98,13 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header) *big.Int {
 		return baseFee
 	}
 }
+
```
pratikspatil024, line 102: comment for future upstream references
```
// BOR specific
```

params/protocol_params.go
manav2401: It's inverse of elasticity multiplier so I've kept it as reference without mentioning anything about testing.

cffls:
There is a [linter error](https://github.com/0xPolygon/bor/actions/runs/20754511137/job/59592705720?pr=1968), otherwise looks good to me

PR #1989, opened 2026-01-14
Dynamic EIP-1559 params setup

lucca30:
# Description

This PR proposes a more flexible validation over EIP-1559 configurable TargetGas and BaseFeeDenominator, in the same way is already done by the gas limit. Which means that consensus will not block changes over this parameters set by the block producer. The goal is to allow more dynamic setups by the block producer whenever necessary.

The only proposed restriction is on the boundary range of 5% of parentBaseFee for increase or decrease. Which is already above the current limit of 1.7% on a scenario of 0% gas usage on parent block.

# Changes


adamdossa:
@claude Please can you do a thorough review of this change, highlighting any potential safety issues, bugs or edge-cases.

consensus/misc/eip1559/eip1559.go
```diff
@@ -43,7 +43,17 @@ func VerifyEIP1559Header(config *params.ChainConfig, parent, header *types.Heade
 	if header.BaseFee == nil {
 		return errors.New("header is missing baseFee")
 	}
-	// Verify the baseFee is correct based on the parent header.
+
+	// After Dandeli hard fork, base fee validation is removed to allow dynamic configuration.
+	// This enables validators to set base fees according to new consensus rules without
+	// strict protocol enforcement, supporting more flexible fee markets.
+	// Pre-Dandeli blocks still require strict base fee validation for consensus safety.
+	if config.Bor != nil && config.Bor.IsDandeli(header.Number) {
```
cffls, line 58: Do we want to add a sanity check here, something like base fee shouldn't be 10% more different from the last block?

lucca30: That's a great topic to be discussed. The idea of the baseFeeDenominator is exactly set how fast the change happens. So if we put a check we are basically limiting the ability of configuration on this. So it's a good topic to be raised and discussed

cffls: The existing logic checks how base fee should change strictly, which means it doesn't allow a range of values. I am thinking about putting a range of values, e.g. [0%-10%] for it with this change.

lucca30: Agreed on that. Since our current maximum change is 1.7%, i set a maximum of 5% change, which seems to be enough to cap any out of scope change

internal/cli/server/config.go
```diff
@@ -417,6 +417,12 @@ type SealerConfig struct {
 	// BlockTime is the block time defined by the miner. Needs to be larger or equal to the consensus block time. If not set (default = 0), the miner will use the consensus block time.
 	BlockTime    time.Duration `hcl:"-,optional" toml:"-"`
 	BlockTimeRaw string        `hcl:"blocktime,optional" toml:"blocktime,optional"`
+
+	// TargetGasPercentage is the target gas as percentage of gas limit (1-100, default 65) for post-Dandeli blocks
+	TargetGasPercentage uint64 `hcl:"target-gas-percentage,optional" toml:"target-gas-percentage,optional"`
+
+	// BaseFeeChangeDenominator is the base fee change rate (must be >0, default 64) for post-Dandeli blocks
+	BaseFeeChangeDenominator uint64 `hcl:"base-fee-change-denominator,optional" toml:"base-fee-change-denominator,optional"`
```
cffls, line 425: One question - is it possible for a BP to set a value for these two params, resulting base fee change to be greater than 5% limit? If so, we probably want to have a sanity check on node start, which will error out/stop the node if the resulting base fee is invalid.

eth/peer.go
```diff
@@ -726,13 +726,13 @@ func (p *ethPeer) doWitnessRequest(
 		}
 	}()
 	witReqsWg.Add(1)
+	mapsMu.Lock()
```
cffls, line 729: Thank you for fixing this! ☺️

internal/cli/server/config.go
lucca30: Lower values will allow that. But we are already capping the `CalcBaseFee` function so it returns within this margin, making it acceptable. 

I think if we try to block params before just because a specific gas usage could trigger a change bigger than 5% than we are including an extra complexity for setup which does not seems necessary. 


wdyt on just capping on the calc as we are already doing?

eth/peer.go
lucca30: this one started repeatedly getting rejected. so easy to caught. Now with this ai feedbacks it would probably be caught.

PR #1991, opened 2026-01-16
BP Metrics

lucca30:
# Description

This PR aims to provide a few metrics that were not observed on the block production. The first of them are the cache metrics, which when importing blocks but not on the designated block producer.

As you can see on this before/after images, now block producers shows their cache metrics. You will observer the hits are slightly less but this occurs because when importing we use both parallel and sequential processors.

Before:
before_bp_metrics

After:
after_bp_metrics

The other deliverable were the `FinalizeAndAssemble` and Commit metrics, to let us clearly track the time we take on StateRoot calculations and the time it take to produce the block from start to the end

# Changes

### DIFF
--- a/consensus/beacon/consensus.go
+++ b/consensus/beacon/consensus.go
@@ -404,43 +404,6 @@ func (beacon *Beacon) FinalizeAndAssemble(chain consensus.ChainHeaderReader, hea
 	// Assemble the final block.
 	block := types.NewBlock(header, body, receipts, trie.NewStackTrie(nil))
 
-	// Create the block witness and attach to block.
-	// This step needs to happen as late as possible to catch all access events.
-	if chain.Config().IsVerkle(header.Number) {
-		keys := state.AccessEvents().Keys()
-
-		// Open the pre-tree to prove the pre-state against
-		parent := chain.GetHeaderByNumber(header.Number.Uint64() - 1)
-		if parent == nil {
-			return nil, nil, 0, fmt.Errorf("nil parent header for block %d", header.Number)
-		}
-		preTrie, err := state.Database().OpenTrie(parent.Root)
-		if err != nil {
-			return nil, nil, 0, fmt.Errorf("error opening pre-state tree root: %w", err)
-		}
-		postTrie := state.GetTrie()
-		if postTrie == nil {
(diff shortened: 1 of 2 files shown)


miner/worker.go
```diff
@@ -1757,9 +1804,14 @@ func (w *worker) commit(env *environment, interval func(), update bool, start ti
...
 		var block *types.Block
 		var err error
+
+		// Track time for FinalizeAndAssemble (state root calculation + block assembly)
+		finalizeStart := time.Now()
 		block, env.receipts, err = w.engine.FinalizeAndAssemble(w.chain, env.header, env.state, &types.Body{
 			Transactions: env.txs,
 		}, env.receipts)
+		finalizeDuration := time.Since(finalizeStart)
+		finalizeAndAssembleTimer.Update(finalizeDuration)
```
manav2401, line 1817: IMO, `FinalizeAndAssemble` also involves operations to fetch heimdall data (e.g. span or state-sync events) and processing them (for sprint end blocks) so can skew the average for state root calculations. We should instead have another metric which wraps the call to `IntermediateRoot`, wdyt?

lucca30: Good point. We can definitely make the metric more granular around `IntermediateRoot`. That said, the Heimdall fetch itself is usually negligible and, if it ever becomes an issue, we can address it directly by prefetching / running it in parallel instead of doing it during finalization. So I’m fine with either approach, but the skew isn’t fundamentally unavoidable.

cffls: I think we can calculate the time it takes to fetch data from heimdall, and deduct that value from total time. We want to avoid false alarm as much as possible if heimdall is lagging.

cffls:
Looks good! Would be great to bump test coverage to 90%.

miner/worker.go
lucca30: Included a `IntermediateRootTimer` to keep track specifically on the StateRoot calculation so. Not thinking on doing one just for heimdall fecthing since its easier to just improve this by pipe this call before the block production.

wdyt?
@manav2401 @cffls

cffls: Adding IntermediateRootTimer makes sense. Thanks!

PR #2000, opened 2026-01-21
init: CLAUDE.md file

kamuikatsurgi:
# Description

Adds `CLAUDE.md` file.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,6 +97,13 @@ Bor focuses on high throughput, low gas fees, and full EVM compatibility.
 4. **Context**: Always propagate for cancellation
 5. **Database**: Use `ethdb` interfaces, batch writes when possible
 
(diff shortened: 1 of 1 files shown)


lucca30:
Nice initiative @kamuikatsurgi ! One suggestion from some work I did on this project and was a common repetitive prompt is:

- Take care on covering the introduced lines with tests
- Avoid introduction of duplicates

So would be good introducing them here. Also, I see a few comments just about sprint and validator stuff, is it worth being there? If so, shouldn't we introduce more data here?

kamuikatsurgi:
@lucca30

> So would be good introducing them here. Also, I see a few comments just about sprint and validator stuff, is it worth being there? If so, shouldn't we introduce more data here?

That is just to give some context on when and when not to comment section. Nothing with the specifics there.

PR #2006, opened 2026-01-22
Add Claude Assistant Workflows

adamdossa:
# Description

Add Claude Assistant Workflows

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -36,6 +36,7 @@ jobs:
           plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
           plugins: 'code-review@claude-code-plugins'
           prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
+          claude_args: '--model claude-opus-4-5-20251101'
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
 
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -51,6 +51,8 @@ jobs:
           additional_permissions: |
             actions: read
 
+          claude_args: '--model claude-opus-4-5-20251101'
+
           # Optional: Give a custom prompt to Claude. If this is not specified, Claude will perform the instructions specified in the comment that tagged it.
           # prompt: 'Update the pull request description to include a summary of changes.'
 


.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Run Claude Code Review
+        id: claude-review
+        uses: anthropics/claude-code-action@2316a9a8dbe4e7ac26008088f966f8c290e9a2e1
```
manav2401, line 33: We can also specify the mode to use. I am not sure what's the default one but we'll see once we try this on a PR. If it's sonnet, we should explicitly use opus.

.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          anthropic_api_key: ${{ secrets.CLAUDE_API_KEY }}
+          plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
+          plugins: 'code-review@claude-code-plugins'
+          prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
```
manav2401, line 38: We can be more verbose on the prompt but let's see how it performs initially. Can change later.

.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,44 @@
...
+    # Optional: Only run on specific file changes
+    # paths:
+    #   - "src/**/*.ts"
+    #   - "src/**/*.tsx"
+    #   - "src/**/*.js"
+    #   - "src/**/*.jsx"
+
+jobs:
+  claude-review:
+    # Optional: Filter by PR author
```
kamuikatsurgi: We could add some logic here so that it only runs on PR opened by members, or only they can trigger it.

.github/workflows/claude-code-review.yml
adamdossa: Added opus explicitly as a model parameter

.github/workflows/claude-code-review.yml
adamdossa: Added logic for this.

.github/workflows/claude-code-review.yml
adamdossa: The code-review plugin looks a good initial approach, although it does rely on having CLAUDE.md in a decent state.

We can iterate this once we start seeing results and use an explicit prompt if that works better for us.

https://github.com/anthropics/claude-code/blob/main/plugins/code-review/README.md

PR #2012, opened 2026-01-24
fix(ci): use custom kurtosis branch

kamuikatsurgi:
# Description

Use the custom kurtosis-pos branch to fix the CI.

### DIFF
--- a/.github/workflows/kurtosis-stateless-e2e.yml
+++ b/.github/workflows/kurtosis-stateless-e2e.yml
@@ -20,7 +20,7 @@ env:
 
 jobs:
   e2e-tests:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


cffls:
If time permits, maybe it is worth waiting for tx-eip1559-gas-price-configs to merge into main. Otherwise, the kurtosis branch will get outdated very quickly.

kamuikatsurgi:
> If time permits, maybe it is worth waiting for tx-eip1559-gas-price-configs to merge into main. Otherwise, the kurtosis branch will get outdated very quickly.

It can't be merged into main because the breaking change in hv2 is not yet released and it will break kurtosis CI.

Till a new hv2 version, and then a new kurtosis-pos version gets released, I will keep merging main into this branch to make sure it doesn't get outdated.

PR #2029, opened 2026-01-30
added rate limiting to address biased trie cache preloading

pratikspatil024:
# Description

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/triedb/pathdb/biased_fastcache.go
+++ b/triedb/pathdb/biased_fastcache.go
@@ -177,13 +177,21 @@ func (c *AddressBiasedCache) preloadAddressAsync(db ethdb.Database, addr common.
 		// Apply rate limiting after reading, based on actual bytes read
 		if limiter != nil {
 			if err := limiter.WaitN(c.ctx, len(nodeData)); err != nil {
-				log.Info("Preload interrupted during rate limit wait",
+				if c.ctx.Err() != nil {
+					log.Info("Preload interrupted during shutdown",
+						"account hash", accountHash.Hex(),
+						"entries", entriesLoaded,
+						"max depth", maxDepthReached,
+						"size", common.StorageSize(totalBytesLoaded).String(),
+						"elapsed", time.Since(startTime))
+					return
+				}
+				// Node exceeds burst size — skip it and continue preloading
+				log.Warn("Preload skipping oversized node",
 					"account hash", accountHash.Hex(),
-					"entries", entriesLoaded,
-					"max depth", maxDepthReached,
-					"size", common.StorageSize(totalBytesLoaded).String(),
-					"elapsed", time.Since(startTime))
-				return
+					"node size", len(nodeData),
+					"burst", limiter.Burst())
+				continue
 			}
 		}
 
--- a/triedb/pathdb/biased_fastcache_test.go
+++ b/triedb/pathdb/biased_fastcache_test.go
@@ -690,7 +690,7 @@ func TestAddressBiasedCache_RateLimitInterruption(t *testing.T) {
 
 // TestAddressBiasedCache_ShutdownDuringRateLimitWait specifically tests the scenario
 // where the preload goroutine is blocked in limiter.WaitN() when shutdown occurs.
-// This covers the "Preload interrupted during rate limit wait" log path.
+// This covers the "Preload interrupted during shutdown" log path.
 func TestAddressBiasedCache_ShutdownDuringRateLimitWait(t *testing.T) {
 	db := rawdb.NewMemoryDatabase()
 	addr := common.HexToAddress("0x1234567890123456789012345678901234567890")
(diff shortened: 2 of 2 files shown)


triedb/pathdb/biased_fastcache.go
```diff
@@ -156,6 +174,19 @@ func (c *AddressBiasedCache) preloadAddressAsync(db ethdb.Database, addr common.
 			continue
 		}
 
+		// Apply rate limiting after reading, based on actual bytes read
+		if limiter != nil {
+			if err := limiter.WaitN(c.ctx, len(nodeData)); err != nil {
+				log.Info("Preload interrupted during rate limit wait",
```
cffls: Would this line flood the log?

triedb/pathdb/biased_fastcache.go
```diff
@@ -156,6 +174,19 @@ func (c *AddressBiasedCache) preloadAddressAsync(db ethdb.Database, addr common.
...
+					"account hash", accountHash.Hex(),
+					"entries", entriesLoaded,
+					"max depth", maxDepthReached,
+					"size", common.StorageSize(totalBytesLoaded).String(),
+					"elapsed", time.Since(startTime))
+				return
```
cffls: Not sure why the function returns here. If we return here it means the cache won't be warmed.

triedb/pathdb/biased_fastcache.go
pratikspatil024: This log will be printed only once (per address) during shutdown.

triedb/pathdb/biased_fastcache.go
pratikspatil024: Correct, this is expected as this will be triggered during shutdown.

pratikspatil024:
> Looks good overall! What long does it take to preload mainnet with the default throttling setting?

@cffls - with `500KB/s` on our mainnet devnode, for a cache of 4096MB, it takes around 1h30m.

I think this `500KB/s` default could be a bit too optimistic for the BPs, so we can bump the default values a bit, I'll check the logs and confirm.

pratikspatil024:
@claude addressed your comments.
Review this PR again.

triedb/pathdb/biased_fastcache.go
```diff
@@ -156,6 +174,19 @@ func (c *AddressBiasedCache) preloadAddressAsync(db ethdb.Database, addr common.
 			continue
 		}
 
+		// Apply rate limiting after reading, based on actual bytes read
+		if limiter != nil {
+			if err := limiter.WaitN(c.ctx, len(nodeData)); err != nil {
```
lucca30, line 179: Regarding `len(nodeData)` to measure the rate:

- The cache is composed by `key` + `nodeData`

```
        key := append(accountHash.Bytes(), item.path...)
		addrCache.Set(key, nodeData)
```

- Key is basically `accountHash` + `path`

I guess is super low, but hash is 32 bytes size, so if nodeData has something like 100 bytes than would be good account it

triedb/pathdb/biased_fastcache.go
cffls: From the [api doc](https://pkg.go.dev/golang.org/x/time/rate#Limiter.WaitN), looks like it will also return an error if n exceeds the limit size?

triedb/pathdb/biased_fastcache.go
pratikspatil024: Thanks @lucca30. I agree that the cache is composed by `key + nodeData`, but we only read `nodeData` from the database. And because we want to put a rate limit on that, I think it is better to only account for `nodeData`.
Let me know if you think otherwise.

lucca30: Yeah, agree.

triedb/pathdb/biased_fastcache.go
pratikspatil024: Yes, you are right. It will return an error immediately if n exceeds the burst size. Here the burst is 64*1024 (64KB). If a single trie node is larger than 64KB, it will return an error immediately.
But I believe individual trie node are well under 64KB, so this is unlikely to happen.
I can differentiate the error/increase the burst size if you think this is accurate

cffls: Thanks! I think it would be good to differentiate the error in case the trie node is larger than 64KB or the burst size is lowered in the future. If that happens, we can print an error/warning message and continue with warm up I guess?

pratikspatil024: Addressed [here](https://github.com/0xPolygon/bor/pull/2029/commits/f8b198388f028adf1e67982b3f0fee16ac333444), thank you!

cffls (approved):

PR #2031, opened 2026-02-02
Prefetch Transactions from Pool & PIP-66 Back

lucca30:
# Description

- Implements state prefetching from the transaction pool during block production to improve cache hit rates and reduce block building latency. 
- Brings PIP-66 back with a fix to avoid early span checks due to the early announcement. 
- Change on where we wait from `Seal` to `Prepare` phase

### 1. State Prefetching from Transaction Pool
- **Location**: `miner/worker.go`, `core/state_prefetcher.go`
- Prefetches state (accounts, storage) for pending transactions before block production starts
- Runs prefetch in parallel with block building to warm up the state cache

### 2. Configuration Flags
- **`--miner.prefetch`**: Enable/disable transaction pool prefetching (default: disabled)
- **`--miner.prefetch.gaslimit.percent`**:  Percentage of block gas limit to be used as internal limit of prefetched transactions.

### 3. Enhanced Metrics
- **Location**: `miner/worker.go`, `core/state/reader.go`, `core/blockchain.go`
- Separated worker cache metrics from blockchain import metrics (prefix: `worker/chain/...`)
- Added prefetch attribution metrics to measure effectiveness:
  - `hit_from_prefetch`: Cache hits attributed to prefetch
  - `prefetch_used_unique`: Unique accounts loaded by prefetch and used during processing

### Prefetch Flow
1. When `commitWork` is called, two parallel operations start:
   - State prefetch for pending pool transactions
   - Current block production (which waits on `prepare` phase)
2. Block building uses the warmed cache from prefetch
3. Metrics track which cache hits came from prefetch vs. prior execution

## Results

From a local test with `polycli` running both `random` and `Uniswap` mode we observed that many storages that never were prefetched are now being prefetched and also a slightly increase on accounts being early prefetched

cache_storage

cache_account

### Worker Commit Transactions Time
Instance 1 has prefetch enabled, results in ms

image

PS: The second instance is the one with prefetchFromPool activated

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/miner/worker_test.go
+++ b/miner/worker_test.go
@@ -2140,7 +2140,8 @@ func TestPrefetchRaceWithSetExtra(t *testing.T) {
 // This test verifies that Go's GC correctly handles StateDB lifecycle even when prefetch
 // goroutines continue running after commitWork() returns, proving that no goroutine leaks occur.
 func TestPrefetchGoroutineLifecycle(t *testing.T) {
-	t.Parallel()
+	// Note: t.Parallel() removed - this test measures global goroutine count
+	// and must run serially to avoid interference from other parallel tests
 
 	var (
 		engine      consensus.Engine
@@ -2396,7 +2397,8 @@ func TestStateDBLifecycle_WithoutWait(t *testing.T) {
 // This simulates rapid block production where new prefetch goroutines spawn before previous ones complete,
 // validating that overlapping goroutines safely manage their own StateDB lifecycle with no panics, races, or leaks.
 func TestRapidBlockProduction_WithoutWait(t *testing.T) {
-	t.Parallel()
+	// Note: t.Parallel() removed - this test measures global goroutine count
+	// and must run serially to avoid interference from other parallel tests
 
 	w, b, engine, ctrl := setupBorWorkerWithPrefetch(t, 100, 200*time.Millisecond)
 	defer engine.Close()
@@ -2462,7 +2464,8 @@ func TestRapidBlockProduction_WithoutWait(t *testing.T) {
 // This integration test covers metrics tracking, cache effectiveness, and ensures no goroutine leaks
 // occur during normal block production with prefetch enabled.
 func TestPrefetchE2E(t *testing.T) {
-	t.Parallel()
+	// Note: t.Parallel() removed - this test measures global goroutine count
+	// and must run serially to avoid interference from other parallel tests
 
 	// Setup worker with prefetch enabled
 	w, b, engine, ctrl := setupBorWorkerWithPrefetch(t, 100, 1*time.Second)
@@ -2599,7 +2602,8 @@ func TestReorgDuringPrefetch(t *testing.T) {
 // This test produces 10 consecutive blocks with prefetch enabled, monitoring for
 // goroutine leaks, memory accumulation, and consistent prefetch behavior.
 func TestPrefetchMultiBlock(t *testing.T) {
-	t.Parallel()
+	// Note: t.Parallel() removed - this test measures global goroutine count
+	// and must run serially to avoid interference from other parallel tests
 
 	// Setup worker with prefetch enabled
 	w, b, engine, ctrl := setupBorWorkerWithPrefetch(t, 100, 1*time.Second)


lucca30:
> ## Code Review
> **Bug found in consensus/bor/bor.go at line 1080**
> 
> The delay calculation delay = time.Until(parentTime) computes time until the PARENT block timestamp. Since the parent block has already been mined, its timestamp is in the past, making this a negative duration. When time.After(delay) is called with a negative duration at line 1102 (
> 
> https://github.com/0xPolygon/bor/blob/0077186ac312253bff6edce42c6013d07afc8bdd/consensus/bor/bor.go#L1101-L1103
> 
> ), it returns immediately without any actual waiting.
> The old Seal logic correctly waited until header.GetActualTime() (the new block target time, which is in the future):
> 
> https://github.com/0xPolygon/bor/blob/0077186ac312253bff6edce42c6013d07afc8bdd/consensus/bor/bor.go#L1367-L1370
> 
> **Suggested fix** at line 1080 (
> 
> https://github.com/0xPolygon/bor/blob/0077186ac312253bff6edce42c6013d07afc8bdd/consensus/bor/bor.go#L1078-L1082
> 
> ):
> ```
> delay = time.Until(time.Unix(int64(header.Time), 0))
> ```
> 
> **Note:** The test at line 1741 passes because it sets genesis time 3 seconds in the **future**, which makes parentTime positive. In production, the parent block is always in the past.
> 
> Checked for bugs and CLAUDE.md compliance.


The delay calculation `delay = time.Until(parentTime)` is **intentional and correct** for the early announcement feature introduced in Rio.

## Design Intent

The Rio fork allows **early block announcement** to improve block propagation speed. The primary producer (succession == 0) can produce and announce blocks as soon as `parent.Time` has passed, rather than waiting until `header.Time`.

**Key timing relationships:**
- `header.Time = parent.Time + producerDelay` (where `producerDelay ≥ 2s`)
- Old behavior: Wait until `header.Time` before sealing
- New behavior: Wait until `parent.Time` before preparing (allowing earlier production)

## Why negative delays are correct

In production, when the parent block is in the past:
1. `delay = time.Until(parentTime)` is negative
2. `time.After(negative_duration)` returns immediately
3. **This is the intended fast path** - produce the block immediately without artificial delay

cmd/utils/flags.go
```diff
@@ -598,6 +598,18 @@ var (
 		Usage:    "0x prefixed public address for the pending block producer (not used for actual block production)",
 		Category: flags.MinerCategory,
 	}
+	MinerPrefetchFlag = &cli.BoolFlag{
```
cffls: `cmd/utils` aren't used by bor (only for geth). Can probably remove this.

consensus/bor/bor.go
```diff
@@ -402,13 +402,30 @@ func (c *Bor) verifyHeader(chain consensus.ChainHeaderReader, header *types.Head
...
+		// Using strict CalcProducerDelay would reject valid blocks, so we just ensure announcement
+		// time comes after parent time to allow for flexible blocktime.
+		var parent *types.Header
+
+		if len(parents) > 0 {
+			parent = parents[len(parents)-1]
+		} else {
+			parent = chain.GetHeader(header.ParentHash, number-1)
+		}
+		if parent == nil || now < parent.Time {
```
cffls, line 416: Why are we comparing current time with parent's block time, which is presumed to be imported already? Should it have the same check logic as bhilai below?

consensus/bor/bor.go
```diff
@@ -1043,14 +1073,18 @@ func (c *Bor) Prepare(chain consensus.ChainHeaderReader, header *types.Header) e
 		actualNewBlockTime := parentActualBlockTime.Add(c.blockTime)
 		header.Time = uint64(actualNewBlockTime.Unix())
 		header.ActualTime = actualNewBlockTime
+		delay = time.Until(parentActualBlockTime)
 	} else {
-		header.Time = parent.Time + CalcProducerDelay(number, succession, c.config)
+		parentTime := time.Unix(int64(parent.Time), 0)
+		producerDelay := CalcProducerDelay(number, succession, c.config)
+		header.Time = parent.Time + producerDelay
+		delay = time.Until(parentTime)
```
cffls: This would be simpler?

```
header.Time = parent.Time + CalcProducerDelay(number, succession, c.config)
delay = time.Until(time.Unix(int64(parent.Time), 0))
```

core/state/reader.go
```diff
@@ -418,13 +431,40 @@ func newReader(codeReader ContractCodeReader, stateReader StateReader) *reader {
...
+	// that was prefetched. Used to compute unique-usage/precision.
+	usedByProcess uint32
+}
+
+// storageCacheEntry is the cached storage slot plus attribution metadata.
+// Note: stored inline (no per-slot heap alloc).
+type storageCacheEntry struct {
+	value  common.Hash
+	origin readerRole
+}
```
cffls, line 459: These cache seems a big change. Could you please briefly explain how these cache works?

consensus/bor/bor.go
lucca30: Yeah it pretty much has the same goal on the early announcement check we did on Bhilai, but a bit more flexible. Here's the challenge:

- After Rio HF we introduced BlockTime, which allows intervals between blocks bigger than block period (which is the value agreed through consensus)
- An early announcement can happens as soon as it finished block production (which starts right after the parent time)

So this check, like the Bhilai one, just works as a sanity check to avoid any try on the network to send a future block.

consensus/bor/bor.go
lucca30: Yeah, it is. It was just done that way because some logs used during development. Missed simplifying it again, thanks.

core/state/reader.go
lucca30: Sure!

The change here aims to meet one specific goal: let us know which caches were provided because of the prefetch cache and which were not.

So in order to do that I'm replacing the cache map values:
- `types.StateAccount` by `accountCacheEntry`
- `common.Hash` by `storageCacheEntry`

But is important to see that all behaviors stays exactly the same. The introduction of these structures just lets us track who cached it first givin us a bit more precision on the gains with this feature. I did it in the best way to not impact performance and not impact memory, since I'm just introducing 1 byte for storageCacheEntry and 5 bytes for accountCacheEntry.

It's important pointing that there is an extra metric for `accounts` which tracks how many unique caches were hit by prefetch, which gives us a precise measurement of how many misses we effectively avoided, since if a value is called twice they will be cached at second time.

lucca30:
> This PR is quite large to review, and it essentially contains two features. If possible, could you split it into two, one is about PIP-66 and the other is txns prefetch?

@cffls 

Yeah, I do also agree it got big. But is important to point that the Transaction Prefetching benefits and partially depends on PIP-66. So it would be a bit hard trying to split those changes now. What I'll try to do is provided a summary of snippets related to the PIP-66 below.

### 1. Early Announcement Check
This section introduces a new and more flexible way to check early announcements too early because of the introduced `BlockTime` which is not on consensus. (better explained on another PR comment)
https://github.com/0xPolygon/bor/blob/6acb9e10d628c88443e84a9df3d2f6fd0b568f85/consensus/bor/bor.go#L405-L438

### 2. Accounting the Announcement Time into timeout for span check
As we discussed on a chat, this address the original issue we got when introduction PIP-66. Which were considering the announcement time on the TTL instead of block time itself.
https://github.com/0xPolygon/bor/blob/6acb9e10d628c88443e84a9df3d2f6fd0b568f85/consensus/bor/bor.go#L496-L508

### 3. Wait on `Prepare`

This is a bit new, but fits with this feature which is moving the wait we do from `Seal`. This gives us better conditions to build the background task to build the new block.

https://github.com/0xPolygon/bor/blob/6acb9e10d628c88443e84a9df3d2f6fd0b568f85/consensus/bor/bor.go#L959

consensus/bor/bor.go
cffls: Makes sense, thanks!

core/state/reader.go
cffls: Thank you for explaining!

marcello33 (approved):

core/state/reader.go
```diff
@@ -524,22 +592,36 @@ func (r *readerWithCache) storage(addr common.Address, slot common.Hash) (common
...
 type readerWithCacheStats struct {
 	*readerWithCache
+	role readerRole
+
 	accountHit  atomic.Int64
 	accountMiss atomic.Int64
 	storageHit  atomic.Int64
 	storageMiss atomic.Int64
+
+	// New: attribute PROCESS hits that were served by PREFETCH-origin entries.
```
cffls: Redundant word `New:` could be removed.

core/state/reader.go
```diff
@@ -454,68 +494,96 @@ func newReaderWithCache(reader Reader) *readerWithCache {
...
-		value  common.Hash
 		ok     bool
 		bucket = &r.storageBuckets[addr[0]&0x0f]
 	)
 	// Try to resolve the requested storage slot in the local cache
 	bucket.lock.RLock()
 	slots, ok := bucket.storages[addr]
 	if ok {
-		value, ok = slots[slot]
+		ent, ok2 := slots[slot]
```
cffls: Why not reuse `ok`?
```
ent, ok := slots[slot]
```

core/state/reader.go
```diff
@@ -92,10 +92,23 @@ type ReaderStats struct {
...
+// PrefetchStats exposes additional attribution stats for evaluating prefetch effectiveness.
+type PrefetchStats struct {
+	// Hits in PROCESS that came from PREFETCH-origin entries.
+	AccountHitFromPrefetch int64
+	StorageHitFromPrefetch int64
+	// Unique keys PREFETCH inserted into the shared local cache.
+	AccountInsert int64
+	StorageInsert int64
+	// Unique prefetched account keys that PROCESS actually used.
+	PrefetchAccountUsedByProcessUnique int64
```
cffls: Maybe `AccountHitFromPrefetchUnique` is more consistent with the other fields?
Also, why not have another field for counting unique storage hit, e.g. `StorageHitFromPrefetchUnique`?

miner/worker.go
```diff
@@ -1758,6 +1844,116 @@ func (w *worker) commitWork(interrupt *atomic.Int32, noempty bool, timestamp int
...
+				continue
+			}
+
+			transactions = append(transactions, tx)
+			gaspool.SubGas(tx.Gas())
+			txs.Shift()
+		}
+
+		block := types.NewBlock(header, &types.Body{Transactions: transactions}, nil, trie.NewStackTrie(nil))
+		result := prefetcher.Prefetch(block, throwaway, vm.Config{}, interruptPrefetch)
```
cffls: Here, we are the prefetch of the account/storage values from snapshot is guaranteed, but not their underlying tries. I think it is worth invoking `IntermediateRoot` at the end of the prefetch to warm up the trie as well, because trie fetching is the major bottleneck compared to the flat value fetching from the snapshot.

core/state/reader.go
lucca30: It was intended to make it more semantic, giving a single meaning to each var. But yeah I do also agree it has not many reason for this case

core/state/reader.go
lucca30: Nice, I liked this name. Changing here

lucca30: > Also, why not have another field for counting unique storage hit, e.g. StorageHitFromPrefetchUnique?

That's a good question! I actually tried it but, while `account` stores their data as a pointer, `storage` do it with a value.

Here the original develop code:

https://github.com/0xPolygon/bor/blob/f5ee550d21d51bb45132ea152ea64e14c77ecd96/core/state/reader.go#L427
https://github.com/0xPolygon/bor/blob/f5ee550d21d51bb45132ea152ea64e14c77ecd96/core/state/reader.go#L436


Changing `storage` for pointer could lead to a high heap allocation, which makes pressure on gc. Specially considering the volume of storage entries being a way more massive.

### Why do we need pointers to use this metric?

We need it to track the `usedByProcess` field. With a pointer, we can directly access the field of the pointer and use `atomic.CompareAndSwapUint32`. But if we try it with a map of values, it will demand a write lock on the map, since we need to get a copy, change it and write on the map.

So introducing this metric for `storage` without pointers means more locks on the cache map, which is not desirable. In the other hand, changing it for a pointer, as explained in the start, leads to GC pressure.

lucca30: I was thinking on maybe use a sampling approach where we just mark 1% to reduce the locks but I dont think it's worth.

lucca30: Also, it's good to consider the impact made by the current change:

For every `storage` cached, we previously just used `32 bytes` for `hash`. Now we are using `33 bytes` (there are no padding). I asked Claude to do a simulation an it summarized it as the impact:

Approach | GC Impact | Memory | Speed | Metrics
-- | -- | -- | -- | --
Pointer (*storageCacheEntry) | 💀💀💀 (GC death) | High | Slow | ✅ Works + Unique
Current (value) | ✅ None | +3% | -5% | ✅ Works
Original (Hash only) | ✅ None | Baseline | Baseline | ❌ No metrics

- +3% on memory
- -5% on speed

So wdyt about the tradeoff?

PS: worth to consider this measure it's just about the cache of storages. Which means the access on the storage, which is fast, will be slightly slow, and the memory (which is not the full bor) will be slightly higher.

Metric | Original | With Entry | Increase
-- | -- | -- | --
Per entry | 32 B | 33 B | +1 B (3.1%)
1M entries | 32 MB | 33 MB | +1 MB
10M entries | 320 MB | 330 MB | +10 MB

miner/worker.go
lucca30: That's an awesome catch!

It still something I'll need to do a deep research. As you can see, [Geth team previously had](https://github.com/ethereum/go-ethereum/pull/32134/changes#diff-cf3fbfff4358b57b0d24c889c0c442592814cab289ca416c78412379b80a66e3) `IntermediateRoot` being called by each transaction. But it was removed after we merged the upstream.

That said, our use case is slightly different: we are using **idle time to process it**. So I think it's worth bringing it back to compare whether it's worth or not.

Also, with interrupt flag we can process between each `IntermediateRoot` called, which means low impact when block is going go actually start.

lucca30: done by including it back. Tried to simulate locally but it was not possible

consensus/bor/bor.go
```diff
@@ -1059,6 +1091,23 @@ func (c *Bor) Prepare(chain consensus.ChainHeaderReader, header *types.Header) e
...
+		// if signer is not empty (RPC nodes have empty signer)
+		if currentSigner.signer != (common.Address{}) {
+			var err error
+			successionNumber, err = snap.GetSignerSuccessionNumber(currentSigner.signer)
+			if err != nil {
+				return err
+			}
+		}
+
+		if successionNumber == 0 {
```
cffls: If signer is empty, `successionNumber` will still be `0`, and we will end up waiting? It seems that RPC nodes shouldn't wait for this.

lucca30: the `waitOnPrepare` protects for the RPC use case. But you are completely right, I should just check it inside the if condition.

lucca30:
The latest Claude review pointed two issues, second one I agree and fixed. The first one is a false positive.

PR #2034, opened 2026-02-03
Enabling pip-66 back

lucca30:
# Description

Isolating the feature from original PR: https://github.com/0xPolygon/bor/pull/2031

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,7 +1,7 @@
 # This file configures github.com/golangci/golangci-lint.
 version: '2'
 run:
-  go: 1.25.6
+  go: 1.25.7
   tests: true
 linters:
   default: none
--- a/cmd/keeper/go.mod
+++ b/cmd/keeper/go.mod
@@ -1,6 +1,6 @@
 module github.com/ethereum/go-ethereum/cmd/keeper
 
-go 1.25.6
+go 1.25.7
 
 require (
 	github.com/ProjectZKM/Ziren/crates/go-runtime/zkvm_runtime v0.0.0-20260104020744-7268a54d0358
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -198,6 +198,11 @@ type BlockChainConfig struct {
 	// Maps account address to cache size in bytes
 	AddressCacheSizes map[common.Address]int
 
+	// PreloadRateLimit limits cache preload I/O in bytes per second per address.
+	// This prevents preloading from overwhelming the disk during sync.
+	// 0 = unlimited (legacy behavior), default = 1MB/s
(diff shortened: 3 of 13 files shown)


cffls:
Also, looks like this integration test failed: https://github.com/0xPolygon/bor/actions/runs/21642843860/job/62387000004?pr=2034#step:7:4364

tests/bor/fee_receiver_change_test.go
```diff
@@ -107,10 +107,10 @@ func TestCoinbaseRedirection(t *testing.T) {
 	}
 	sendTransactions(faucets[0], 5, earlyNonce)
 
-	// Wait for block 14 (just before Rio)
+	// Wait for block 15 (last pre-Rio block)
 	for {
 		block := nodes[0].BlockChain().CurrentBlock()
-		if block.Number.Uint64() >= 14 {
+		if block.Number.Uint64() >= 15 {
```
cffls, line 113: Thanks for fixing the test! However, I failed to understand why it is required to make this change. We haven't changed anything regarding the Rio HF number in the test, right?

lucca30: The issue observed here is because of the way the test is relying to capture the state and insert transactions(in terms of timing). It was assuming the original pattern which `Seal` has pause instead of `Prepare`.


**Before PIP-66:**
```
Prepare (fast) → [DELAY in Seal] → Block sealed
```

**After PIP-66:**
```
[DELAY in Prepare] → Prepare → Seal (fast) → Block sealed
```

### Transaction Distribution Change

**Without PIP-66:**
- Block 2: 5 transactions
- Block 15: **0 transactions**
- Block 16: **5 transactions** (Rio activation block)

**With PIP-66:**
- Block 2: 5 transactions
- Block 15: **5 transactions** (delayed by timing change)
- Block 16: **0 transactions**


So after the change **block 15 started including txs**, impacting on validator gains before Rio.


**Why it passed without PIP-66:** Block 15 had 0 transactions, so no validator gain
**Why it failed with PIP-66:** Block 15 had 5 transactions, validators correctly received fees, but test treated this as "post-Rio" gain


### Fixed Test (Passes with PIP-66)

```go
// Wait for block 15 (last pre-Rio block)
// Read state at block 15 (using StateAt(block15.Root()))
// ...
// Wait for block 17
// Read state at block 17 (using StateAt(block17.Root()))
// Calculate validator gain from block 15 to 17
```

PR #2038, opened 2026-02-04
Clean up lastRebroadcast on transaction rejection

LarryArnault45:
Removes transaction hash from lastRebroadcast map when rejected during promotion, preventing memory accumulation.


cffls:
Thanks for the fix! Could you rebase on top of `develop` branch?

kamuikatsurgi (approved):

lucca30:
@claude Can you analyze the changes. Deep looking if there is any security breach introduced, check if the changes helps or somehow that can potentially lead to any bug?

PR #2040, opened 2026-02-04
lisovo: CLZ opcode and P256 gas cost changes

kamuikatsurgi:
# Description

Introduces `TBDHF` to support CLZ opcode and P256 gas cost changes.

### DIFF
--- a/internal/cli/server/chains/amoy.go
+++ b/internal/cli/server/chains/amoy.go
@@ -39,7 +39,7 @@ var amoyTestnet = &Chain{
 				MadhugiriBlock:    big.NewInt(28899616),
 				MadhugiriProBlock: big.NewInt(29287400),
 				DandeliBlock:      big.NewInt(31890000),
-				LisovoBlock:       nil,
+				LisovoBlock:       big.NewInt(33634700),
 				StateSyncConfirmationDelay: map[string]uint64{
 					"0": 128,
 				},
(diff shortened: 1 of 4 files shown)


internal/cli/server/chains/amoy.go
```diff
@@ -39,6 +39,7 @@ var amoyTestnet = &Chain{
 				MadhugiriBlock:    big.NewInt(28899616),
 				MadhugiriProBlock: big.NewInt(29287400),
 				DandeliBlock:      big.NewInt(31890000),
+				TBDHFBlock:        nil,
```
pratikspatil024: (here and at other places)
better to leave it for when we are scheduling the HF.
Also, there's a json file where we need to set this.

lucca30:
I'm going to ask to hold this PR until I merge mine into this one. So we introduce both changes in a single merge to develop

internal/cli/server/chains/amoy.go
kamuikatsurgi: I think keeping it nil doesn't hurt.

kamuikatsurgi: nil anyways means hf isn't activated.

pratikspatil024: > there's a json file where we need to set this

better to add it there as well then.

pratikspatil024: for consistency.

lucca30:
Here's the HF https://github.com/0xPolygon/bor/pull/2041

cffls (approved):

internal/cli/server/chains/amoy.go
kamuikatsurgi: this is now done.

pratikspatil024 (approved):

PR #2042, opened 2026-02-05
Fix more race conditions detected by -race

cffls:
# Description

Fix more race conditions detected by -race

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -222,6 +222,26 @@ type TraceConfig struct {
 	BorTx           *bool
 }
 
+// deepCopyTraceConfig returns a deep copy of the given TraceConfig so that
+// each goroutine can safely mutate its own copy without racing.
+func deepCopyTraceConfig(config *TraceConfig) TraceConfig {
+	cpy := *config
+	if config.Config != nil {
+		loggerCfg := *config.Config
+		cpy.Config = &loggerCfg
+	}
+	if config.BorTx != nil {
+		cpy.BorTx = newBoolPtr(*config.BorTx)
+	}
+	if config.BorTraceEnabled != nil {
+		cpy.BorTraceEnabled = newBoolPtr(*config.BorTraceEnabled)
+	}
+	if config.TracerConfig != nil {
+		cpy.TracerConfig = append(json.RawMessage{}, config.TracerConfig...)
+	}
+	return cpy
+}
+
 // TraceCallConfig is the config for traceCall API. It holds one more
 // field to override the state for tracing.
 type TraceCallConfig struct {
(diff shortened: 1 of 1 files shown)


miner/worker.go
lucca30, line 810: i see some `w.current` reads here without the introduced `w.currentMu`, is it ok?

eth/tracers/api.go
```diff
@@ -383,11 +383,13 @@ func (api *API) traceChain(start, end *types.Block, config *TraceConfig, closed
 
 					var err error
 
+					// Make a copy of config for this transaction to avoid race conditions
+					txConfig := *config
 					if stateSyncPresent && i == len(txs)-1 && includeStateSyncTx {
-						config.BorTx = newBoolPtr(true)
+						txConfig.BorTx = newBoolPtr(true)
 					}
```
lucca30: This one I think that could run into an issue because `TraceConfig` struct holds pointers, not values. So let me try to do an example with `BorTx *bool`.

If two routines make a copy of `*config`, they still pointing to the same `txConfig.BorTx` address on memory. Looking a bit on the code, doesn't seems they do any change on the pointer, instead, they always instantiate a new one. But anyway in a future change this can lead to an issue

eth/tracers/api.go
```diff
@@ -863,11 +865,13 @@ func (api *API) traceBlock(ctx context.Context, block *types.Block, config *Trac
 
 				var err error
 
+				// Make a copy of config for this transaction to avoid race conditions
+				txConfig := *config
```
lucca30: same here

core/txpool/legacypool/list.go
```diff
@@ -811,6 +816,8 @@ func (l *pricedList) SetBaseFee(baseFee *big.Int) {
 	if baseFee != nil {
 		base.SetFromBig(baseFee)
 	}
+	l.reheapMu.Lock()
 	l.urgent.baseFee = base
+	l.reheapMu.Unlock()
 	l.Reheap()
 }
```
lucca30, line 823: Do you know if the time window between `l.reheapMu.Unlock()` here and the `l.reheapMu.Lock()` inside `Reheap` function is okay?

This window may lead to a inconsitency. If possible, would be good not unlocking until reheap is done.

miner/worker.go
cffls: The currentMu lock was added specifically to protect against the test goroutine reading w.current while the worker goroutine writes to it. Within mainLoop() itself, there's no concurrent access as it's all sequential on one goroutine.

lucca30: Nice, makes sense

eth/tracers/api.go
cffls: Good point, I will make a function to create a deep copy of the config.

core/txpool/legacypool/list.go
cffls: `reheapMu` is also acquired in `l.Reheap()`. If we don't unlock it before calling reheap, dead lock will happen.

lucca30: Makes sense about the deadlock 

My concern is about `urgent` state consistency.

Right now we update `l.urgent.baseFee`, unlock, and then re-lock inside `Reheap()`. The window is tiny, but another goroutine could still observe an inconsistent `baseFee` vs heap state.

Not sure this causes real issues today, but it feels fragile. One option could be to move the `baseFee` update into `Reheap()` (optionally), so everything happens under the same lock.

PR #2050, opened 2026-02-11
consensus/bor, eth: add Heimdall HTTP/gRPC/WS failover with health recovery

pratikspatil024:
# Description

When a node's Heimdall endpoint goes down, the bor node loses access to span/checkpoint/milestone data. For block producers this can cause unnecessary span rotations, reducing chain stability. Automatic failover to a secondary Heimdall endpoint keeps the node operational during transient outages without manual intervention.

  - Adds multi-endpoint failover for Heimdall HTTP, gRPC, and WebSocket clients. Existing `--bor.heimdall`, `--bor.heimdallgRPC`, and `--bor.heimdallWS` flags now accept comma-separated URLs (e.g. `"url1,url2,url3"`).
  - Introduces a shared `HealthRegistry` that runs background health probes against all configured endpoints, tracks consecutive successes, enforces a promotion cooldown, and proactively switches away from unhealthy endpoints.
  - On a failover-eligible error (network timeouts, 5xx), `MultiHeimdallClient` cascades through endpoints in priority order (healthy+cooled > healthy+uncooled > unhealthy). Client errors (4xx) and `503 ServiceUnavailable` are not retried
   since secondaries would return the same response.
  - The WS client gains the same failover support, reconnecting to the next healthy endpoint when the active connection drops.
  - Adds Prometheus metrics for failover switches, probe attempts/successes, active endpoint index, and healthy endpoint count (both HTTP/gRPC and WS).

  ## Default timeouts & thresholds
  | Parameter | HTTP/gRPC | WebSocket |
  |---|---|---|
  | Per-attempt timeout | 30s | — |
  | Probe timeout | 5s | 10s |
  | Health check interval | 10s | 10s |
  | Consecutive success threshold | 3 | 3 |
  | Promotion cooldown | 60s | 60s |
  | Reconnect delay | — | 10s |

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/consensus/bor/heimdall/failover_client_test.go
+++ b/consensus/bor/heimdall/failover_client_test.go
@@ -122,6 +122,55 @@ func (m *mockHeimdallClient) Close() {
 	}
 }
 
+// testConnErr is a reusable connection-refused error for tests.
+var testConnErr = &net.OpError{Op: "dial", Net: "tcp", Err: errors.New("connection refused")}
+
+// newConnRefusedMock creates a mock where both API calls and health probes always fail.
+func newConnRefusedMock() *mockHeimdallClient {
+	return &mockHeimdallClient{
+		getSpanFn: func(_ context.Context, _ uint64) (*types.Span, error) {
+			return nil, testConnErr
+		},
+		fetchStatusFn: func(_ context.Context) (*ctypes.SyncInfo, error) {
+			return nil, testConnErr
+		},
+	}
+}
+
+// newToggleMock creates a mock whose API calls and health probes fail when down.Load() is true.
+func newToggleMock(down *atomic.Bool) *mockHeimdallClient {
+	return &mockHeimdallClient{
+		getSpanFn: func(_ context.Context, spanID uint64) (*types.Span, error) {
+			if down.Load() {
+				return nil, testConnErr
+			}
+			return &types.Span{Id: spanID}, nil
+		},
+		fetchStatusFn: func(_ context.Context) (*ctypes.SyncInfo, error) {
+			if down.Load() {
+				return nil, testConnErr
+			}
+			return &ctypes.SyncInfo{}, nil
+		},
+	}
+}
+
+// newProbeToggleMock creates a mock where API calls always fail but health probes
+// succeed when down.Load() is false.
+func newProbeToggleMock(down *atomic.Bool) *mockHeimdallClient {
+	return &mockHeimdallClient{
+		getSpanFn: func(_ context.Context, _ uint64) (*types.Span, error) {
+			return nil, testConnErr
+		},
+		fetchStatusFn: func(_ context.Context) (*ctypes.SyncInfo, error) {
+			if down.Load() {
+				return nil, testConnErr
+			}
+			return &ctypes.SyncInfo{}, nil
+		},
+	}
+}
+
 // newInstantMulti creates a MultiHeimdallClient with instant health registry
 // behavior: consecutiveThreshold=1, promotionCooldown=0, fast health-check interval.
 func newInstantMulti(clients ...Endpoint) *MultiHeimdallClient {
@@ -296,20 +345,7 @@ func TestFailover_ProbeBackToPrimary(t *testing.T) {
 	primaryDown := atomic.Bool{}
 	primaryDown.Store(true)
 
-	primary := &mockHeimdallClient{
-		getSpanFn: func(_ context.Context, spanID uint64) (*types.Span, error) {
-			if primaryDown.Load() {
-				return nil, &net.OpError{Op: "dial", Net: "tcp", Err: errors.New("connection refused")}
-			}
-			return &types.Span{Id: spanID}, nil
-		},
-		fetchStatusFn: func(_ context.Context) (*ctypes.SyncInfo, error) {
-			if primaryDown.Load() {
-				return nil, &net.OpError{Op: "dial", Net: "tcp", Err: errors.New("connection refused")}
-			}
-			return &ctypes.SyncInfo{}, nil
-		},
-	}
+	primary := newToggleMock(&primaryDown)
 	secondary := &mockHeimdallClient{}
 
 	fc := newInstantMulti(primary, secondary)
@@ -335,14 +371,7 @@ func TestFailover_ProbeBackToPrimary(t *testing.T) {
 }
 
 func TestFailover_ProbeBackFails(t *testing.T) {
-	primary := &mockHeimdallClient{
-		getSpanFn: func(_ context.Context, _ uint64) (*types.Span, error) {
-			return nil, &net.OpError{Op: "dial", Net: "tcp", Err: errors.New("connection refused")}
-		},
-		fetchStatusFn: func(_ context.Context) (*ctypes.SyncInfo, error) {
-			return nil, &net.OpError{Op: "dial", Net: "tcp", Err: errors.New("connection refused")}
-		},
-	}
+	primary := newConnRefusedMock()
 	secondary := &mockHeimdallClient{}
 
 	fc := newInstantMulti(primary, secondary)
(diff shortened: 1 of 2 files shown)


pratikspatil024:
> ## Code review
> ### Bug: Missing timeout for secondary client calls
> **File:** [consensus/bor/heimdall/failover_client.go](https://github.com/0xPolygon/bor/blob/1629daf7edb6a575b62fac3b1cfad72ac79682eb/consensus/bor/heimdall/failover_client.go#L147-L179)
> 
> The primary client calls use a 30-second timeout via `context.WithTimeout` (lines 122-123 and 157-158), but the secondary client calls at lines 148, 153, and 178 use the original `ctx` directly without any timeout.
> 
> If a caller passes `context.Background()` (which has no deadline) and the secondary endpoint is also unresponsive, these calls will hang indefinitely. This defeats the purpose of having a failover mechanism since the system could still hang when both endpoints are problematic.
> 
> **Affected lines:**
> 
> * Line 148: `return fn(ctx, f.clients[1])` - after probe fails
> * Line 153: `return fn(ctx, f.clients[1])` - on secondary, not yet time to probe
> * Line 178: `return fn(ctx, f.clients[1])` - failover to secondary
> 
> **Suggested fix:** Apply the same timeout to secondary calls:
> 
> ```go
> // For each secondary call, wrap with timeout:
> subCtx, cancel := context.WithTimeout(ctx, f.attemptTimeout)
> defer cancel()
> return fn(subCtx, f.clients[1])
> ```

@claude - This is intentional. The 30s timeout on primary exists specifically to bound the failover decision.
Once we're on secondary, there's nothing left to switch to, so a timeout would just cause unnecessary failures.

Adding a timeout on secondary would actually be harmful in two ways:

1. Both endpoints down - the call fails after 30s, but the caller (`FetchWithRetry`) just retries the whole thing. Each retry cycle would now burn 60s (30s primary + 30s secondary) for no benefit.
2. Secondary is healthy but slow - a large state sync query that legitimately takes >30s would get killed prematurely.

Callers in Bor never pass `context.Background()`, they always use contexts tied to the node lifecycle (via `closeCh` or consensus engine contexts), so there's always a cancellation path. Added a comment clarifying this rationale.

cmd/utils/bor_flags.go
```diff
@@ -23,6 +23,13 @@ var (
 		Value: "http://localhost:1317",
 	}
 
+	// HeimdallSecondaryURLFlag flag for secondary heimdall url (failover)
+	HeimdallSecondaryURLFlag = &cli.StringFlag{
+		Name:  "bor.heimdall.secondary",
```
cffls: Can we accept a list of heimdall url? It will make the infra more flexible. Or even better, we can reuse the original flag for heimdall, and make it accept either a single url (format used today) or a list of url separated by a comma.

pratikspatil024: It should be doable, I'll work on that. Thanks!

pratikspatil024: Done [here](https://github.com/0xPolygon/bor/pull/2050/commits/25b7cad29555147361c787e2ffd5f1e4915aca17).

pratikspatil024:
@claude - I updated the PR by extending the failover logic to grpc and ws client. Also now accepting a list of urls (multiple) instead of just secondary url. Review this PR again.

consensus/bor/heimdall/failover_client.go
```diff
@@ -0,0 +1,277 @@
...
+	FetchMilestone(ctx context.Context) (*milestone.Milestone, error)
+	FetchMilestoneCount(ctx context.Context) (int64, error)
+	FetchStatus(ctx context.Context) (*ctypes.SyncInfo, error)
+	Close()
+}
+
+// FailoverHeimdallClient wraps N heimdall clients (primary at index 0, failovers
+// at 1..N-1) and transparently cascades through them when the active client is
+// unreachable. After a cooldown period it probes the primary again.
+type FailoverHeimdallClient struct {
```
cffls: Does `MultiHeimdallClient` sound better? Since this structure does not only contain the failover client, but also the primary client.

consensus/bor/heimdall/failover_client.go
```diff
@@ -0,0 +1,277 @@
...
+// If on a non-primary client past the cooldown, it probes the primary first.
+func callWithFailover[T any](f *FailoverHeimdallClient, ctx context.Context, fn func(context.Context, Endpoint) (T, error)) (T, error) {
+	f.mu.Lock()
+	active := f.active
+	shouldProbe := active != 0 && time.Since(f.lastSwitch) >= f.cooldown
+	f.mu.Unlock()
+
+	// If on a non-primary client and cooldown has elapsed, probe primary
+	if shouldProbe {
+		subCtx, cancel := context.WithTimeout(ctx, f.attemptTimeout)
```
cffls: This may be risky if primary is still not up after the cooldown, resulting a lag. Instead of querying on receiving a new request, we can probe primary in a background goroutine, and when it is back up, update the active accordingly.

consensus/bor/heimdall/failover_client.go
```diff
@@ -0,0 +1,277 @@
...
+
+	if !isFailoverError(err, ctx) {
+		var zero T
+		return zero, err
+	}
+
+	// Cascade through clients [1, 2, ..., N-1]
+	log.Warn("Heimdall failover: primary failed, cascading to next client", "err", err)
+
+	return cascadeClients(f, ctx, fn, 0, err)
```
cffls: Same as above - by the time when we discover a failure in backup client, it might be too late to switch over to the next one if the request times out, causing a span rotation. We can always probe the active client in a background goroutine, which will automatically update the `active` if something goes wrong with the current one.

eth/ethconfig/config.go
```diff
@@ -45,6 +46,25 @@ import (
...
+func parseURLs(s string) []string {
+	if s == "" {
+		return nil
+	}
+
+	parts := strings.Split(s, ",")
+
+	var out []string
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
```
marcello33, line 60: Shall we verify these are valid URLs or we leave it up to the operator?

eth/ethconfig/config.go
```diff
@@ -325,33 +345,62 @@ func CreateConsensusEngine(chainConfig *params.ChainConfig, ethConfig *Config, d
...
+				}
+
+				if len(heimdallClients) == 0 {
 					heimdallClient = heimdall.NewHeimdallClient(ethConfig.HeimdallURL, ethConfig.HeimdallTimeout)
+				} else if len(heimdallClients) == 1 {
+					heimdallClient = heimdallClients[0]
 				} else {
-					heimdallClient = grpcClient
+					heimdallClient = heimdall.NewFailoverHeimdallClient(heimdallClients...)
+					log.Info("Heimdall failover enabled", "endpoints", len(heimdallClients))
```
marcello33: If we change to `multiple` instead of failover, worth to update this log as well. It could be misleading.

eth/ethconfig/config.go
```diff
@@ -325,33 +345,62 @@ func CreateConsensusEngine(chainConfig *params.ChainConfig, ethConfig *Config, d
...
-				heimdallWSClient, err = heimdallws.NewHeimdallWSClient(ethConfig.HeimdallWSAddress)
+
+			if len(wsAddrs) > 0 {
+				heimdallWSClient, err = heimdallws.NewHeimdallWSClient(wsAddrs...)
 				if err != nil {
 					return nil, err
 				}
+
+				if len(wsAddrs) > 1 {
+					log.Info("Heimdall WS failover enabled", "endpoints", len(wsAddrs))
```
marcello33: Same as before

pratikspatil024: clarified [here](https://github.com/0xPolygon/bor/pull/2050/changes/b0cc4f5e2495d1ca85e7613222e72b8e06e86b3e), thanks

eth/ethconfig/config.go
pratikspatil024: clarified [here](https://github.com/0xPolygon/bor/pull/2050/changes/b0cc4f5e2495d1ca85e7613222e72b8e06e86b3e), thanks

eth/ethconfig/config.go
pratikspatil024: I think it should be ok leaving it up to the operator.

consensus/bor/heimdall/failover_client.go
pratikspatil024: added a background health check [here](https://github.com/0xPolygon/bor/pull/2050/commits/3825a5d774cbb784ad4c90b3c5ba3e391673dcec). Thanks!

consensus/bor/heimdall/failover_client.go
pratikspatil024: added a background health check [here](https://github.com/0xPolygon/bor/pull/2050/commits/3825a5d774cbb784ad4c90b3c5ba3e391673dcec). Thanks!

consensus/bor/heimdall/failover_client.go
pratikspatil024: updated [here](https://github.com/0xPolygon/bor/pull/2050/commits/6cc879a686385396a37e00017c796a38567852c4), thanks!

consensus/bor/heimdall/failover_client.go
pratikspatil024: did some updates [here](https://github.com/0xPolygon/bor/pull/2050/changes/be4fe9dc2fa77418cea9d528288ff56027780d15)

consensus/bor/heimdall/failover_client.go
pratikspatil024: did some updates [here](https://github.com/0xPolygon/bor/pull/2050/changes/be4fe9dc2fa77418cea9d528288ff56027780d15)

pratikspatil024:
@claude - review this PR again.

PR #2053, opened 2026-02-13
fix: remove kzgPointEvaluation precompile from PoS precompiled contracts

kamuikatsurgi:
# Description

kzgPointEvaluation (0x0a) accidentally added to PrecompiledContractsPrague.

The upstream merge of go-ethereum v1.16.4 (commit 9151b05ef3) introduced kzgPointEvaluation at address 0x0a into PrecompiledContractsPrague.

How does this cause the 3140 gas discrepancy?

The contract at 0x863134579e4812f9d78081e9f519fae9d01f2a10 (a "Buy And Free" gas token) has a function (selector 0x31) that repeatedly CALLs address 0x0a in a tight loop to burn gas: CALL(gas=GAS, to=0x0a, value=0, inOffset=0, inSize=0xc0, retOffset=0xc0, retSize=0x40)

v2.5.8 (correct): Not a precompile
v2.6.0-beta (broken): kzgPointEvaluation precompile


core/vm/contracts.go
```diff
@@ -236,7 +233,6 @@ var PrecompiledContractsLisovo = PrecompiledContracts{
 	common.BytesToAddress([]byte{0x07}):       &bn256ScalarMulIstanbul{},
 	common.BytesToAddress([]byte{0x08}):       &bn256PairingIstanbul{},
 	common.BytesToAddress([]byte{0x09}):       &blake2F{},
-	common.BytesToAddress([]byte{0x0a}):       &kzgPointEvaluation{},
```
kamuikatsurgi, line 239: We can remove this from other HFs, but not lisovo, which would need another HF to disable this precompile.

PR #2057, opened 2026-02-17
fix(core): cap verifyPendingHeaders to prevent OOM from unbounded header reads

kamuikatsurgi:
# Description

`verifyPendingHeaders` re-reads ALL headers from the last whitelisted milestone to the current head every 2 seconds. When the gap is large (e.g., during sync or when milestones can't be whitelisted), this causes unbounded memory growth, leading to OOM.

## pprof analysis
| `ReadHeader` in pprof | 5 GB (60% of heap) |
|---|---|
| `verifyPendingHeaders` cumulative | 7.1 GB (85% of heap) |

## Fix
Cap the header collection loop to at most `DefaultSpanLength + 1` blocks from the current head. This covers 2 spans of reorg depth, the maximum reorg protection needed while keeping memory bounded.

### DIFF
--- a/core/blockchain_test.go
+++ b/core/blockchain_test.go
@@ -4829,24 +4829,23 @@ func testHeaderVerificationLoop(t *testing.T, scheme string) {
 	// Test case 1: Valid chain - no rewinds should happen
 	t.Run("ValidChain", func(t *testing.T) {
 		engine := ethash.NewFaker()
-		genesis := &Genesis{
-			Config:  params.TestChainConfig,
-			BaseFee: big.NewInt(params.InitialBaseFee),
-		}
 
-		// Create a mock validator that has a finalized block at height 3
-		mockValidator := &mockChainValidator{
-			hasMilestone:    true,
-			milestoneNumber: 3,
-			milestoneHash:   common.HexToHash("0x123"),
+		config := *params.TestChainConfig
+		config.Bor = &params.BorConfig{
+			RioBlock: big.NewInt(0),
+		}
+		genesis := &Genesis{
+			Config:  &config,
+			BaseFee: big.NewInt(params.InitialBaseFee),
 		}
 
 		// Generate blocks
 		_, blocks, _ := GenerateChainWithGenesis(genesis, engine, 8, nil)
 
-		// Create blockchain with mock validator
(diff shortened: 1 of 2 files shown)


marcello33 (approved):

core/blockchain.go
```diff
@@ -4232,6 +4227,8 @@ func (bc *BlockChain) startHeaderVerificationLoop() {
 
 // verifyPendingHeaders checks headers after the latest finalized block
 // and rewinds the chain if invalid headers are found.
+// Verification is capped to params.DefaultSpanLength + 1 blocks from the current head.
+// This covers atleast 2 spans for max reorg protection without unbounded memory growth.
 func (bc *BlockChain) verifyPendingHeaders() {
```
cffls, line 4236: This function was supposed to verify pending/unfinalized blocks seemed by the consensus, not by the local node. Even though the milestone is not whitelisted locally, it has already been settled onchain. I think a simpler and safer approach is to directly query the latest milestone block number from heimdall, which should've been much more recent compared to local finalized block, and use it as the start block for checking.

kamuikatsurgi: But what if the local heimdall is not in sync, and its latest milestone is way behind the current bor head? This issue would cause again bor to go OOM, which would mean bor would need to wait till heimdall is fully synced or its milestone is somewhere near the local head for this func to not cause OOM.

cffls: Yes that's the expectation. Today, bor isn't allowed to sync before heimdall is in sync, see this check here: https://github.com/0xPolygon/bor/blob/65f1507393b6b120a28804e9a5269c97d238ac35/consensus/bor/bor.go#L1694-L1695

kamuikatsurgi: Gotcha, will refactor accordingly. Thanks!

kamuikatsurgi: Refactored.

cffls:
Thanks! The change looks good. Unit test and integration test failed, may need to take a look.

kamuikatsurgi:
> Thanks! The change looks good. Unit test and integration test failed, may need to take a look.

Thanks, fixed the tests. Will merge it once all the tests pass.

PR #2059, opened 2026-02-17
consensus, crypto, p2p: fix issues 44, 47, 48, 49

marcello33:
# Description

Fixes for issues 44, 47, 48 and 49

### DIFF
--- a/consensus/bor/heimdallgrpc/state_sync.go
+++ b/consensus/bor/heimdallgrpc/state_sync.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	stateSyncTotalTimeout = 5 * time.Minute
+	stateSyncTotalTimeout = 1 * time.Minute
 )
 
 func (h *HeimdallGRPCClient) StateSyncEvents(ctx context.Context, fromID uint64, to int64) ([]*clerk.EventRecordWithTime, error) {


consensus/bor/heimdallgrpc/state_sync.go
```diff
@@ -14,17 +14,21 @@ import (
...
+)
+
 func (h *HeimdallGRPCClient) StateSyncEvents(ctx context.Context, fromID uint64, to int64) ([]*clerk.EventRecordWithTime, error) {
 	log.Info("Fetching state sync events", "fromID", fromID, "to", to)
 
 	var err error
 
-	ctxWithTimeout, cancel := context.WithTimeout(ctx, defaultTimeout)
-	defer cancel()
+	totalCtx, totalCancel := context.WithTimeout(ctx, stateSyncTotalTimeout)
```
kamuikatsurgi: Could we have better var names here?

kamuikatsurgi (approved):
Just a small nit

consensus/bor/heimdallgrpc/state_sync.go
marcello33: smth like `global*`?

kamuikatsurgi: maybe like `stateSyncEventsCtx`?

consensus/bor/heimdallgrpc/state_sync.go
```diff
@@ -14,17 +14,21 @@ import (
 	"github.com/0xPolygon/heimdall-v2/x/clerk/types"
 )
 
+const (
+	stateSyncTotalTimeout = 5 * time.Minute
```
kamuikatsurgi: I think this timeout is way too much, given the 2 sec block time. How'd we get this timeout? Would bor stall for 5 mins for state sync if somehow we have a case like that?

consensus/bor/heimdallgrpc/state_sync.go
marcello33: Changed to
`globalCtx, cancel := context.WithTimeout(ctx, stateSyncTotalTimeout)`
Will push once CI is complete, thanks

consensus/bor/heimdallgrpc/state_sync.go
```diff
@@ -33,36 +37,47 @@ func (h *HeimdallGRPCClient) StateSyncEvents(ctx context.Context, fromID uint64,
...
+			ToTime:     time.Unix(to, 0),
+			Pagination: pagination,
+		}
 
-	res, err := h.clerkQueryClient.GetRecordListWithTime(ctx, req)
-	if err != nil {
-		return nil, err
-	}
+		var res *types.RecordListWithTimeResponse
+		pageCtx, pageCancel := context.WithTimeout(ctx, defaultTimeout)
```
kamuikatsurgi, line 52: If we lower the `stateSyncTotalTimeout`, we could just reuse the `totalCtx` here or whatever name it's changed to.

consensus/bor/heimdallgrpc/state_sync.go
kamuikatsurgi: I think in that case, liveness of the bor chain is much more important than the bridge events.

marcello33: Down to 1m

manav2401 (approved):

PR #2062, opened 2026-02-18
feat: add tracing support for StateSyncTx

milando12:
# Description

Adding tracing support for `StateSyncTx`.

# Changes

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

Ran the Node with this code for few days - no problems showed up and I verified that `StateSyncTx` trace that we get is correct.


consensus/bor/bor.go
```diff
@@ -1069,27 +1069,54 @@ func (c *Bor) Finalize(chain consensus.ChainHeaderReader, header *types.Header,
...
 		start := time.Now()
 		cx := statefull.ChainContext{Chain: chain, Bor: c}
+
+		// Start tracing StateSyncTx (if present in the block body)
+		var stateSyncTx *types.Transaction
+		if c.config.IsMadhugiri(header.Number) && len(body.Transactions) > 0 {
+			lastTx := body.Transactions[len(body.Transactions)-1]
+			if lastTx.Type() == types.StateSyncTxType {
+				stateSyncTx = lastTx
+				if hooks := vmCfg.Tracer; hooks != nil && hooks.OnTxStart != nil {
```
maoueh: This should be moved probably just before `if c.config.IsMadhugiri(header.Number) && len(body.Transactions) > 0 {` so not tracing cost is paid if there is no tracer active.

consensus/bor/bor.go
```diff
@@ -1069,27 +1069,54 @@ func (c *Bor) Finalize(chain consensus.ChainHeaderReader, header *types.Header,
...
+		if cfg := hc.GetVMConfig(); cfg != nil {
+			vmCfg = *cfg
+		}
+	}
+
 	if IsSprintStart(headerNumber, c.config.CalculateSprint(headerNumber)) {
 		start := time.Now()
 		cx := statefull.ChainContext{Chain: chain, Bor: c}
+
+		// Start tracing StateSyncTx (if present in the block body)
```
maoueh: I think we might have integration tests locally (we maintain our own Polygon Live Tracing branch for now).

I'll check if I could add that on top of your PR.

milando12: That sounds great. Feel free to add them on top of this PR whenever you're ready.

consensus/bor/bor.go
milando12: Makes sense - fixed in commit ee7c1221e866abec2ac618a0b003472eedc9f75b

pratikspatil024:
@claude - review this PR.

kamuikatsurgi:
> @claude - review this PR.

It won't be able to review this PR is from a fork, and hence it doesn't have perms to fetch the branch.

consensus/bor/bor.go
```diff
@@ -1139,20 +1152,37 @@ func (c *Bor) Finalize(chain consensus.ChainHeaderReader, header *types.Header,
...
 	)
 
+	vmCfg := extractVMConfig(chain)
+
 	if IsSprintStart(headerNumber, c.config.CalculateSprint(headerNumber)) {
 		start := time.Now()
 		cx := statefull.ChainContext{Chain: chain, Bor: c}
+
+		// Start tracing StateSyncTx (if present in the block body)
+		if hooks := vmCfg.Tracer; hooks != nil && hooks.OnTxStart != nil {
```
marcello33: it seems like this starts a single tx trace before `checkAndCommitSpan` and the `CommitStates` loop, but `statefull.ApplyMessage` does not emit per-call `OnTxStart`/`OnTxEnd` hooks. Can this cause any issue on sprint blocks (span commits and multiple state-syncs) breaking tracers that keep single tx state? If that's the case, maybe the hooks need to wrap each system call, not the whole finalize.

consensus/bor/bor.go
```diff
@@ -1324,7 +1362,7 @@ func (c *Bor) FinalizeAndAssemble(chain consensus.ChainHeaderReader, header *typ
 			StateSyncData: stateSyncData,
 		})
 		body.Transactions = append(body.Transactions, stateSyncTx)
-		receipts = insertStateSyncTransactionAndCalculateReceipt(stateSyncTx, header, body, state, receipts)
+		receipts = insertStateSyncTransactionAndCalculateReceipt(stateSyncTx, header, body, state, receipts, vmCfg)
```
marcello33: Can we make the tracer lifecycle symmetric between `Finalize` and `FinalizeAndAssemble`? I think there's a difference there wrt `insertStateSyncTransactionAndCalculateReceipt` calls

cmd/cli/main.go
```diff
@@ -5,6 +5,11 @@ import (
 
 	"github.com/ethereum/go-ethereum/internal/cli"
 	"github.com/ethereum/go-ethereum/params"
+
+	// Force-load the tracer engines to trigger registration
+	_ "github.com/ethereum/go-ethereum/eth/tracers/js"
```
marcello33: Do we need to register `js` and `native` tracers here? IIUC, you're only leveraging `LiveDirectory` in this PR.

docs/cli/example_config.toml
```diff
@@ -31,11 +31,13 @@ enable-private-tx = true # Enable private transaction relay
...
-  json = false                    # Format logs with JSON
-  backtrace = ""                  # Request a stack trace at a specific logging statement (e.g. "block.go:271")
-  debug = true                    # Prepends log messages with call-site location (file and line number)
-  enable-block-tracking = false   # Enables additional logging of information collected while tracking block lifecycle
+vmodule = ""                               # Per-module verbosity: comma-separated list of <pattern>=<level> (e.g. eth/*=5,p2p=4)
+json = false                               # Format logs with JSON
+backtrace = ""                             # Request a stack trace at a specific logging statement (e.g. "block.go:271")
+debug = true                               # Prepends log messages with call-site location (file and line number)
+enable-block-tracking = false              # Enables additional logging of information collected while tracking block lifecycle
+vmtrace = "supply"                         # Name of a tracer to record internal VM operations during blockchain synchronization (costly)
```
marcello33: Shouldn't `VMTrace` and  `VMTraceJsonConfig` be top level fields? I see they’re under `[log]` here

consensus/bor/statefull/processor.go
```diff
@@ -70,7 +70,7 @@ func (m Callmsg) Data() []byte         { return m.CallMsg.Data }
 func GetSystemMessage(toAddress common.Address, data []byte) Callmsg {
 	return Callmsg{
 		ethereum.CallMsg{
-			From:     systemAddress,
+			From:     SystemAddress,
```
marcello33: Can we use a getter here instead of exposing?

marcello33:
Can we ad integration tests that verify the trace output for a block containing a StateSyncTx?
Also, I think you need to solve the conflicts with base branch

manav2401:
Closing this in favor of https://github.com/0xPolygon/bor/pull/2172. Some tests won't work in this PR as it's created from a fork.

PR #2063, opened 2026-02-18
consensus, core, eth, internal, miner, rpc, tests: implement additional rpc methods

marcello33:
# Description

This PR implements a set of RPC methods in the `bor_` namespace to match the `erigon_` namespace APIs, along with some additional `eth_` and `debug_` endpoints.

`bor_` namespace additions: 
- `bor_blockNumber`
- `bor_forks` 
- `bor_getHeaderByNumber`
- `bor_getHeaderByHash`
- `bor_getLogsByHash`, 
- `bor_getBlockReceiptsByBlockHash`
- `bor_getBlockByTimestamp`
- `bor_getLogs`
- `bor_getLatestLogs`
- `bor_getBalanceChangesInBlock`

These mirror the equivalent `erigon_*` methods and behave identically with two minor (intentional) differences: 
1. `bor_getBlockByTimestamp` accepts both hex strings and decimal integers for the timestamp parameter, which is more permissive than Erigon but necessary to avoid breaking existing callers. 
2. `bor_getLatestLogs` returns the block timestamp under the field name `blockTimestamp` rather than Erigon's `timestamp` (changing this would break existing clients).

Also, `bor_getLogs` implements the standard `FilterCriteria` type with strict hex validation, mutual exclusion between blockHash and block range params, and support for both single-address and address-array formats.

Under `eth_` namespace, the following have been implemented:
- `eth_etherbase` 
- `eth_hashrate`
- `eth_mining`
- `eth_protocolVersion`
- `eth_getWork`
- `eth_submitWork`
- `eth_submitHashrate`

`GetWork`, `SubmitWork`, and `SubmitHashrate` return an explicit error on bor (non-PoW chain).
`Hashrate` returns `0x0` and `Mining` returns `false` rather than erroring (like in Erigon), which is more informative for our chain. 

`debug_accountAt` is also implemented via DebugAPI (and matches Erigon implementation)

Unit tests are added for all new methods. 
All methods have been manually tested against a mainnet pos-dev-node, with responses verified for parity against a live mainnet Erigon pos-dev-node.

Additionally, some tests are moved to different files, next to the actual implementations.

# Changes

# Checklist

- [x] Includes RPC methods changes

## Testing

- [x] I have added unit tests
- [x] I have tested this code manually on amoy and mainnet

### Manual tests

Tested on mainnet too.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -36,7 +36,7 @@ jobs:
           plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
           plugins: 'code-review@claude-code-plugins'
           prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
-          claude_args: '--model claude-opus-4-5-20251101'
+          claude_args: '--comment --model claude-opus-4-5-20251101'
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
 
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -51,7 +51,7 @@ jobs:
           additional_permissions: |
             actions: read
 
-          claude_args: '--model claude-opus-4-5-20251101'
+          claude_args: '--comment --model claude-opus-4-5-20251101'
 
           # Optional: Give a custom prompt to Claude. If this is not specified, Claude will perform the instructions specified in the comment that tagged it.
           # prompt: 'Update the pull request description to include a summary of changes.'
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # ─── BUILDER STAGE ───────────────────────────────────────────────────────────────
-FROM mirror.gcr.io/library/golang:1.26-alpine AS builder
+FROM golang:1.26-alpine AS builder
 
 ARG BOR_DIR=/var/lib/bor/
 ENV BOR_DIR=$BOR_DIR
@@ -22,7 +22,7 @@ RUN --mount=type=ssh \
     make bor
 
 # ─── RUNTIME STAGE ────────────────────────────────────────────────────────────────
-FROM mirror.gcr.io/library/alpine:3.21
+FROM alpine:3.23
 
 ARG BOR_DIR=/var/lib/bor/
 ENV BOR_DIR=$BOR_DIR
--- a/Dockerfile.alltools
+++ b/Dockerfile.alltools
@@ -7,12 +7,13 @@ ADD . /bor
 RUN cd /bor && make bor-all
 
 # Pull all binaries into a second stage deploy alpine container
-FROM alpine:latest
+FROM alpine:3.23
 
 RUN set -x \
     && apk add --update --no-cache \
     ca-certificates \
     && rm -rf /var/cache/apk/*
+
 COPY --from=builder /bor/build/bin/* /usr/bin/
 
 EXPOSE 8545 8546 30303 30303/udp
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,4 +1,4 @@
-FROM mirror.gcr.io/library/alpine:3.21
+FROM alpine:3.23
 
 ARG BOR_DIR=/var/lib/bor/
 ENV BOR_DIR=$BOR_DIR
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -125,6 +125,10 @@ var (
 	errNonContiguousHeaderRange = errors.New("non-contiguous headers in checkpoint range")
 )
 
+// maxAllowedFutureBlockTimeSeconds is the maximum number of seconds that a block
+// timestamp may exceed the local clock.
+const maxAllowedFutureBlockTimeSeconds = uint64(30)
+
 // SignerFn is a signer callback function to request a header to be signed by a
 // backing account.
 type SignerFn func(accounts.Account, string, []byte) ([]byte, error)
@@ -424,6 +428,12 @@ func (c *Bor) verifyHeader(chain consensus.ChainHeaderReader, header *types.Head
 		}
 		if parent == nil || now < parent.Time {
 			log.Error("Block announced too early post rio", "number", number, "headerTime", header.Time, "now", now)
+			return consensus.ErrFutureBlock
+		}
+		// Upper-bound check: a block whose timestamp is more than maxAllowedFutureBlockTimeSeconds
+		// ahead of the local clock is rejected.
+		if header.Time > now+maxAllowedFutureBlockTimeSeconds {
+			log.Error("Block timestamp too far in future post rio", "number", number, "headerTime", header.Time, "now", now)
 			return consensus.ErrFutureBlock
 		}
 	} else if c.config.IsBhilai(header.Number) {
--- a/consensus/bor/verify_header_test.go
+++ b/consensus/bor/verify_header_test.go
@@ -215,6 +215,54 @@ func TestVerifyHeader(t *testing.T) {
 			},
 			expectedError: consensus.ErrFutureBlock,
 		},
+		// Rio timestamp upper-bound tests: verify that far-future timestamps are rejected.
+		{
+			// A header with a timestamp set 100 years in the future must be
+			// rejected by the upper-bound check introduced in Rio.
+			name: "far-future timestamp in Rio mode is rejected",
+			setupChain: makeSetupChain(signerAddr, func(opts *chainSetupOptions) {
+				opts.rioBlock = big.NewInt(0)
+			}),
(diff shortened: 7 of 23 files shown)


marcello33:
@claude review this PR focusing on performance and security. Also, make sure no regression is introduced.

marcello33:
@claude addressed your points. Please, rerun your review.
Only minor diff is about the "configurable" max range for logs. 
Current implementation is a hard-coded constant, which is in line with what we do for `GetLatestLogMaxLogCount` and `GetLatestLogMaxBlockCount`

internal/ethapi/api.go
```diff
@@ -2658,6 +2721,93 @@ func (api *DebugAPI) PeerStats() interface{} {
...
+
+	for idx := uint64(0); idx <= lastTxIdx && idx < uint64(len(block.Transactions())); idx++ {
+		tx := block.Transactions()[idx]
+		msg, err := core.TransactionToMessage(tx, signer, block.BaseFee())
+		if err != nil {
+			return nil, fmt.Errorf("transaction %#x failed to convert to message: %v", tx.Hash(), err)
+		}
+
+		stateDb.SetTxContext(tx.Hash(), int(idx))
+		if _, err := core.ApplyMessage(evm, msg, new(core.GasPool).AddGas(tx.Gas())); err != nil {
```
manav2401, line 2797: For state-sync transactions (identifiable by tx type), might be worth skipping this as it will fail.

internal/ethapi/bor_api.go
```diff
@@ -2,14 +2,51 @@ package ethapi
...
+	// GetLatestLogMaxLogCount is the maximum number of logs that can be retrieved
+	GetLatestLogMaxLogCount = 30000
+	// GetLatestLogMaxBlockCount is the maximum number of blocks that can be scanned
+	GetLatestLogMaxBlockCount = 1000
+	// GetLatestLogMaxBlockScan caps blocks scanned in LogCount
+	GetLatestLogMaxBlockScan = 100000
+	// GetLogsMaxBlockRange is the maximum block range for bor_getLogs
+	GetLogsMaxBlockRange = 1000
+)
+
```
manav2401, line 49: Can we derive these values from some existing flag?

internal/ethapi/bor_api.go
```diff
@@ -155,3 +192,1055 @@ func (api *BorAPI) GetWitnessByBlockNumberOrHash(ctx context.Context, blockNrOrH
...
 	return RPCMarshalWitness(witness), nil
 }
+
+// GetBlockReceiptsByBlockHash returns all transaction receipts for a block by hash.
+//
+// Parameters:
+//   - blockHash: canonical block hash
+//
+// Returns an array of marshaled receipts, or error if the block is not found
+func (api *BorAPI) GetBlockReceiptsByBlockHash(ctx context.Context, blockHash common.Hash) ([]map[string]interface{}, error) {
```
manav2401, line 202: Can we define a common function for these methods so that we can re-use them for `bor` and `eth` namespace instead of duplicate logic?

internal/ethapi/bor_api.go
```diff
@@ -155,3 +192,1055 @@ func (api *BorAPI) GetWitnessByBlockNumberOrHash(ctx context.Context, blockNrOrH
...
+		header := api.b.CurrentHeader()
+		if header == nil {
+			return 0, errors.New(ErrHeaderNotFound)
+		}
+		blockNum = header.Number.Uint64()
+
+	case rpc.EarliestBlockNumber:
+		blockNum = 0
+
+	case rpc.SafeBlockNumber:
```
manav2401, line 358: Might be worth disabling this case altogether (or return finalized instead?).

manav2401: Okay maybe other functions also have it so we can keep it.

internal/ethapi/bor_api.go
```diff
@@ -155,3 +192,1055 @@ func (api *BorAPI) GetWitnessByBlockNumberOrHash(ctx context.Context, blockNrOrH
...
+// - "finalized" → finalized block
+// - numeric (>=0) → that block number
+// - unknown negative → latest executed (CurrentBlock)
+//
+// Parameters:
+//   - blockNrPtr: Optional block number or tag
+//     If nil, returns the latest executed block number
+//
+// Returns the block number as hexutil.Uint64
+func (api *BorAPI) BlockNumber(ctx context.Context, blockNrPtr *rpc.BlockNumber) (hexutil.Uint64, error) {
```
manav2401, line 332: In `eth` namespace, this function is defined as:
```
func (api *BlockChainAPI) BlockNumber() hexutil.Uint64
```
It always returns latest block number. Do we want to match both implementations? If adding a tag (e.g. latest, finalized) doesn't break the `eth` one, it might be good to update it but am okay with both.

internal/ethapi/bor_api.go
```diff
@@ -155,3 +192,1055 @@ func (api *BorAPI) GetWitnessByBlockNumberOrHash(ctx context.Context, blockNrOrH
...
+	modifiedAddresses := make(map[common.Address]bool)
+
+	// Add miner
+	modifiedAddresses[block.Coinbase()] = true
+
+	// Process all transactions to collect explicit addresses
+	signer := types.MakeSigner(api.b.ChainConfig(), block.Number(), block.Time())
+	for _, tx := range block.Transactions() {
+		// Add sender
+		if sender, err := types.Sender(signer, tx); err == nil {
```
manav2401, line 620: For the latest `prague` signer, this will fail for state-sync transactions. Maybe I'll create a separate task to handle it so that it doesn't return an error for state-sync transactions. 

But maybe we should just skip state sync transactions here.

internal/ethapi/bor_api.go
```diff
@@ -155,3 +192,1055 @@ func (api *BorAPI) GetWitnessByBlockNumberOrHash(ctx context.Context, blockNrOrH
...
+	// Collect modified addresses from pending transactions
+	modifiedAddresses := make(map[common.Address]bool)
+
+	// Add miner
+	modifiedAddresses[pendingBlock.Coinbase()] = true
+
+	// Process all pending transactions
+	signer := types.MakeSigner(api.b.ChainConfig(), pendingBlock.Number(), pendingBlock.Time())
+	for _, tx := range pendingBlock.Transactions() {
+		if sender, err := types.Sender(signer, tx); err == nil {
```
manav2401, line 710: Same as above.

internal/ethapi/api_test.go
```diff
@@ -4557,191 +4567,6 @@ func TestCreateAccessListWithStateOverrides(t *testing.T) {
...
-	}
-	return signRes.Raw, signRes.Tx
-}
-
-// makeSelfSignedRaw is a convenience for a 0-ETH self-transfer.
-func makeSelfSignedRaw(t *testing.T, api *TransactionAPI, addr common.Address) (hexutil.Bytes, *types.Transaction) {
-	return makeSignedRaw(t, api, addr, addr, big.NewInt(0))
-}
-
-func TestSendRawTransactionSync_Success(t *testing.T) {
```
manav2401, line 4677: I think these are not related to bor and should stay in this file.

internal/ethapi/api.go
marcello33: Thanks Manav for catching this, will fix.

internal/ethapi/bor_api.go
marcello33: Agreed there's an overlap here. However, the marshaling is already shared via `marshalReceipt`. In contrast, the wrapping logic is different because `GetBlockReceipts` accepts `blockNrOrHash` and handles `pending` blocks, whilst `GetBlockReceiptsByBlockHash` validates canonicality. So I believe that factoring it out would add some complexity. Keep in mind that the implementation here was done with the goal of having parity with erigon.

internal/ethapi/bor_api.go
marcello33: This is intentionally different from the eth namespace version to keep parity w/ erigon. There, this method acts as a sort of tag-resolver (latest, finalized, etc...). I think updating the eth version would change its behavior for existing callers.

internal/ethapi/bor_api.go
marcello33: Good call, thanks. Will add a StateSyncTxType skip

internal/ethapi/bor_api.go
marcello33: Right, same answer, thanks. Will add a StateSyncTxType skip

internal/ethapi/api_test.go
marcello33: You're right, will move back

internal/ethapi/bor_api.go
marcello33: There's a flag (`--rpc.logquerylimit (default 1000)`), but that limits the number of addresses/topics in `eth_getLogs`, not block ranges or result counts. 
We could make them configurable (like `rpc.logquerylimit`) but maybe a separate change?

PR #2068, opened 2026-02-19
core/stateless: fix witness RLP backward compat with WIT/0 peers

cffls:
Nodes upgraded from WIT/0 to WIT/1 were incorrectly jailing WIT/0 peers with "rlp: too few elements for stateless.ExtWitness". A 5-field ExtWitness encoding (Context, Headers, Codes, State, Keys) was accidentally introduced during an upstream geth merge, replacing the original 3-field wire format (Context, Headers, State) that WIT/0 peers still use.

Introduce BorWitness as the canonical 3-field wire format. EncodeRLP now encodes using BorWitness (State only, codes excluded). DecodeRLP tries BorWitness first and falls back to ExtWitness for backward compatibility with peers that temporarily ran the extended encoding.

ExtWitness is retained as the representation for the JSON debug API (ExecutionWitness, ExecutionWitnessByHash) and as the fallback decode target for the transitional period.

### DIFF
--- a/core/stateless/encoding_test.go
+++ b/core/stateless/encoding_test.go
@@ -1,0 +1,182 @@
+package stateless
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+// TestDecodeRLP_BorWitnessFormat verifies that a witness RLP-encoded in the
+// canonical 3-field BorWitness format is correctly decoded.
+func TestDecodeRLP_BorWitnessFormat(t *testing.T) {
+	contextHeader := &types.Header{Number: big.NewInt(100)}
+	parentHeader := &types.Header{Number: big.NewInt(99)}
+	node1 := []byte("statenode1")
+	node2 := []byte("statenode2")
+
+	bw := &BorWitness{
+		Context: contextHeader,
+		Headers: []*types.Header{parentHeader},
+		State:   [][]byte{node1, node2},
+	}
+
+	encoded, err := rlp.EncodeToBytes(bw)
+	if err != nil {
+		t.Fatalf("failed to encode BorWitness: %v", err)
+	}
+
+	var w Witness
+	if err := rlp.DecodeBytes(encoded, &w); err != nil {
+		t.Fatalf("DecodeRLP failed for BorWitness input: %v", err)
+	}
+
+	if w.context.Number.Cmp(contextHeader.Number) != 0 {
+		t.Errorf("context number: got %v, want %v", w.context.Number, contextHeader.Number)
+	}
+	if len(w.Headers) != 1 || w.Headers[0].Number.Cmp(parentHeader.Number) != 0 {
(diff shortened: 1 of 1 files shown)


core/stateless/encoding.go
```diff
@@ -24,7 +24,28 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
-// ToExtWitness converts our internal witness representation to the consensus one.
+// BorWitness is the canonical 3-field RLP encoding used for network
+// transmission in Bor. The State field carries all proof data — both
+// contract bytecodes and MPT state trie nodes — as a flat list of byte slices.
+type BorWitness struct {
+	Context *types.Header
```
kamuikatsurgi, line 31: Should we add `json:` here too, if we want to switch to this type for APIs?

core/stateless/encoding.go
```diff
@@ -24,7 +24,28 @@ import (
...
+	Headers []*types.Header
+	State   [][]byte
```
kamuikatsurgi, line 33: Should we make this `[]hexutil.Bytes`? To match geth's version.

kamuikatsurgi (approved):
a few nits

core/stateless/encoding.go
cffls: I would be cautious about doing so, because the encoding of this will need to be compatible with old version. It seems safer to keep it as it is.

core/stateless/encoding.go
manav2401: If you're talking about `Context` specifically, that's an internal field. But agree, we should have json fields for Headers and State as done in `ExtWitness`.

kamuikatsurgi: Not talking about `Context` specifically.

core/stateless/encoding.go
kamuikatsurgi: Just want to make sure we are handling `[][]byte` <-> `[]hexutil.Bytes` conversions safely in case then.

cffls: Yes, it will be fine. Since the definition of `hexutil.Bytes` is `[]byte`, so they are identical.

kamuikatsurgi: So in that case, should we go ahead and make it `[]hexutil.Bytes`? Since they are identical? This way, we can have as much similarity in two witness types as possible to avoid future mismatches.

kamuikatsurgi: Just a thought, otherwise we can keep it as is for now.

core/stateless/encoding.go
cffls: The `json:` is used in debug api [here](https://github.com/0xPolygon/bor/blob/00920dd6a220bda7ddd7b12a5b97951e12cf19db/eth/api_debug.go#L496). Since it is using ExtWitness, BorWitness becomes irrelevant.

core/stateless/encoding.go
cffls: I'd like to keep this actually. It might make it easier for git to diff the changes without touching BorWitness. For future changes, we want everything goes to ExtWitness and not BorWitness.

kamuikatsurgi: Cool.

PR #2077, opened 2026-02-23
(feat): re-enable snap sync

manav2401:
# Description

This PR re-enables snap sync in bor and does the following changes. 
- Remove changes which would force bor to use full sync
- Allow `--syncmode=snap` flag to be set
- Adds a filter in downloader which would avoid syncing receipts from eth/68 peers as bor receipts are only served in eth/69. 
- Adds a test to validate the same. 
- Uncomment all snap sync related tests which were skipped earlier

A small refactor is being done to add a `kind()` method in all downloader queues to make comparison easy instead of using `reflect.TypeOf` in every iteration. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/sync.go
+++ b/eth/sync.go
@@ -198,7 +198,7 @@ func peerToSyncOp(mode downloader.SyncMode, p *eth.Peer) *chainSyncOp {
 
 func (cs *chainSyncer) modeAndLocalHead() (downloader.SyncMode, *big.Int) {
 	// If we're in snap sync mode, return that directly
-	if cs.handler.snapSync.Load() && !cs.handler.statelessSync.Load() {
+	if cs.handler.snapSync.Load() {
 		block := cs.handler.chain.CurrentSnapBlock()
 		td := cs.handler.chain.GetTd(block.Hash(), block.Number.Uint64())
 		return downloader.SnapSync, td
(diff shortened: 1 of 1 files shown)


eth/downloader/bor_downloader_test.go
```diff
@@ -646,12 +662,11 @@ func testForkedSync(t *testing.T, protocol uint, mode SyncMode) {
 // currently and is not dropped.
 func TestHeavyForkedSync68Full(t *testing.T) { testHeavyForkedSync(t, eth.ETH68, FullSync) }
 
-// TODO v1.16.1: Enable when snap sync gets fixed
+// Skipping as eth/68 nodes are filtered out during snap sync
```
marcello33: Shall we t.Skip such tests instead of commenting them out? Same elsewhere

cffls:
Looks good! Would be great if you can add a node with snap sync mode enabled to e2e and verify basic things in the test.

eth/downloader/bor_downloader_test.go
manav2401: Makes sense, doing it.

PR #2085, opened 2026-02-24
Dynamic Target Gas for stable prices

lucca30:
# Description

This PR introduces **dynamic EIP-1559 target gas percentage** adjustment (post-Lisovo) to reduce artificial scarcity and mitigate **priority fee spikes** during short demand bursts, while keeping base fee behavior predictable.

## Why
Priority fee is pure market bidding and can spike aggressively when blocks are full. Our current “dynamic gas limit” approach reduces real capacity even when we could safely provide more, increasing the chance of 100% utilization and triggering sharp priority-fee bidding. Instead, we keep capacity and adjust **gas target** directly (since target is currently derived from gas limit) to reduce priority fee spike frequency/severity.

## New CLI flags (Sealer)
- `--miner.enableDynamicTargetGas`  
  Enables dynamic target gas adjustment based on the parent block base fee (**post-Lisovo**). Mutually exclusive with `enableDynamicGasLimit`.
- `--miner.targetGasMin`  
  Minimum target gas percentage (1–100) used when base fee is below the configured lower bound.
- `--miner.targetGasMax`  
  Maximum target gas percentage (1–100) used when base fee is above the configured upper bound.

## Behavior
Adds `BorConfig.GetDynamicTargetGasPercentage(parentBaseFee, number)`:
- If **pre-Lisovo**, feature disabled, **pre-London** (`parentBaseFee == nil`), or config incomplete (`TargetBaseFee == nil`) → **fallback** to existing `GetTargetGasPercentage(number)`.
- Computes bounds around `TargetBaseFee` using optional `BaseFeeBuffer`:
  - `upper = desired + buffer`
  - `lower = max(desired - buffer, 0)` (underflow-safe)
- If `parentBaseFee > upper` → return `TargetGasMax` (valid 1–100), otherwise fallback to static.
- If `parentBaseFee < lower` → return `TargetGasMin` (valid 1–100), otherwise fallback to static.
- If within buffer → return static `GetTargetGasPercentage(number)` (no change).
# Changes

# Breaking changes

No breaking changes, is compatible with post-Lisovo blocks.

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -878,9 +878,9 @@ func DefaultConfig() *Config {
 			GasLimitMax:              miner.DefaultConfig.GasLimitMax,
 			TargetBaseFee:            miner.DefaultConfig.TargetBaseFee,
 			BaseFeeBuffer:            miner.DefaultConfig.BaseFeeBuffer,
-			EnableDynamicTargetGas:   miner.DefaultConfig.EnableDynamicTargetGas,
-			TargetGasMinPercentage:   miner.DefaultConfig.TargetGasMinPercentage,
-			TargetGasMaxPercentage:   miner.DefaultConfig.TargetGasMaxPercentage,
+			EnableDynamicTargetGas:   false,
+			TargetGasMinPercentage:   50,                                         // 50% floor
+			TargetGasMaxPercentage:   80,                                         // 80% ceiling
 			GasPrice:                 big.NewInt(params.BorDefaultMinerGasPrice), // bor's default
 			ExtraData:                "",
 			Recommit:                 125 * time.Second,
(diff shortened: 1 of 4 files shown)


params/config.go
```diff
@@ -1040,6 +1049,67 @@ func (c *BorConfig) GetTargetGasPercentage(number *big.Int) uint64 {
...
+		return c.GetTargetGasPercentage(number)
+	}
+
+	// Config guard: if essential field is absent, fall back
+	if c.TargetBaseFee == nil {
+		log.Error("EnableDynamicTargetGas is set but TargetBaseFee is nil, falling back to static")
+		return c.GetTargetGasPercentage(number)
+	}
+
+	baseFee := parentBaseFee.Uint64()
```
marcello33: I believe we might have a silent truncation here. For values exceeding `math.MaxUint64` (unlikely, almost impossible, but good to fix for security), the truncated value would produce incorrect comparisons against the upper or lower bounds, potentially returning the wrong target gas percentage.

params/config.go
```diff
@@ -1040,6 +1049,67 @@ func (c *BorConfig) GetTargetGasPercentage(number *big.Int) uint64 {
...
+	desired := *c.TargetBaseFee
+	var buffer uint64
+	if c.BaseFeeBuffer != nil {
+		buffer = *c.BaseFeeBuffer
+	}
+
+	upperBound := desired + buffer
```
marcello33: We could have an overflow here, if `> math.MaxUint64`, right?
Maybe worth adding a check or use `big.Int`

internal/cli/server/config.go
```diff
@@ -1276,6 +1290,25 @@ func (c *Config) buildEth(stack *node.Node, accountManager *accounts.Manager) (*
...
+				return nil, fmt.Errorf("miner.targetGasMin (%d) must be less than miner.targetGasMax (%d)", c.Sealer.TargetGasMin, c.Sealer.TargetGasMax)
+			}
+			if c.Sealer.TargetBaseFee == 0 {
+				return nil, fmt.Errorf("miner.targetBaseFee must be greater than 0 when dynamic target gas is enabled")
+			}
+			if c.Sealer.BaseFeeBuffer >= c.Sealer.TargetBaseFee {
+				log.Warn("miner.baseFeeBuffer >= miner.targetBaseFee; lower bound will be 0 (all base fees treated as below target)")
+			}
+		}
+
```
marcello33, line 1323: Correct me if I am wrong here, but apparently the config validation only checks `TargetGasMin < TargetGasMax`, right?
Should we validate that it is between min and max?

consensus/misc/eip1559/eip1559_test.go
```diff
@@ -1548,3 +1548,138 @@ func TestLowBaseFeeIntegrationWithExistingBoundary(t *testing.T) {
...
+	min := tGasMin
+	max := tGasMax
+	dbf := desiredFee
+	buf := buffer
+
+	borCfg := &params.BorConfig{
+		DandeliBlock:           big.NewInt(dandeliBlock),
+		LisovoBlock:            big.NewInt(lisovoBlock),
+		EnableDynamicTargetGas: &enabled,
+		TargetGasMin:           &min,
```
marcello33: `TargetGasMin` and `TargetGasMax` are percentages, right? I'd name them smth like `TargetGasMinPercentage` just to indicate that

consensus/misc/eip1559/eip1559.go
```diff
@@ -183,10 +183,11 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header) *big.Int {
 // calcParentGasTarget calculates the target gas based on parent block gas limit. Earlier
```
marcello33, line 183: Is the PR stale against develop? Maybe worth merging it back

kamuikatsurgi:
This adds new flags, once you're done with the PR, can you merge develop and run make docs?

params/config.go
lucca30: Good catch! Fixed by switching to big.Int arithmetic throughout the function: `parentBaseFee` is now compared directly as a *big.Int without converting to uint64, so there's no truncation risk for values exceeding math.MaxUint64.

params/config.go
lucca30: Fixed together with the truncation issue above. `upperBound` and `lowerBound` are now computed using `big.Int` (`new(big.Int).Add(bigDesired, bigBuffer)`), so no overflow is possible regardless of the configured values.

consensus/misc/eip1559/eip1559_test.go
lucca30: Good suggestion! Renamed adding suffix `Percentage`

consensus/misc/eip1559/eip1559.go
lucca30: Merged develop into the branch. All good now.

internal/cli/server/config.go
lucca30: You're right. It was only checking `min < max`. Added a validation: when `EnableDynamicTargetGas` is set and `TargetGasPercentage` (the static fallback) is explicitly configured, it must fall strictly within (`TargetGasMinPercentage`, `TargetGasMaxPercentage`). Added two test cases covering the in-range and out-of-range scenarios.

marcello33 (approved):

PR #2087, opened 2026-02-25
fix priced heap corruption in replacesPending path

lucca30:
# Description

## Issue

Three [CI failures ](https://github.com/0xPolygon/bor/actions/runs/22381143650/job/64781936521?pr=2086#step:7:110) caused by the same root-cause corruption pattern:

```
pool internal state corrupted: total priced transaction count 275 != 200  # TestTransactionFutureAttack
pool internal state corrupted: total priced transaction count 2 != 1      # TestSetCodeTransactions
pool internal state corrupted: total priced transaction count 7 != 4      # TestStatusCheck
```

`pool.all` and the `priced` heap diverge because items are double-added to the heap.

## Root cause

Three places in `legacypool.go` launched goroutines to insert transactions into
the `priced` heap while `pool.all` already held the transaction:

| Location | Goroutine |
|---|---|
| `add()` — `replacesPending` path | `go pool.priced.PutMany(drop)` |
| `add()` — pending replacement path | `go pool.priced.Put(tx)` |
| `enqueueTx()` — queue path | `go pool.priced.Put(tx)` |

All three share the same race with `Reheap()`:

```
Goroutine G (pending):   waiting to acquire reheapMu
Reheap():                acquires reheapMu → rebuilds heap from pool.all
                         → correctly adds the tx (still in pool.all)
                         → releases reheapMu
Goroutine G runs:        acquires reheapMu → Put/PutMany double-adds
                         the same tx → priced.Len() > all.Count()
```

This is a logical race — not detected by `-race` because both operations are
correctly synchronized via `reheapMu`; the problem is ordering, not concurrent
unsynchronized access. CI machines with higher CPU counts and different goroutine
scheduling make the bad ordering likely.

## Reproducing locally

`runtime.Gosched()` was added in `validatePoolInternals` (test helper) between
`Reheap()` and the count read to widen the race window and make failures
deterministic:

```bash
# All three of these should fail reliably without the fix, pass with it:
go test -count=50 -timeout 180s -run TestTransactionFutureAttack ./core/txpool/legacypool/
go test -count=10 -timeout 60s  -run TestSetCodeTransactions     ./core/txpool/legacypool/
go test -count=10 -timeout 60s  -run TestStatusCheck             ./core/txpool/legacypool/
```

## Fix

Remove all three goroutines. Call `Put`/`PutMany` synchronously while `pool.mu` is held.

```go
// replacesPending path — before:
if replacesPending {
    if async {
        go pool.priced.PutMany(drop)
    } else {
        for _, dropTx := range drop { pool.priced.Put(dropTx) }
    }
    return false, ErrFutureReplacePending
}
// after:
if replacesPending {
    pool.priced.PutMany(drop)
    return false, ErrFutureReplacePending
}

// pending replacement path — before:
pool.all.Add(tx)
if async {
    go pool.priced.Put(tx)
} else {
    pool.priced.Put(tx)
}
// after:
pool.all.Add(tx)
pool.priced.Put(tx)

// enqueueTx — before:
if addAll {
    pool.all.Add(tx)
    go pool.priced.Put(tx)
}
// after:
if addAll {
    pool.all.Add(tx)
    pool.priced.Put(tx)
}
```

## Why the fix is safe (no deadlock)

The original goroutines were introduced to avoid holding `pool.mu` while waiting
for `reheapMu`. This concern is valid in principle, but no deadlock can occur
because the lock hierarchy is strictly one-directional:

```
pool.mu (RWMutex) ──► reheapMu (Mutex) ──► lookup.lock (RWMutex)
```

Every `pricedList` method (`Put`, `PutMany`, `Removed`, `Discard`, `Reheap`,
`SetBaseFee`) acquires `reheapMu` and optionally `lookup.lock` — **none of them
ever acquire `pool.mu`**. A `reheapMu → pool.mu` back-edge does not exist, so
holding `pool.mu` while calling `Put`/`PutMany` cannot deadlock.

This pattern is already established in the codebase: `LegacyPool.Clear()`
holds `pool.mu` and calls `pool.priced.Reheap()` — a full heap rebuild.

## Proof via lock-ordering stress tests

Three tests were added to `legacypool2_test.go` that create the exact scenarios
that would deadlock if the lock ordering were wrong:

| Test | What it stresses |
|---|---|
| `TestLockOrderingPricedHeapNoDeadlock` | `pool.mu(W) + PutMany` ∥ `Reheap()` |
| `TestLockOrderingReplacePendingNoDeadlock` | End-to-end: `addRemotesSync(futureTxs)` ∥ `Reheap()` |
| `TestLockOrderingRemovedNoDeadlock` | `pool.mu(W) + Removed` ∥ `Reheap()` |

Each test calls `t.Fatal("deadlock detected...")` via `time.After(10s)` if
either goroutine stops making progress.

```bash
# Run lock-ordering proofs
go test -v -race -count=1 -timeout 60s \
  -run "TestLockOrdering" \
  ./core/txpool/legacypool/

# Run all three originally failing tests (10 runs each, should all pass)
go test -count=10 -timeout 120s \
  -run "TestTransactionFutureAttack|TestSetCodeTransactions|TestStatusCheck" \
  ./core/txpool/legacypool/
```

## Files changed

| File | Change |
|---|---|
| `core/txpool/legacypool/legacypool.go` | Remove all three async goroutines in `replacesPending`, pending replacement, and `enqueueTx` paths; call `Put`/`PutMany` synchronously |
| `core/txpool/legacypool/legacypool_test.go` | Add `runtime.Gosched()` in `validatePoolInternals` to reliably expose the race window in tests |
| `core/txpool/legacypool/legacypool2_test.go` | Add three lock-ordering stress tests |

# Changes

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1136,13 +1136,7 @@ func (pool *LegacyPool) add(tx *types.Transaction, async bool) (replaced bool, e
 			delete(pool.lastRebroadcast, old.Hash())
 		}
 		pool.all.Add(tx)
-		if async {
-			// We don't want to get blocked on this due to internal lock, so
-			// call the function to insert transactions into the heap async.
-			go pool.priced.Put(tx)
-		} else {
-			pool.priced.Put(tx)
-		}
+		pool.priced.Put(tx)
 		pool.queueTxEvent(tx)
 		log.Trace("Pooled new executable transaction", "hash", hash, "from", from, "to", tx.To())
 
@@ -1205,9 +1199,7 @@ func (pool *LegacyPool) enqueueTx(hash common.Hash, tx *types.Transaction, addAl
 	}
 	if addAll {
 		pool.all.Add(tx)
-		// We don't want to get blocked on this due to internal lock, so
-		// call the function to insert transactions into the heap async.
-		go pool.priced.Put(tx)
+		pool.priced.Put(tx)
 	}
 	return replaced != nil, nil
 }


lucca30:
@claude Can you do a deep review on the proposed fix in order to evaluate the risks involved by replacing a goroutine by a synchronous code. Check if the lock hierarchy actually protects us from deadlocks and if the introduced tests ensures us no deadlock was introduced.

manav2401:
Nice find on the logical race condition. I've been thinking about the change and I think there's a better way to prevent the race without losing some performance benefits. 

1. I think the fix you've applied doesn't eliminate the logical race condition fully. While it makes the call directly instead of doing it in a go routine, the call itself is still blocked by `reheapMu` which will be acquired while reheaping. So with your fix the original issue can still happen as follows - if a transaction is added into `pool.all` and then reheap starts, it will use this transaction to build the heap but the call to `Put` transaction into heap (irrespective of it being called with go routine or not) is still blocked on the reheap lock and will double add the transaction into heap. 
2. The calls were made inside a go routine to gain some performance benefits by unblocking the global pool lock when some internal re-arrangements in the heap are going on. Removing the go routine will defeat the performance gains we could get. 
3. As the logical race is still valid issue, I've worked on a slightly different implementation in https://github.com/0xPolygon/bor/pull/2095. Happy to discuss more on this. Thanks!

PR #2092, opened 2026-02-27
core/vm, core, miner: fix interrupt propagation to nested EVM calls

lucca30:
# Description

## Problem

The block-building interrupt flag (`interruptBlockBuilding`) is not propagated to nested EVM calls. When the interpreter executes opcodes that make nested calls — `CALL`, `DELEGATECALL`, `STATICCALL`, `CALLCODE`, `CREATE` — they all pass `nil` for the interrupt parameter to `evm.Run()`. This causes `Run()` to create a dummy `new(atomic.Bool)` that never fires, making nested contract execution **completely uninterruptible**.

Since virtually all real transactions involve at least one nested call (proxy patterns, DEX routers, multi-hop swaps), the per-opcode interrupt check is effectively bypassed for the bulk of EVM execution time.

### Real-world impact

Block [83,527,864](https://polygonscan.com/block/83,527,864) on Polygon PoS mainnet took **5.654s** to build instead of ~2s. The interrupt timer fired on schedule (~1.5s after block building started), but a transaction executing a complex nested call (proxy/DEX pattern) ran for an additional **~4.1 seconds** because the interrupt couldn't reach the nested execution. The block was announced **3.7 seconds late**.

### Root cause

The interrupt flag flows through this chain for the top-level call:

```
worker.commitTransaction → core.ApplyTransaction → ApplyTransactionWithEVM
  → ApplyMessageNoFeeBurnOrTip → st.execute(interrupt)
    → evm.Call(..., interrupt) → evm.Run(contract, input, false, interrupt) ✓ PASSED
```

But when opcodes make nested calls, they all pass `nil`:

```go
instructions.go  opCall:         evm.Call(..., nil)       // ← nil!
evm.go           CallCode:       evm.Run(..., nil)        // ← nil!
evm.go           DelegateCall:   evm.Run(..., nil)        // ← nil!
evm.go           StaticCall:     evm.Run(..., nil)        // ← nil!
evm.go           Create:         evm.Run(..., nil)        // ← nil!
```

When `interrupt` is nil, `interpreter.go` creates a dummy `new(atomic.Bool)` that is always false — the nested execution never checks the real interrupt flag.

## Test: proving the bug exists

`TestInterruptNestedCall` in `core/vm/runtime/runtime_test.go` demonstrates the issue with two sub-tests:

- **TopLevel**: An infinite loop at the top level, interrupted after 50ms. Proves the interrupt mechanism works for flat code.
- **NestedCall**: The same infinite loop inside a `STATICCALL`. On pre-fix code, the interrupt is ignored inside the nested call.

### Pre-fix results (commit `029c58680`)

To reproduce the bug, checkout the test commit before the fix and run:

```bash
git checkout 029c58680
go test -run TestInterruptNestedCall -v ./core/vm/runtime/
```

```
=== RUN   TestInterruptNestedCall/TopLevel
--- PASS: (0.05s)
=== RUN   TestInterruptNestedCall/NestedCall
    runtime_test.go:204: Bug confirmed: nested call took 1.12s (interrupt ignored inside STATICCALL)
--- PASS: (1.12s)
```

The nested call took **1.12 seconds** — it ran until gas exhaustion, completely ignoring the interrupt set at 50ms.

### Post-fix results (commit `c01a363ec`)

```bash
git checkout c01a363ec
go test -run TestInterruptNestedCall -v ./core/vm/runtime/
```

```
=== RUN   TestInterruptNestedCall/TopLevel
--- PASS: (0.05s)
=== RUN   TestInterruptNestedCall/NestedCall
    runtime_test.go:201: Fix verified: nested call interrupted in 51ms
--- PASS: (0.05s)
```

The nested call now stops in **~51ms** — a **22x improvement**, matching the top-level interrupt latency.

## Fix

Move `interrupt` from a function parameter to a field on the `EVM` struct. This ensures all call depths — top-level, nested, and deeply nested — share the same interrupt flag.

```go
// New field on EVM struct
type EVM struct {
    // ...
    interrupt *atomic.Bool
}

// Set once at block-building start
func (evm *EVM) SetInterrupt(interrupt *atomic.Bool) {
    evm.interrupt = interrupt
}

// Run() reads from the struct — all call depths see the same flag
func (evm *EVM) Run(contract *Contract, input []byte, readOnly bool) (ret []byte, err error) {
    interrupt := evm.interrupt
    if interrupt == nil {
        interrupt = new(atomic.Bool)
    }
    for {
        if interrupt.Load() { return nil, ErrInterrupt }
        // ... execute opcode ...
    }
}
```

The `interrupt` parameter is removed from `Call()`, `Run()`, and the entire `ApplyTransaction` chain (29 files, removing parameter threading through `ApplyMessage`, `ApplyMessageNoFeeBurnOrTip`, `execute()`, etc.). The miner sets the interrupt once via `evm.SetInterrupt(&w.interruptBlockBuilding)` in `makeEnv`.

## Performance validation

An [earlier optimization](https://github.com/0xPolygon/bor/pull/1570) replaced `context.Done()` with `atomic.Bool` for the per-opcode interrupt check, yielding a **25x throughput improvement** (10.34 ns/op → 0.41 ns/op). We validated that our refactor does not regress this optimization.

### Why there should be zero impact

Both pre-fix and post-fix produce **identical hot-path code** in the interpreter loop:

```go
// Both versions: local variable used in the hot loop
interrupt := /* parameter (pre-fix) | evm.interrupt (post-fix) */
for {
    if interrupt.Load() { ... }  // ← same local *atomic.Bool in both cases
}
```

The only difference is one extra struct field read at function entry (`evm.interrupt`), which happens **once per `Run()` call** — negligible compared to millions of opcode iterations per transaction.

### BenchmarkSimpleLoop comparison (100M gas, 6 runs, 5s benchtime)

```
                                       │   pre-fix   │              post-fix              │
                                       │   sec/op    │   sec/op     vs base               │
SimpleLoop/loop-100M                      133.6m ± 0%   134.5m ± 0%  +0.63% (p=0.002 n=6)
SimpleLoop/call-reverting-100M            465.3m ± 0%   466.5m ± 0%  +0.26% (p=0.026 n=6)
SimpleLoop/call-nonexist-100M             348.4m ± 0%   350.7m ± 0%  +0.67% (p=0.002 n=6)
SimpleLoop/call-EOA-100M                  312.9m ± 0%   315.5m ± 0%  +0.82% (p=0.002 n=6)
SimpleLoop/call-identity-100M             315.8m ± 1%   321.5m ± 0%  +1.80% (p=0.002 n=6)
SimpleLoop/staticcall-identity-100M       127.3m ± 1%   130.6m ± 1%  +2.58% (p=0.002 n=6)
SimpleLoop/loop2-100M                     155.2m ± 0%   166.8m ± 0%  +7.51% (p=0.002 n=6)

Memory allocations: unchanged across all benchmarks.
```

**Interpretation**: The critical `loop-100M` benchmark (pure opcode loop, most sensitive to per-opcode overhead) shows **+0.63%** — well within system noise. Most benchmarks fall in the 0.2–1.8% range. The `loop2-100M` outlier at +7.5% is attributed to thermal throttling: benchmarks were run sequentially (~4 minutes each), and the consistent slight increase across **all** benchmarks (never a decrease) is a classic sign of CPU thermal effects from back-to-back benchmark suites on Apple Silicon. The per-opcode interrupt check code is literally identical in both versions — the local `*atomic.Bool` variable and `.Load()` call produce the same machine code.

# Changes

# Breaking changes

No breaking changes

### DIFF
--- a/eth/tracers/js/tracer_test.go
+++ b/eth/tracers/js/tracer_test.go
@@ -67,7 +67,7 @@ func runTrace(tracer *tracers.Tracer, vmctx *vmContext, chaincfg *params.ChainCo
 
 	tracer.OnTxStart(evm.GetVMContext(), types.NewTx(&types.LegacyTx{Gas: gasLimit, GasPrice: vmctx.txCtx.GasPrice}), contract.Caller())
 	tracer.OnEnter(0, byte(vm.CALL), contract.Caller(), contract.Address(), []byte{}, startGas, value.ToBig())
-	ret, err := evm.Run(contract, []byte{}, false, nil)
+	ret, err := evm.Run(contract, []byte{}, false)
 	tracer.OnExit(0, ret, startGas-contract.Gas, err, true)
 	// Rest gas assumes no refund
 	tracer.OnTxEnd(&types.Receipt{GasUsed: gasLimit - contract.Gas}, nil)
--- a/eth/tracers/logger/logger_test.go
+++ b/eth/tracers/logger/logger_test.go
@@ -55,7 +55,7 @@ func TestStoreCapture(t *testing.T) {
 
 	var index common.Hash
 	logger.OnTxStart(evm.GetVMContext(), nil, common.Address{})
-	_, err := evm.Run(contract, []byte{}, false, nil)
+	_, err := evm.Run(contract, []byte{}, false)
 	if err != nil {
 		t.Fatal(err)
 	}


lucca30:
@claude Do a detailed and careful review of this code changes. Analyze if any issue, bug, security breach was introduced. Analyze if the refactor kept the behavior as it should and if just addressed the missed opcode calls. Check if there are any potential issue with performance introduced.

PR #2094, opened 2026-03-02
blockstm: replace O(n²) UpdateDeps with inverted-index DepsBuilder

kamuikatsurgi:
The block producer computes a transaction dependency DAG during sequential execution and embeds it in the block header. Importers use it to schedule parallel BlockSTM execution without speculative re-execution overhead.

The old algorithm (`UpdateDeps`) scanned every preceding transaction's full write-set for each new transaction: O(n²) per block. At 300 tx/block, this was the dominant cost in the miner, confirmed by CPU profiling.

## Algorithm

DepsBuilder replaces it with two data structures:

**Inverted index** — `lastWriter map[Key]int`: maps each state key to the index of the most recent transaction that wrote it. For a new tx at index i, a single pass over its read-set resolves all direct dependencies in O(R), where R is the read-set size. Only the latest writer matters; earlier writers are already covered by transitivity.

**Word-parallel bitsets** — `directDeps []TxBitset` and `reachable []TxBitset`: each bitset is []uint64 of width (N+63)/64. Transitive reduction runs in O(N/64) per tx using two bitset operations:
```go
reachable[i] = ∪ reachable[j] for j in directDeps[i]     // transitive closure
directDeps[i] &^= reachable[i]    // remove redundant edges
```

Total complexity: O(n·R) vs O(n²·W) previously, where R = avg read-set size and W = avg write-set size. Default bitset width is 512; ensureCapacity doubles on overflow and widens all existing bitsets atomically before appending new ones — maintaining the invariant that all bitsets have identical len(words) at all times.

The output format (map[int]map[int]bool) is unchanged, full backwards compatibility with old importers.

## Correctness fixes

ValidatorBytes preservation — If DepsBuilder errored, the original code skipped rlp.DecodeBytes and re-encoded a zero-value BlockExtraData, silently dropping ValidatorBytes. These are consensus-critical on Cancun sprint transitions. Fix: always decode the existing extra data before overwriting TxDependency.

delayFlag off-by-one (pre-existing) — mvReadMapList[i-1] checked the wrong transaction's reads when deciding whether to include the DAG hint. The last transaction's coinbase/burnt-contract balance reads were never checked. If only the last tx read these balances, the hint was incorrectly embedded. Fixed to mvReadMapList[i].

## Backwards compatibility

Verified on a devnet with 1 new validator node and 3 older RPC nodes. After 1,352+
blocks:
- Zero consensus errors on all nodes
- All three checkpoints logged on old nodes: decode, dependency verification, parallel
  execution completion

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -106,6 +106,10 @@ var (
 	intermediateRootTimer = metrics.NewRegisteredTimer("worker/intermediateRoot", nil)
 	// commitTimer measures total time for complete block building (tx execution + finalization + state root)
 	commitTimer = metrics.NewRegisteredTimer("worker/commit", nil)
+	// writeBlockAndSetHeadTimer measures total time for WriteBlockAndSetHead in the seal result loop.
+	// This covers the entire gap between block sealing and event posting: witness encoding, batch write,
+	// state commit, and (in hashdb mode) trie GC. Spikes here directly delay block broadcasting.
+	writeBlockAndSetHeadTimer = metrics.NewRegisteredTimer("worker/writeBlockAndSetHead", nil)
 
 	// Cache hit/miss metrics for block production (miner path)
 	// These are the same meters used by the import path in blockchain.go
@@ -997,7 +1001,9 @@ func (w *worker) resultLoop() {
 			}
 
 			// Commit block and state to database.
+			writeStart := time.Now()
 			_, err = w.chain.WriteBlockAndSetHead(block, receipts, logs, task.state, true)
+			writeBlockAndSetHeadTimer.Update(time.Since(writeStart))
 
 			if err != nil {
 				log.Error("Failed writing block to chain", "err", err)
@@ -1012,7 +1018,7 @@ func (w *worker) resultLoop() {
 				"elapsed", common.PrettyDuration(time.Since(task.createdAt)))
 
 			// Broadcast the block and announce chain insertion event
-			w.mux.Post(core.NewMinedBlockEvent{Block: block, Witness: witness})
+			w.mux.Post(core.NewMinedBlockEvent{Block: block, Witness: witness, SealedAt: time.Now()})
 
 			sealedBlocksCounter.Inc(1)
 
@@ -1073,6 +1079,8 @@ func (w *worker) makeEnv(header *types.Header, coinbase common.Address, witness 
 		prefetchReader: genParams.prefetchReader,
 		processReader:  genParams.processReader,
 	}
+	env.evm.SetInterrupt(&w.interruptBlockBuilding)
+
 	// Keep track of transactions which return errors so they can be removed
 	env.tcount = 0
 	env.mvReadMapList = []map[blockstm.Key]blockstm.ReadDescriptor{}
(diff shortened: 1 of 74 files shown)


kamuikatsurgi:
@claude, review this PR focusing on performance and security. Also, make sure no regression or bug is introduced.

miner/worker.go
```diff
@@ -1382,11 +1395,11 @@ mainloop:
 			delayFlag := true
 
 			for i := 1; i <= len(env.mvReadMapList)-1; i++ {
-				reads := env.mvReadMapList[i-1]
+				reads := env.mvReadMapList[i]
```
cffls, line 1406: Was this due to a bug, or the items in mvReadMapList is shifted by one position?

miner/worker.go
```diff
@@ -1315,18 +1317,17 @@ mainloop:
...
 				}
 
-				temp := blockstm.TxDep{
-					Index:         env.tcount - 1,
-					ReadList:      env.state.MVReadList(),
-					FullWriteList: env.depsMVFullWriteList,
+				temp := blockstm.TxReadWriteSet{
+					Index:     env.tcount - 1,
+					ReadList:  env.state.MVReadList(),
+					WriteList: env.state.MVFullWriteList(),
```
cffls, line 1338: Why do we need a new data structure here? The existing structure seems sufficient.

miner/worker.go
kamuikatsurgi: Yes, a pre-existing bug:
```
2. delayFlag Off-by-One (worker.go:1398)
Before: reads := env.mvReadMapList[i-1] - checked the previous transaction's reads, not the current one.

After: reads := env.mvReadMapList[i] - correctly checks transaction i's reads.

This fix ensures the last transaction's coinbase/burnt-contract balance reads are properly detected.
```

miner/worker.go
kamuikatsurgi: TxDep.FullWriteList carries the entire write history of all preceding transactions because `UpdateDeps` needed to scan every previous writer for each new tx, that's where the O(n²) came from.

`DepsBuilder` doesn't need that. Its inverted index (lastWriter map[Key]int) means it only needs each transaction's own writes to update the map.

kamuikatsurgi: Inverted index has lastWriter map[Key]int, as each tx commits, it updates the map with its writes. For a new tx, just look up each key; it reads directly. 

After Tx0 writes A, Tx1 writes B, Tx2 writes A: lastWriter = { A:2, B:1 }
Tx3 reads [A, B]:
A → lastWriter[A] = 2  ← dep on Tx2
B → lastWriter[B] = 1  ← dep on Tx1

One map lookup per read key instead of scanning all previous transactions. O(R) instead of O(n·W).

miner/worker.go
cffls: Interesting, I wonder why this wasn't captured before. Is there a unit test to ensure this bug is fixed?

miner/worker.go
cffls: Makes sense, thanks!

miner/worker.go
kamuikatsurgi: Not yet, I'll add it.

kamuikatsurgi: Added in [5a5a194](https://github.com/0xPolygon/bor/pull/2094/commits/5a5a1940381168312c2eb02a6a3ca021db165bd2).

PR #2095, opened 2026-03-02
core/txpool: use reheap snapshot to avoid duplicate entries in heap

manav2401:
# Description

A different solution to the problem stated in https://github.com/0xPolygon/bor/pull/2087. 

This PR tries to add a new counter which just keeps track of the number of times reheap is done. We use this info while adding transactions to figure out whether to actually add the transaction to the heap or avoid it (so that there are no duplicates). This also ensures the calls still happen inside a go routine to keep the global tx pool lock as free as possible for the miner so that the performance benefits are intact. 

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -36,7 +36,7 @@ jobs:
           plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
           plugins: 'code-review@claude-code-plugins'
           prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
-          claude_args: '--model claude-opus-4-5-20251101'
+          claude_args: '--comment --model claude-opus-4-5-20251101'
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
 
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -51,7 +51,7 @@ jobs:
           additional_permissions: |
             actions: read
 
-          claude_args: '--model claude-opus-4-5-20251101'
+          claude_args: '--comment --model claude-opus-4-5-20251101'
 
           # Optional: Give a custom prompt to Claude. If this is not specified, Claude will perform the instructions specified in the comment that tagged it.
           # prompt: 'Update the pull request description to include a summary of changes.'
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # ─── BUILDER STAGE ───────────────────────────────────────────────────────────────
-FROM mirror.gcr.io/library/golang:1.26-alpine AS builder
+FROM golang:1.26-alpine AS builder
 
 ARG BOR_DIR=/var/lib/bor/
 ENV BOR_DIR=$BOR_DIR
@@ -22,7 +22,7 @@ RUN --mount=type=ssh \
     make bor
 
 # ─── RUNTIME STAGE ────────────────────────────────────────────────────────────────
-FROM mirror.gcr.io/library/alpine:3.21
+FROM alpine:3.23
 
 ARG BOR_DIR=/var/lib/bor/
 ENV BOR_DIR=$BOR_DIR
--- a/Dockerfile.alltools
+++ b/Dockerfile.alltools
@@ -7,12 +7,13 @@ ADD . /bor
 RUN cd /bor && make bor-all
 
 # Pull all binaries into a second stage deploy alpine container
-FROM alpine:latest
+FROM alpine:3.23
 
 RUN set -x \
     && apk add --update --no-cache \
     ca-certificates \
     && rm -rf /var/cache/apk/*
+
 COPY --from=builder /bor/build/bin/* /usr/bin/
 
 EXPOSE 8545 8546 30303 30303/udp
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,4 +1,4 @@
-FROM mirror.gcr.io/library/alpine:3.21
+FROM alpine:3.23
 
 ARG BOR_DIR=/var/lib/bor/
 ENV BOR_DIR=$BOR_DIR
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -125,6 +125,10 @@ var (
 	errNonContiguousHeaderRange = errors.New("non-contiguous headers in checkpoint range")
 )
 
+// maxAllowedFutureBlockTimeSeconds is the maximum number of seconds that a block
+// timestamp may exceed the local clock.
+const maxAllowedFutureBlockTimeSeconds = uint64(30)
+
 // SignerFn is a signer callback function to request a header to be signed by a
 // backing account.
 type SignerFn func(accounts.Account, string, []byte) ([]byte, error)
@@ -424,6 +428,12 @@ func (c *Bor) verifyHeader(chain consensus.ChainHeaderReader, header *types.Head
 		}
 		if parent == nil || now < parent.Time {
 			log.Error("Block announced too early post rio", "number", number, "headerTime", header.Time, "now", now)
+			return consensus.ErrFutureBlock
+		}
+		// Upper-bound check: a block whose timestamp is more than maxAllowedFutureBlockTimeSeconds
+		// ahead of the local clock is rejected.
+		if header.Time > now+maxAllowedFutureBlockTimeSeconds {
+			log.Error("Block timestamp too far in future post rio", "number", number, "headerTime", header.Time, "now", now)
 			return consensus.ErrFutureBlock
 		}
 	} else if c.config.IsBhilai(header.Number) {
--- a/consensus/bor/verify_header_test.go
+++ b/consensus/bor/verify_header_test.go
@@ -215,6 +215,54 @@ func TestVerifyHeader(t *testing.T) {
 			},
 			expectedError: consensus.ErrFutureBlock,
 		},
+		// Rio timestamp upper-bound tests: verify that far-future timestamps are rejected.
+		{
+			// A header with a timestamp set 100 years in the future must be
+			// rejected by the upper-bound check introduced in Rio.
+			name: "far-future timestamp in Rio mode is rejected",
+			setupChain: makeSetupChain(signerAddr, func(opts *chainSetupOptions) {
+				opts.rioBlock = big.NewInt(0)
+			}),
+			createHeader: func(t *testing.T, chain *core.BlockChain) *types.Header {
+				genesis := chain.HeaderChain().GetHeaderByNumber(0)
+				farFuture := uint64(time.Now().Unix()) + 100*365*24*3600 // year ~2126
+				return newStandardTestHeader(genesis, func(opts *headerOptions) {
+					opts.time = farFuture
+				})
+			},
+			expectedError: consensus.ErrFutureBlock,
+		},
+		{
+			// Boundary test: a timestamp just beyond the allowed future window is rejected.
+			name: "timestamp beyond allowed future bound in Rio mode is rejected",
+			setupChain: makeSetupChain(signerAddr, func(opts *chainSetupOptions) {
+				opts.rioBlock = big.NewInt(0)
+			}),
+			createHeader: func(t *testing.T, chain *core.BlockChain) *types.Header {
+				genesis := chain.HeaderChain().GetHeaderByNumber(0)
+				beyondBound := uint64(time.Now().Unix()) + maxAllowedFutureBlockTimeSeconds + 5
+				return newStandardTestHeader(genesis, func(opts *headerOptions) {
+					opts.time = beyondBound
+				})
+			},
+			expectedError: consensus.ErrFutureBlock,
+		},
+		{
+			// Regression: a normal block (timestamp slightly in the past) must still be
+			// accepted in Rio mode after the upper-bound check is added.
+			name: "normal timestamp in Rio mode is accepted",
+			setupChain: makeSetupChain(signerAddr, func(opts *chainSetupOptions) {
+				opts.rioBlock = big.NewInt(0)
+			}),
+			createHeader: func(t *testing.T, chain *core.BlockChain) *types.Header {
+				genesis := chain.HeaderChain().GetHeaderByNumber(0)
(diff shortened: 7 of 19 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -1134,17 +1134,18 @@ func (pool *LegacyPool) add(tx *types.Transaction, async bool) (replaced bool, e
 		// New transaction is better, replace old one
 		if old != nil {
 			pool.all.Remove(old.Hash())
-			pool.priced.Removed(1)
+			go pool.priced.Removed(1)
```
lucca30: Should we check if async flag is true here?

core/txpool/legacypool/list.go
```diff
@@ -644,20 +648,34 @@ func newPricedList(all *lookup) *pricedList {
 	}
 }
 
-// Put inserts a new transaction into the heap.
-func (l *pricedList) Put(tx *types.Transaction) {
+// Put inserts a new transaction into the heap. The `reheapSnapshot` field
+// denotes a counter of last re-heap to prevent duplicate entry.
+func (l *pricedList) Put(tx *types.Transaction, reheapSnapshot uint64) {
```
lucca30, line 653: Would be good covering the changes on `list.go` into `list_test.go`

core/txpool/legacypool/list.go
```diff
@@ -626,6 +626,10 @@ type pricedList struct {
 	// Number of stale price points to (re-heap trigger).
 	stales atomic.Int64
 
+	// Number of reheaps done. This counter should be incremented during every re-heap
+	// operation. It prevents in adding duplicate values to the heap later on.
+	reheaps atomic.Uint64
```
lucca30, line 631: Is an integer enough to identify when a duplicate happens?

core/txpool/legacypool/legacypool.go
```diff
@@ -1134,17 +1134,18 @@ func (pool *LegacyPool) add(tx *types.Transaction, async bool) (replaced bool, e
...
 			pendingReplaceMeter.Mark(1)
 			delete(pool.lastRebroadcast, old.Hash())
 		}
 		pool.all.Add(tx)
+		reheapCount := pool.priced.reheaps.Load()
```
lucca30, line 1146: If `reheap` runs between these 2 lines (1141-1142) (in another thread, for example) there is still a risk of double insertion.

Basically:

1. Reheap reads tx from `pool.all` → adds it to heap → increments `reheaps` to N+1
2. `reheaps.Load()` in the spawning goroutine reads N+1 (post-Reheap)
3. The async `Put` goroutine runs, checks `N+1 != reheaps.Load()` → `N+1 != N+1` → false → inserts again
4. Heap now has 2 entries for the same tx


If you want to reproduce the scenario I wrote locally 2 tests on `legacypool2_test.go` that deterministically reproduce it:

```
// TestReheapCountSnapshotAfterAdd proves the reheapSnapshot mechanism fails to
// prevent duplicate heap entries when Reheap() runs between pool.all.Add(tx)
// and pool.priced.reheaps.Load(). This is the exact race in enqueueTx (lines
// 1211-1212) and in add() (pending replacement path, ~lines 1140-1142).
//
// The test is fully deterministic: no goroutines, no lucky scheduling required.
// It reproduces the interleaving sequentially to prove the logic is unsound.
func TestReheapCountSnapshotAfterAdd(t *testing.T) {
	all := newLookup()
	priced := newPricedList(all)

	key, _ := crypto.GenerateKey()
	tx := pricedTransaction(0, 100000, big.NewInt(1), key)

	// Step 1: Add tx to lookup — equivalent to pool.all.Add(tx) in enqueueTx (line 1211).
	all.Add(tx)

	// Step 2: Reheap runs BETWEEN pool.all.Add(tx) and pool.priced.reheaps.Load().
	// In production this happens on another OS thread (GOMAXPROCS > 1) while add()
	// continues executing. Reheap finds tx in pool.all, adds it to the heap, and
	// increments the counter.
	priced.Reheap() // counter: 0 → 1; tx is now in the heap

	priced.reheapMu.Lock()
	afterReheap := priced.urgent.Len() + priced.floating.Len()
	priced.reheapMu.Unlock()

	if afterReheap != 1 {
		t.Fatalf("setup: expected 1 heap entry after Reheap, got %d", afterReheap)
	}

	// Step 3: Snapshot captured AFTER Reheap — equivalent to pool.priced.reheaps.Load()
	// on line 1212. Because Reheap already ran, the snapshot equals the current counter.
	snapshot := priced.reheaps.Load() // = 1 (post-Reheap)

	// Step 4: The async goroutine runs (equivalent to go pool.priced.Put(tx, snapshot)).
	// Check inside Put: reheapSnapshot (1) != reheaps.Load() (1) → FALSE → NOT skipped.
	// The tx is inserted again even though Reheap already added it.
	priced.Put(tx, snapshot)

	// Invariant: heap entries must equal all entries (one tx, one heap slot).
	priced.reheapMu.Lock()
	heapLen := priced.urgent.Len() + priced.floating.Len()
	priced.reheapMu.Unlock()

	allLen := all.Count()
	if heapLen != allLen {
		t.Errorf("BUG: heap has %d entries for %d tx — "+
			"reheapCount snapshot captured after Reheap allows duplicate insertion "+
			"(snapshot=%d, reheaps=%d)",
			heapLen, allLen, snapshot, priced.reheaps.Load())
	}
}

// TestReheapCountConcurrentRace forces the interleaving that production code is
// susceptible to using channel synchronization: a goroutine representing the async
// Removed trigger calls Reheap() after pool.all.Add(tx) but before the snapshot
// is captured, proving the race is real and not just a sequential "what-if".
func TestReheapCountConcurrentRace(t *testing.T) {
	all := newLookup()
	priced := newPricedList(all)

	key, _ := crypto.GenerateKey()
	tx := pricedTransaction(0, 100000, big.NewInt(1), key)

	addDone := make(chan struct{})
	reheapDone := make(chan struct{})

	// Goroutine models the async go pool.priced.Removed(1) (spawned at ~line 1136
	// in add()) which, when stales exceed the threshold, calls Reheap() on a
	// separate OS thread while add() continues executing.
	go func() {
		<-addDone              // wait: tx is now in pool.all
		priced.Reheap()        // tx is visible in all → added to heap; counter: 0 → 1
		close(reheapDone)
	}()

	// Main goroutine models the enqueueTx / add() path (lines 1211-1212).
	all.Add(tx)      // line 1211: pool.all.Add(tx)
	close(addDone)   // signal: Reheap goroutine may run now
	<-reheapDone     // wait: Reheap has completed (deterministically widening the window)

	// line 1212: snapshot captured — now equal to the post-Reheap value
	snapshot := priced.reheaps.Load() // = 1
	// Equivalent to: go pool.priced.Put(tx, snapshot)
	priced.Put(tx, snapshot)

	priced.reheapMu.Lock()
	heapLen := priced.urgent.Len() + priced.floating.Len()
	priced.reheapMu.Unlock()

	allLen := all.Count()
	if heapLen != allLen {
		t.Errorf("BUG CONFIRMED: heap=%d entries, all=%d tx — "+
			"duplicate caused by reheapCount snapshot race (snapshot=%d, reheaps=%d)",
			heapLen, allLen, snapshot, priced.reheaps.Load())
	}
}
```

manav2401: Hi Lucca, I don't think this test emulates the exact behavior of `add` or `enqueueTx`. Here's how this race condition is not an issue and calls are made in the following order. 

- add transaction to `pool.all` (note that this takes a lock over lookup struct so even if reheap has just started, it will be blocked by this lock). 
- we immediately read the `pool.priced.reheaps` value which means that we get the pre-reheap value and not the post one. 
- because lookup struct was locked and reheap was waiting on it, iterating over the struct would be time consuming (at least a few ms). we update the counter once this iteration is completed. hence we're sure that we'll get the pre-reheap value (which will be N in the example above). 
- this means you'll get N as the snapshot value and `pool.priced.Put` will not add duplicate entry as it was already taken from `pool.all`. 

So, in the above tests, if you move the snapshot read part to immediately after `all.Add(tx)` step, you'll see no races. thanks!

manav2401: although it's a pretty simple test which I will include by changing the order of read a bit to ensure there are no regressions later.

core/txpool/legacypool/list.go
manav2401: Done, thanks!

core/txpool/legacypool/list.go
manav2401: Yup. Resolving this as discussed over call.

core/txpool/legacypool/legacypool.go
lucca30: Yeah, I do also agree these tests forces some artificial situations to prove the point. But for the statement that

> if you move the snapshot read part to immediately after all.Add(tx) step, you'll see no races

Even though I believe it's true for almost all the cases, we still on probabilistic scenario rather than a deterministic one.

I asked Claude to help me on a test to force it. Here's the code so you can paste on `legacy_pool2_test.go`to check.

```
// TestReheapCountSchedulerRace demonstrates the race without artificial channel
// synchronization between Add() and the snapshot read to show that
// placing the snapshot read immediately after Add() not necessarily prevents the race.
//
// With GOMAXPROCS(1), goroutines run cooperatively. runtime.Gosched() between
// Add() and reheaps.Load() yields the CPU to the pre-queued Reheap goroutine.
// For a 1-element pool (fast Range()), Reheap completes entirely before main
// resumes to read the snapshot — no explicit <-reheapDone needed.
//
// In production with GOMAXPROCS > 1, the Gosched() is replaced by true parallel
// execution on a separate OS thread (e.g., runReorg's Reheap at line 1673
// running concurrently with add() on another core).
func TestReheapCountSchedulerRace(t *testing.T) {
	defer runtime.GOMAXPROCS(runtime.GOMAXPROCS(1))

	all := newLookup()
	priced := newPricedList(all)

	key, _ := crypto.GenerateKey()
	tx := pricedTransaction(0, 100000, big.NewInt(1), key)

	done := make(chan struct{})

	// Pre-queue the Reheap goroutine. With GOMAXPROCS=1, it is queued but
	// NOT running — main goroutine holds the CPU.
	go func() {
		priced.Reheap() // runs AFTER Add() (scheduled via Gosched below)
		close(done)
	}()

	// line 1215: pool.all.Add(tx) — goroutine still queued, not running
	all.Add(tx)

	// Yield CPU. With GOMAXPROCS=1, the Reheap goroutine now runs to completion:
	//   Range() starts with no lookup lock contention (Add already released its lock),
	//   sees tx, adds it to heap, increments counter (0→1).
	// This models OS preemption between lines 1215 and 1216 in enqueueTx().
	runtime.Gosched()

	// line 1216: snapshot captured — Reheap has already completed during Gosched()
	snapshot := priced.reheaps.Load() // = 1 (post-Reheap)
	priced.Put(tx, snapshot)          // 1 == 1 → NOT skipped → inserts AGAIN

	<-done // goroutine already done; just cleanup

	priced.reheapMu.Lock()
	heapLen := priced.urgent.Len() + priced.floating.Len()
	priced.reheapMu.Unlock()

	require.Equal(t, all.Count(), heapLen,
		"BUG: heap has %d entries for %d tx — "+
			"scheduler preemption between Add() and reheaps.Load() causes duplicate "+
			"(snapshot=%d, reheaps=%d)",
		heapLen, all.Count(), snapshot, priced.reheaps.Load())
}
```

It forces a scenario with `GOMAXPROCS(1)` and yielding the CPU which can be reflected on production by the true parallel work (harder to emulate). 

To actually address it, the only solution I can see is by removing the async on these parts, as done on original PR https://github.com/0xPolygon/bor/pull/2087. So I was just wondering how much of performance do we get by keeping it async. 

So would be good making a benchmark around this function so we can evaluate how worth will it be to dive into a deterministic solution for the asynchronous approach.

WDYT?

manav2401: Hi Lucca, I think even in this emulation, the snapshot being read is after re-heap which won't be the case in production. Although I agree that there's a very very small window in which new snapshot is read which may lead to duplicate values. 

In production, even if they run in parallel, the window is really small if you follow the following order. 
- `pool.all.Add(tx)` which acquires `pool.all.lock.Lock()`
- `Reheap()` in another routine acquires `reheapMu`
- `Reheap()` has to iterate over `pool.all` which needs `pool.all.lock.Lock()`. 
- The lock might be acquired by step 1 in which case Reheap is temporarily blocked while the tx is being added. 
- As soon as the lock is free, we read the snapshot.
- Meanwhile, the chances of `Reheap` acquiring the `pool.all.lock` back, iterating over entries (even if small number of them), and releasing the lock all before the snapshot is being read in another routine is next to 0. 
- The snapshot counter is only updated once this iteration is completed in reheap. 

Although, I have a solution which requires some refactor but that would solve for this small window as well. The idea is to update the reheap snapshot counter under `pool.all.lock` so it's fully deterministic. Will check feasibility of this idea. 

> To actually address it, the only solution I can see is by removing the async on these parts, as done on original PR https://github.com/0xPolygon/bor/pull/2087. So I was just wondering how much of performance do we get by keeping it async.
Re. this, I think there are 2 things to consider. 
1. The performance gain by keeping async is very high as reheap lock can be acquired for >100ms in lot of cases (500ms worst case). This can be a big bottleneck in block building. 
2. Removing async behavior and doing everything synchronous can still lead to this logical race condition if you think about this scenario. You do `pool.all.Add`, reheap starts immediately and acquires `reheapMu` lock so the next call `pool.priced.Put` is blocked if it's not called in a new go-routine. This means a duplicate entry will still be present even after preventing the function from running in a new go-routine. 

Thanks!

lucca30: Yeah agreed on the low risk involved. Also, the old solution was living in production for a long time and the only time we caught it failing was in tests. Anyway, would be good proposing this deterministic approach and the benchmark to validate.

manav2401:
@claude review this PR focusing on performance and security. Also, make sure no regression and new race condition is introduced. Refer to the PR comments (and the original PR mentioned in description) for more context if needed.

manav2401:
@claude review this PR focusing on performance and security. Also, make sure no regression and new race condition is introduced. Refer to the PR comments (and the original PR mentioned in description) for more context if needed.

PR #2098, opened 2026-03-02
core/rawdb: added filesystem based  witness storage (with --witness.filestore flag)

pratikspatil024:
# Description

Witness payloads are stored once per block (~every 2s) and can exceed 10MB. Storing them in Pebble increases write amplification, flush/compaction pressure, and contention on a resource shared by all chain data. Moving witness blobs to the filesystem reduces DB I/O while preserving existing behavior.

- Introduces a WitnessStore interface that abstracts witness blob storage, with two backends: database (existing behavior) and filesystem (via `--witness.filestore` flag)
- The filesystem backend writes one file per block using atomic temp+rename, 2-level hex directory sharding, and keeps size metadata in Pebble for P2P queries
- Reads fall back to Pebble transparently, enabling seamless migration without downtime. Old witnesses are naturally pruned away within 6400 blocks

### Witness file path layout

`<chaindata>/witnesses/<hex[0:2]>/<hex[2:4]>/0x<block-hash>.rlp`
Witnesses are stored under a `witnesses/` subdirectory within `chaindata` to keep them co-located with chain data while isolated from Pebble's internal files. A 2-level hex directory sharding (`ab/cd/`) distributes files across up to `65,536` buckets, avoiding filesystem performance degradation from large flat directories. With `~6400` files retained at any time, each bucket holds fewer than 1 file on average. The block hash is used as the filename since it's the primary key for all witness operations, and the `.rlp` extension indicates the encoding format.

### Transition behaviour

- Pebble to Filesystem (enable `--witness.filestore`): New witnesses go to the filesystem. Old witnesses in Pebble remain accessible via the read fallback and are naturally pruned within 6400 blocks. No manual intervention needed.
- Filesystem to Pebble (disable `--witness.filestore`): New witnesses go to Pebble. Filesystem witnesses become orphaned, that is, the DB backend does not read or delete them. A warning is logged at startup: `Witness directory exists but --witness.filestore is disabled`. Orphaned files can be removed manually with `rm -rf <chaindata>/witnesses/`.
- Filesystem to Pebble to Filesystem (round-trip): When switching back to filesystem mode, a small number of orphaned files from the first filesystem period may persist. The pruner's cursor advanced during the Pebble period and will not revisit those earlier block heights. These files are harmless and can be ignored or removed manually.

### Write (the main operation this optimizes)

  | Size | DB (Pebble on disk) | FS | FS advantage |
  |------|----------------------|----|--------------|
  | 100KB | ~1ms, 3.3KB alloc | ~200μs, 2.5KB alloc | **~5x faster**, similar allocs |
  | 1MB | ~12ms, 2.3MB alloc | ~320μs, 2.5KB alloc | **~38x faster**, **~900x less memory** |
  | 5MB | ~48ms, 10.9MB alloc | ~1.6ms, 2.5KB alloc | **~30x faster**, **~4400x less memory** |

The DB backend's allocations scale with payload size due to Pebble's internal memtable copies and compaction buffers, while latency is dominated by flush/compaction overhead. The FS backend maintains a constant ~2.5KB allocation regardless of payload size.

### Read

  | Size | DB (Pebble on disk) | FS |
  |------|----------------------|----|
  | 100KB | ~3.8μs | ~29μs |
  | 1MB | ~120μs | ~122μs |
  | 5MB | ~200μs | ~279μs |

At realistic witness sizes (1–10MB), read performance is nearly equivalent between the two backends. Hot reads are served by the in-memory `witnessCache` LRU and rarely hit disk.

### Delete

  | Backend | Latency |
  |---------|---------|
  | DB (Pebble on disk) | ~2μs |
  | FS | ~100μs |

FS deletes are slower (syscall vs tombstone marker), but pruning runs in the background every 120s on old blocks and is not latency-sensitive.
# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/node/database.go
+++ b/node/database.go
@@ -71,7 +71,7 @@ func openDatabase(o internalOpenOptions) (ethdb.Database, error) {
 	}
 	// Resolve witness store directory relative to the database directory.
 	var witnessStoreDir string
-	if o.WitnessFileStore && o.directory != "" {
+	if o.directory != "" {
 		witnessStoreDir = filepath.Join(o.directory, "witnesses")
 	}
 


core/rawdb/witness_store_bench_test.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+		})
+	}
+}
+
+func BenchmarkReadWitness_DB(b *testing.B) {
+	for _, size := range []int{100 * 1024, 1024 * 1024, 5 * 1024 * 1024} {
+		payload := makePayload(size)
+		b.Run(sizeLabel(size), func(b *testing.B) {
+			db := NewMemoryDatabase()
```
lucca30: I think would be worth do the bench against an actual PebbleDB on disk, instead of Memory DB. So them we can prepare test loads that forces compaction and gives more clear results.

pratikspatil024: make sense, thanks. I'll update it

pratikspatil024: updated [here](https://github.com/0xPolygon/bor/pull/2098/commits/f55b6c1346cc8e78dc3b9339fdf6faa2e432ec06).

core/rawdb/witness_store.go
```diff
@@ -0,0 +1,31 @@
...
+	"fmt"
+	"path/filepath"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// WitnessStore abstracts witness blob storage so callers can transparently
+// use either the key-value database (Pebble/LevelDB) or the filesystem.
+type WitnessStore interface {
+	ReadWitness(hash common.Hash) []byte
```
cffls, line 13: How is the transition phase handled? For example, if a node is restarted with witness file store enabled for the first time, and was requested with an old witness, which happen to exist in its database, will the node have a fallback logic to check its database? I think the same question applies to pruning as well.

cffls: My bad, just saw the fallback logic is implemented below.

cffls (approved):
lgtm!

internal/cli/server/config.go
```diff
@@ -1601,6 +1604,7 @@ func (c *Config) buildEth(stack *node.Node, accountManager *accounts.Manager) (*
 	n.EnableParallelStatelessImport = c.Witness.EnableParallelStatelessImport
 	n.EnableParallelStatelessImportWorkers = c.Witness.ParallelStatelessImportWorkers
 	n.WitnessAPIEnabled = c.Witness.WitnessAPI
+	n.WitnessFileStore = c.Witness.FileStore
```
cffls, line 1608: Can we enable this by default?

pratikspatil024: Done [here](https://github.com/0xPolygon/bor/pull/2098/commits/2b0a8460beff9ecaf61d73fca26a8ff9c0c20807), thanks!

pratikspatil024:
@claude - review this PR.

PR #2100, opened 2026-03-03
internal/cli, packaging, docs: add config.toml updater, refresh templates, and document workflow

kamuikatsurgi:
# Description

- Added a new Bor config updater at `scripts/update-config-toml/main.go` to keep `config.toml` templates aligned with the current `server.DefaultConfig()` and config schema.
- Updater now merges existing file values on top of defaults (preserves intentional overrides), fills missing keys, and rewrites files in a consistent format.
- Regenerated Bor config templates with improved comments/formatting and synced defaults.
- Documented workflow in CLAUDE.md (dry-run/apply commands, verification expectations, and failure-handling guidance).


kamuikatsurgi:
> LGTM. Few small nits.
> 
> * Might be good to add this in Makefile under the `make docs` field - we anyways run that to update docs.
> * There's too much context about this in CLAUDE.md which I am not sure is needed. We should remove it or trim it down to a line or two (with just the `make docs` command maybe).

Done.

cmd/generate-default-config/main.go
pratikspatil024, line 1: would be better to move this file to `internal/cli`

docs/cli/example_config.toml
pratikspatil024, line 1: Is it possible to beautify this file? The previous one looked neat.
Also, the previous one had a few examples for references (it was not the default one) for example `eth.requiredblocks`, it would be good to have that. If it is not possible, you can create a new file and keep the older one in the codebase as outdated-with-examples

cmd/generate-default-config/main.go
kamuikatsurgi: Moved to `cmd/generate-default-config/main.go` where `cmd/clidoc/main.go` exists.

docs/cli/example_config.toml
kamuikatsurgi: I do have the standard .toml formatting in the code, so the file is fully formatted. I would prefer to have all the comments explaining the configs and their default values. I can create a new file instead of modifying the existing one.

kamuikatsurgi: Added a new file `default_config.toml`.

PR #2102, opened 2026-03-04
fix: avoid mirror.gcr.io images

kamuikatsurgi:


lucca30:
Do we have sure we are not going to reach limits again with DockerHub? Also, would be good not using `latest` tag for `alpine` images.

kamuikatsurgi:
> Do we have sure we are not going to reach limits again with DockerHub? Also, would be good not using `latest` tag for `alpine` images.

Fixed.

PR #2105, opened 2026-03-04
fix(server): wire Identity config to node.UserIdent in buildNode

minhd-vu:
# Description

Fix the `--identity` flag not being applied when using `bor server`.

## Problem

The `bor server` command accepts an `Identity` config field but `buildNode()` never sets `node.Config.UserIdent`, causing the identity to be silently ignored.

This means nodes running with `bor server --identity=xxx` always show the default client name in P2P communications, regardless of the `--identity` flag value.

Note: The legacy CLI (`bor --identity=xxx`) is unaffected as it uses `setNodeUserIdent()` inherited from go-ethereum.

## Root Cause

The `bor server` command uses a separate HCL-based configuration system (`internal/cli/server/`) that doesn't exist in upstream go-ethereum. The `buildNode()` function manually constructs a `node.Config` struct but was missing the `UserIdent` field:

```go
cfg := &node.Config{
    Name:    clientIdentifier,
    // UserIdent was missing here
    DataDir: c.DataDir,
    ...
}
```

## Fix

Add the missing field assignment in `buildNode()`:

```go
cfg := &node.Config{
    Name:      clientIdentifier,
    UserIdent: c.Identity,  // <-- Added
    DataDir:   c.DataDir,
    ...
}
```

# Changes

# Checklist

# Cross repository changes

## Testing

- [ ] I have added unit tests

### Manual tests

1. Start bor with identity: `bor server --identity=my-node ...`
2. Connect a peer and verify the client name includes "my-node"
3. Alternatively, use `polycli p2p ping <enode>` to observe the peer's client name

### DIFF
--- a/internal/cli/server/config.go
+++ b/internal/cli/server/config.go
@@ -1713,6 +1713,7 @@ func (c *Config) buildNode() (*node.Config, error) {
 
 	cfg := &node.Config{
 		Name:                  clientIdentifier,
+		UserIdent:             c.Identity,
 		DataDir:               c.DataDir,
 		DBEngine:              c.DBEngine,
 		KeyStoreDir:           c.KeyStoreDir,


cffls:
Does this work for config.toml?

minhd-vu:
> Does this work for config.toml?

It should work for `config.toml`. The Identity field has both tags:

```go
Identity string `hcl:"identity,optional" toml:"identity,optional"`
```

kamuikatsurgi (approved):

PR #2106, opened 2026-03-05
internal/cli, node, rpc: port missing geth flags

kamuikatsurgi:
# Description

Flags added:

- `--rpc.batch-request-limit` / `--rpc.batch-response-max-size` - Geth's node.Config fields used in every RPC handler were never set by bor, always defaulting to zero (no limits).
- `--vmtrace` / `--vmtrace.jsonconfig` - `eth/backend.go` uses these to create a live tracer at startup; bor never wired them.

Bor's `RPCBatchLimit` was only applied to `HTTP` servers via `httpServer.SetRPCBatchLimit()`, it was never applied to `WebSocket` or `IPC` handlers. Geth's `BatchRequestLimit`/`BatchResponseMaxSize` are set on the in-process handler and passed to every RPC handler, including `WS` and `IPC`.

### DIFF
--- a/internal/cli/server/testdata/default.toml
+++ b/internal/cli/server/testdata/default.toml
@@ -196,4 +196,3 @@ devfakeauthor = false
   addr = "127.0.0.1"
   memprofilerate = 524288
   blockprofilerate = 0
-  cpuprofile = ""


internal/cli/server/command.go
```diff
@@ -219,6 +219,12 @@ func (c *Command) Run(args []string) int {
 		}()
 	}
 
+	if err := StartCPUProfile(c.config.Pprof.CPUProfile); err != nil {
```
manav2401: I don't think this flag is meant to be used with usual bor server run command. 

In geth, it's added in `internal/debug` and not `cmd` which suggests that it's supposed to be used separately (not with main geth process) to debug. 

In our `bor debug pprof` command, it anyways captures CPU profile so maybe we can skip this flag.

internal/cli/server/server.go
```diff
@@ -310,6 +314,38 @@ func (s *Server) Stop() {
 	}
 }
 
+// StartCPUProfile starts CPU profiling, writing to the given file.
+// Call StopCPUProfile to flush and close the file.
+func StartCPUProfile(file string) error {
```
manav2401: We already have this at `internal/cli/server/pprof/proof.go`

internal/cli/server/flags.go
```diff
@@ -63,11 +75,17 @@ func (c *Command) Flags(config *Config) *flagset.Flagset {
 		Value:   &c.cliConfig.KeyStoreDir,
 		Default: c.cliConfig.KeyStoreDir,
 	})
-	f.Uint64Flag(&flagset.Uint64Flag{
-		Name:    "rpc.batchlimit",
```
manav2401, line 67: This will break for those who are using these flags. We should have both flags in at least one release and give a warn log about deprecating this. Later in next release, we remove this. 

Also, maybe both should have same default to begin with?

internal/cli/server/command.go
kamuikatsurgi: Okay, will remove it.

internal/cli/server/flags.go
kamuikatsurgi: > This will break for those who are using these flags. We should have both flags in at least one release and give a warn log about deprecating this. Later in next release, we remove this.

It's just a flag name change. Node operators wouldn't mind changing a flag name, imo, if we correctly describe this in the release notes. wdyt?

kamuikatsurgi: The older flag had bugs and unclean code impl, keeping both would make impl further complicated.

Bug:
```
 Bor's RPCBatchLimit was only applied to HTTP servers via httpServer.SetRPCBatchLimit(), it was never applied to WebSocket or IPC handlers. Geth's BatchRequestLimit/BatchResponseMaxSize are set on the in-process handler and passed to every RPC handler, including WS and IPC.
```

Impl:
```go
if batch {
		if s.BatchLimit > 0 && len(reqs) > int(s.BatchLimit) {
			if err1 := codec.writeJSON(ctx, errorMessage(fmt.Errorf("batch limit %d exceeded: %d requests given", s.BatchLimit, len(reqs))), true); err1 != nil {
				log.Warn("WARNING - requests given exceeds the batch limit", "err", err1)
				log.Debug("batch limit %d exceeded: %d requests given", s.BatchLimit, len(reqs))
			}
		} else {
			//nolint:contextcheck
			h.handleBatch(reqs)
		}
```

Whereas, it's just a single line now:
```go
h.handleBatch(reqs)
```

manav2401: Okay yeah do make a note of informing this in release notes. 

Also, while you're doing this, can you remove the `log-level` flags as well because it's deprecated and we have `verbosity`. Remove relevant code of handling both flags as well. Thanks!

internal/cli/server/command.go
kamuikatsurgi: Removed.

internal/cli/server/server.go
kamuikatsurgi: Removed.

internal/cli/server/flags.go
kamuikatsurgi: `log-level` was already removed in my last PR.

manav2401 (approved):

marcello33 (approved):

PR #2111, opened 2026-03-05
eth: fix overflow impacting milestone lock and syncing

marcello33:
# Description

Fixes issue 66.


lucca30:
@claude Do a careful review on this PR. Looking for any potential bug introduced by the changes or vulnerabilities. Also classify if this PR is a high, medium or low risk change in terms of how much time we need to test it before feel safe to trust it.

Check if the tests are aligned with what the fix is proposed for and if the fix address the issue it wants to fix.

eth/bor_api_backend.go
```diff
@@ -41,7 +45,12 @@ func (b *EthAPIBackend) GetRootHash(ctx context.Context, starBlockNr uint64, end
 }
 
 // GetVoteOnHash returns the vote on hash
-func (b *EthAPIBackend) GetVoteOnHash(ctx context.Context, starBlockNr uint64, endBlockNr uint64, hash string, milestoneId string) (bool, error) {
+func (b *EthAPIBackend) GetVoteOnHash(ctx context.Context, _ uint64, endBlockNr uint64, hash string, milestoneId string) (bool, error) {
+	// Reject invalid block numbers (overflowing with the confirmation offset or exceeding the valid range).
+	if endBlockNr > math.MaxInt64-16 {
```
lucca30: The `16` here must always match the `uint64(16)` on line 67. If the confirmation offset is ever changed, this guard needs updating too — otherwise the overflow protection silently breaks. Consider extracting a named constant (e.g. `tipConfirmationOffset = 16`) shared by both lines.

eth/bor_api_backend_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	tests := []struct {
+		name       string
+		endBlockNr uint64
+	}{
+		{"max uint64", math.MaxUint64},
+		{"max uint64 minus 15", math.MaxUint64 - 15},
+		{"max int64", math.MaxInt64},
+		{"max int64 minus 15", math.MaxInt64 - 15},
+	}
```
lucca30, line 24: The table only covers cases that should be rejected. It's worth adding a case for `math.MaxInt64 - 16` (the exact boundary) to assert it is allowed — so if the guard is ever tightened accidentally, a test catches it. e.g. `{"max int64 minus 16 (boundary, should pass)", math.MaxInt64 - 16}` with an assertion that `err == nil`.

cmd/keeper/go.sum
lucca30, line 170: It seems `go.sum` is incomplete.

Can you run the following and commit the updated `go.sum`?

```
cd cmd/keeper && go mod tidy
```

PR #2118, opened 2026-03-06
[WIP] Don't merge

marcello33:


triedb/pathdb/buffer.go
```diff
@@ -120,10 +130,18 @@ func (b *buffer) empty() bool {
...
 }
 
-// full returns an indicator if the size of accumulated content exceeds the
-// configured threshold.
+// full returns an indicator if the buffer should be flushed.
+// A flush is triggered when trie nodes exceed their allocation or when
+// the total buffer size exceeds the hard limit.
 func (b *buffer) full() bool {
-	return b.size() > b.limit
+	return b.nodes.size > b.nodeLimit || b.size() > b.limit
```
cffls, line 137: I am trying to think if this is going to trigger flush more frequently if we hit 20% of allocation for trie nodes quickly even when the cache for state isn't full. Is it possible only to flush only trie or state individually?

cffls: nvm, saw the `shouldCarryStates()` below.

PR #2120, opened 2026-03-09
eth/protocols/wit: remove peer lock held across p2p.Send to prevent broadcast stalls and added metrics

pratikspatil024:
# Description

- Remove p.lock from the WIT protocol peer, it was held across synchronous `p2p.Send()` calls (up to `20s` write timeout), which could block the block broadcast loop when a peer was slow or disconnecting
- Make `KnownCache` thread-safe internally so external locking is no longer needed
- Add metrics for observability: `wit/peer/send_witness_duration` and `eth/broadcast_loop_duration`

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/protocols/wit/broadcast.go
+++ b/eth/protocols/wit/broadcast.go
@@ -1,13 +1,6 @@
 package wit
 
-import (
-	"time"
-
-	"github.com/ethereum/go-ethereum/log"
-	"github.com/ethereum/go-ethereum/metrics"
-)
-
-var witnessSendTimer = metrics.NewRegisteredTimer("wit/peer/send_witness_duration", nil)
+import "github.com/ethereum/go-ethereum/log"
 
 // broadcastWitness is a write loop that multiplexes witness and witness announcements
 // to the remote peer. The goal is to have an async writer that does not lock up
@@ -18,11 +11,9 @@ func (p *Peer) broadcastWitness() {
 	for {
 		select {
 		case witness := <-p.queuedWitness:
-			start := time.Now()
 			if err := p.sendNewWitness(witness); err != nil {
 				return
 			}
-			witnessSendTimer.Update(time.Since(start))
 			p.logger.Debug("propagated witness", "hash", witness.Header().Hash())
 
 		case packet := <-p.queuedWitnessAnns:
--- a/eth/protocols/wit/peer_test.go
+++ b/eth/protocols/wit/peer_test.go
@@ -13,7 +13,6 @@ import (
 	"github.com/ethereum/go-ethereum/core/stateless"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
-	"github.com/ethereum/go-ethereum/metrics"
 	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 )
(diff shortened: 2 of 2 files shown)


eth/protocols/wit/peer.go
```diff
@@ -44,9 +44,6 @@ type Peer struct {
 	resDispatch chan *response // Dispatch channel to fulfill witness requests
 
 	term chan struct{} // Termination channel to stop the broadcaster
-	// TODO(@pratikspatil024) - review all the instances of the lock and unlock
```
pratikspatil024, line 47: 😈

kamuikatsurgi:
Looks like erigon got stuck at block 256 in the E2E test. Could you take a look and see if these changes did not affect anything?

adamdossa:
@claude Can you do a detailed review of this PR.

eth/protocols/wit/peer.go
```diff
@@ -275,8 +242,11 @@ func newKnownCache(max int) *KnownCache {
 	}
 }
 
-// Add adds a witness to the set.
+// Add adds a witness to the set, evicting old entries if at capacity.
 func (k *KnownCache) Add(hash common.Hash) {
+	k.mu.Lock()
```
manav2401, line 247: The `hashes` uses a mapset which is thread safe. Do we still need a lock here?

pratikspatil024: The lock protects the eviction logic. Without it, two concurrent `Add()` calls could both check `Cardinality()`, both decide to `Pop()`, and over-evict. The mapset is threadsafe per operation, but the check then pop then add sequence needs to be atomic.

pratikspatil024: Also @claude WDYT? About @manav2401's comment

manav2401: Ah I see, thanks!

jhkimqd:
Quick question @pratikspatil024 - I could be missing something, but while testing this on Kurtosis-PoS, it seems like the `wit/peer/send_witness_duration` metric is not incrementing - I have a mixture of Erigon, Archive nodes, RPCs snycing/producing witnesses, etc.. Would we need to enable some specific metric to see this count increase?

For the functionality, I've tested on separate Kurtosis devnets on Bor images with and without this fix (along with a custom timeout added to simulate slow `p2p.Send()`), and I was seeing different behaviors  - span rotations on the previous images, and normal network progression with the fix which seems to be expected!

pratikspatil024:
Awesome, thanks for looking into this and testing this fix, @jhkimqd!

Looking at the `wit/peer/send_witness_duration` metric not increasing issue, I back tracked this part of the code where the metric is set:
```
		case witness := <-p.queuedWitness:
			start := time.Now()
			if err := p.sendNewWitness(witness); err != nil {
				return
			}
			witnessSendTimer.Update(time.Since(start))
			p.logger.Debug("propagated witness", "hash", witness.Header().Hash())

```

and found out that the entry point which pushed the witnesses into the `queuedWitness` channel is never used!!
```
// TODO(@pratikspatil024) - use this to broadcast the witness
// BroadcastWitness broadcasts the witness to all peers who are not aware of it
func (h *handler) BroadcastWitness(witness *stateless.Witness) {
	// broadcast the witness to all peers who are not aware of it
	for _, peer := range h.peers.peersWithoutWitness(witness.Headers[0].Hash()) {
		peer.Peer.AsyncSendNewWitness(witness)
	}
}
```

So the metric not increasing makes sense. The delay in `p2p.Send()` would not be from this `BroadcastWitness` but from responding to a witness fetching request.

PR #2123, opened 2026-03-09
Adding workflow to publish the docker image on GHCR

0xsajal:
# Description

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/.github/workflows/release_ghcr.yml
+++ b/.github/workflows/release_ghcr.yml
@@ -1,4 +1,4 @@
-name: build_and_publish_to_ghcr.yml
+name: Release GHCR
 
 on:
   push:


adamdossa:
@claude Any issues you can see with this new workflow? Any improvements you'd suggest?

0xsajal:
Addressing the comments from Claude:

1. Missing Dockerfile Specification - Since we use the default dockerfile we do not need to explicitly mention it in the workflow, it gets picked up automatically.
2. Missing Genesis Files - The node operators need to download this file (depending on the network)after they pull the image, so we do not need to add this step here.
3. No Build Caching - I tested this on our tester repos and it does not make any difference on the time it takes to build and publish the image on GHCR. Hence I removed these lines to minimise the number of lines on the workflow file.
4.  IMAGE_NAME Casing Inconsistency - We want it to be consistent with the repo name and so I kept it this way.
5. Missing latest Tag - The latest tag will be added to the latest image when published on GHCR.
6. Workflow Name Mismatch - I used the naming standards we keep on pos-ops, this will not break anything.
7. No PR Trigger for Testing - We already tested it on the tester repos and it triggers when a new tag is created.
8. Missing Provenance/SBOM Flags - Claude says we did it in a better way.

PR #2134, opened 2026-03-09
Slow TX Observability, Per-TX Latency Metric, and Timeout-Interrupt Diagnostics

lucca30:
# Description
## Summary

This branch adds miner observability and diagnostics in three areas:

1. **Top-10 slow transaction tracking in 10-minute tumbling windows** (producer-only logging).
2. **Per-transaction apply latency metric** for tail analysis.
3. **Timeout interruption diagnostics** with precise timing context.

It also adds dedicated unit tests and benchmark coverage for the new tracker.

---

### 1) Slow tx tracking (windowed top-K)

- Tracks the slowest successful tx applications in a bounded **top 10** structure.
- Data structure: min-heap (`K=10`) with:
  - `O(1)` fast reject when not in top-K
  - `O(log K)` insert/replace when qualifying
- Logging cadence: every **10 minutes**, then reset.
- Logging routine runs only when the node is a block producer (`worker.IsRunning()`).
- If not producing during a tick, tracker is reset without logging to avoid stale carryover.

### 2) Timeout interruption diagnostics

On block-build timeout interruption, logs now include richer context:

- `flagSetAt`
- `flagToAbortDelay`
- `flagToTxInterruptDelay` (when observed)
- last tx context (`txHash`, `txIndex`, `sender`, `txElapsed`)

Also includes dedicated handling/logging for `vm.ErrInterrupt` in tx execution flow.

### 3) Per-tx duration metric

- Adds `worker/txApplyDuration` metric.
- Updated on successful tx application.
- Uses custom timer reservoir (`ExpDecaySample(8192, 0.015)`) to preserve tail visibility at high throughput.

---

## Logging Format Helper

- Added `common.PrettyTime` (UTC format: `MM-DD|HH:MM:SS.mmm`) for consistent log time rendering.
- Includes unit test coverage in `common/format_test.go`.

---

## Test Changes

- New unit tests for top-K tracker:
  - keeps top 10
  - descending snapshot order
  - reset semantics

---

## Benchmark Impact

Benchmark command:

```bash
go test ./miner -run '^$' -bench BenchmarkSlowTxTopWindowTrackerImpact -benchmem
```

Representative run (Apple M4 Pro):

- `Baseline/200tx`: `204.0 ns/op`
- `TrackerMostlyNoInsert/200tx`: `446.7 ns/op`
- `TrackerMixedInsert/200tx`: `853.0 ns/op`

- `Baseline/800tx`: `794.0 ns/op`
- `TrackerMostlyNoInsert/800tx`: `1328 ns/op`
- `TrackerMixedInsert/800tx`: `2492 ns/op`

### Delta vs baseline (impact view)

Assuming 200 tx insertions per block:

- Mostly-no-insert path: `446.7 - 204.0 = +242.7 ns/block` (about `+1.21 ns/tx`)
- Mixed-insert path: `853.0 - 204.0 = +649.0 ns/block` (about `+3.25 ns/tx`)

Assuming 800 tx insertions per block:

- Mostly-no-insert path: `1328 - 794 = +534 ns/block` (about `+0.67 ns/tx`)
- Mixed-insert path: `2492 - 794 = +1698 ns/block` (about `+2.12 ns/tx`)

Allocation profile in tracker modes:

- `480 B/op`, `10 allocs/op` (bounded by `K=10`, not by tx volume)

Conclusion:

- For common-path behavior, overhead is sub-2 ns/tx.
- Even for mixed insertion behavior, overhead stays within low single-digit ns/tx.
- No materially relevant performance regression for 200/800 tx insertion assumptions.

---

## Validation Performed

- `go test ./common`
- `go test ./internal/cli/server`
- `go test ./miner -run TestSlowTx`
- `go test ./miner -run '^$' -bench BenchmarkSlowTxTopWindowTrackerImpact -benchmem`

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/miner/slow_tx_tracker_test.go
+++ b/miner/slow_tx_tracker_test.go
@@ -2,6 +2,7 @@ package miner
 
 import (
 	"math/big"
+	"math/rand"
 	"testing"
 	"time"
 
@@ -19,6 +20,36 @@ func TestSlowTxTopTrackerKeepsTop10(t *testing.T) {
 		tracker.Add(txTimingEntry{
 			hash:     common.BigToHash(big.NewInt(int64(i))),
 			duration: time.Duration(i) * time.Millisecond,
+		})
+	}
+
+	got := tracker.SnapshotAndReset()
+	require.Len(t, got, slowTxTopKSize)
+
+	for i := 0; i < slowTxTopKSize; i++ {
+		expectedDuration := time.Duration(30-i) * time.Millisecond
+		require.Equal(t, expectedDuration, got[i].duration)
+	}
+}
+
+func TestSlowTxTopTrackerRandomOrder(t *testing.T) {
+	t.Parallel()
+
+	tracker := newSlowTxTopTracker()
+
+	// Insert 30 entries in random order and verify top 10 are still correct.
+	durations := make([]int, 30)
+	for i := range durations {
+		durations[i] = i + 1
+	}
+
+	rng := rand.New(rand.NewSource(42))
+	rng.Shuffle(len(durations), func(i, j int) { durations[i], durations[j] = durations[j], durations[i] })
+
+	for _, d := range durations {
+		tracker.Add(txTimingEntry{
+			hash:     common.BigToHash(big.NewInt(int64(d))),
+			duration: time.Duration(d) * time.Millisecond,
 		})
 	}
 
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1234,6 +1234,7 @@ mainloop:
 			txCommitInterruptCounter.Inc(1)
 			logCtx := []interface{}{
 				"number", env.header.Number.Uint64(),
+				"headerTime", common.PrettyTime(time.Unix(int64(env.header.Time), 0)),
 			}
 			if flagSetAt := w.interruptFlagSetAt.Load(); flagSetAt > 0 {
 				flagSetTime := time.Unix(0, flagSetAt)


lucca30:
@claude Do a review again, see if you find any other issues

kamuikatsurgi:
IMO, the per tx latency isn't much useful as it would just be a another metric. It doesn't particularly tell use which txs are slower vs which are faster. I don't see how this can help us.

lucca30:
> IMO, the per tx latency isn't much useful as it would just be a another metric. It doesn't particularly tell use which txs are slower vs which are faster. I don't see how this can help us.

@kamuikatsurgi 
I agree that this data has no good use for the things we have today.
But the goal is to help us refine which should be the ideal `--miner.slowtxthreshold` value to just capture relevant outliers. Today we have no idea on what is the avg time a tx take, how outliers behave, if they are common or not etc

internal/cli/server/config.go
```diff
@@ -447,6 +447,10 @@ type SealerConfig struct {
 
 	// PrefetchGasLimitPercent is the gas limit percentage for prefetching (e.g., 100 = 100%, 110 = 110%)
 	PrefetchGasLimitPercent uint64 `hcl:"prefetch-gaslimit-percent,optional" toml:"prefetch-gaslimit-percent,optional"`
+
+	// SlowTxThreshold warns if a single tx apply time exceeds this threshold. 0 = disabled.
+	SlowTxThreshold    time.Duration `hcl:"-,optional" toml:"-"`
```
cffls: Can we log based on their percentile, e.g. P99.99, instead of absolute values, e.g. 100ms? The absolute values might vary depending on the hardware of the running host, but percentile is pretty much agnostic to the environment.

lucca30: That's an interesting approach. To be honest, my only concern on this strategy is to significantly increase the log usage since we will going to log in all the blocks, instead of just logging the outliers.

lucca30:
@claude Do an analysis for the new changes I did. Checking if any bug, error or sec breach was introduced.

internal/cli/server/config.go
lucca30: Agreed. Approach implemented on the latest commits.

marcello33 (approved):

miner/worker.go
```diff
@@ -1193,11 +1232,28 @@ mainloop:
...
+			if !lastCommitStart.IsZero() {
+				logCtx = append(logCtx, "txHash", lastTxHash.Hex())
+				logCtx = append(logCtx, "txIndex", lastTxIndex)
+				logCtx = append(logCtx, "sender", lastTxSender)
+				logCtx = append(logCtx, "txElapsed", common.PrettyDuration(time.Since(lastCommitStart)))
+			}
 
 			if w.IsRunning() {
-				log.Info("Block building interrupted due to timeout, aborting new transaction commits", "number", env.header.Number.Uint64(), "hash", lastTxHash)
+				log.Info("Block building interrupted due to timeout, aborting new transaction commits", logCtx...)
```
manav2401, line 1255: Small suggestion: might be good to add header time to log here.

miner/slow_tx_tracker_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"math/big"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+func TestSlowTxTopTrackerKeepsTop10(t *testing.T) {
```
manav2401, line 14: Small test to add (or extend this one) would be to insert transactions in random order. Currently, durations are sorted so only limited behavior is checked.

miner/worker.go
lucca30: Nice, agreed. Working on it here

miner/slow_tx_tracker_test.go
lucca30: Nice! Implementing it here

miner/worker.go
lucca30: Done

miner/slow_tx_tracker_test.go
lucca30: Done

PR #2135, opened 2026-03-10
consensus, ethapi: embed gas target and base fee params in block header extra field (post Giugliano HF)

pratikspatil024:
# Description

- Extends `BlockExtraData` with two optional RLP fields (`GasTarget`, `BaseFeeChangeDenominator`) so full node operators and RPC users can determine the EIP-1559 parameters a block producer used
- Populates these fields during `Prepare()` for `post-Giugliano` blocks and verifies their presence during block import in `verifyCascadingFields()`
- Exports `eip1559.CalcGasTarget()` for use by consensus code
- Adds `bor_getBlockGasParams` RPC method returning both values for any block (post Giugliano block)
- Adds optional `borExtra` parameter to `eth_getBlockByNumber` and `eth_getBlockByHash`, when set to true, the response includes a `blockExtraData` object with parsed `gasTarget`, `baseFeeChangeDenominator`, and `txDependency` from the header extra field, eliminating the need for a second RPC call
- Adds `Header.DecodeBlockExtraData()` for single-pass RLP decoding of the full BlockExtraData struct
- Adds RLP backward compatibility and `GetBaseFeeParams` extraction tests

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/consensus/bor/bor_test.go
+++ b/consensus/bor/bor_test.go
@@ -4944,8 +4944,8 @@ func TestPrepare_GiuglianoExtraFields_SprintEnd(t *testing.T) {
 	gasTarget, bfcd := h.GetBaseFeeParams(cfg)
 	require.NotNil(t, gasTarget, "GasTarget should be present in sprint-end header")
 	require.NotNil(t, bfcd, "BaseFeeChangeDenominator should be present in sprint-end header")
-	require.True(t, *gasTarget > 0)
-	require.True(t, *bfcd > 0)
+	require.Equal(t, eip1559.CalcGasTarget(cfg, genesis), *gasTarget)
+	require.Equal(t, params.BaseFeeChangeDenominator(cfg.Bor, genesis.Number), *bfcd)
 }
 
 func TestPrepare_GiuglianoExtraFields_NonSprint(t *testing.T) {
@@ -4966,8 +4966,8 @@ func TestPrepare_GiuglianoExtraFields_NonSprint(t *testing.T) {
 	gasTarget, bfcd := h.GetBaseFeeParams(cfg)
 	require.NotNil(t, gasTarget, "GasTarget should be present in non-sprint header")
 	require.NotNil(t, bfcd, "BaseFeeChangeDenominator should be present in non-sprint header")
-	require.True(t, *gasTarget > 0)
-	require.True(t, *bfcd > 0)
+	require.Equal(t, eip1559.CalcGasTarget(cfg, genesis), *gasTarget)
+	require.Equal(t, params.BaseFeeChangeDenominator(cfg.Bor, genesis.Number), *bfcd)
 }
 
 func TestPrepare_PreGiugliano_NoExtraFields(t *testing.T) {


internal/ethapi/api.go
```diff
@@ -590,7 +590,10 @@ func (api *BlockChainAPI) GetHeaderByHash(ctx context.Context, hash common.Hash)
 //   - When number is -4 the chain safe block is returned.
 //   - When fullTx is true all transactions in the block are returned, otherwise
 //     only the transaction hash is returned.
-func (api *BlockChainAPI) GetBlockByNumber(ctx context.Context, number rpc.BlockNumber, fullTx bool) (map[string]interface{}, error) {
+//   - When borExtra is true, the response includes a "blockExtraData" object with
+//     parsed EIP-1559 gas parameters and transaction dependency metadata from the
+//     block header's Extra field. This parameter is optional.
+func (api *BlockChainAPI) GetBlockByNumber(ctx context.Context, number rpc.BlockNumber, fullTx bool, borExtra *bool) (map[string]interface{}, error) {
```
cffls, line 596: Why is borExtra a pointer while fullTx a bool? Can they both be in bool type? Also, is borExtra optional?

pratikspatil024: borExtra is a *bool pointer specifically to make it optional at the RPC level.

```
  // This works because borExtra is *bool (optional):
  {"method": "eth_getBlockByNumber", "params": ["0x1", true]}

  // This also works:
  {"method": "eth_getBlockByNumber", "params": ["0x1", true, true]}
```
If borExtra were a plain bool, the first call would fail with a missing parameter error.

cffls: Nice! Thanks for the explanation.

pratikspatil024:
@claude - review this PR

internal/ethapi/api.go
manav2401, line 596: Won't the header already have an `extra` field in the response which has the encoded extra data (which users can decode on their end)? Or is the goal to have a readable output to the end user instead of them doing the decoding?

internal/ethapi/bor_api.go
```diff
@@ -314,6 +314,92 @@ func (api *BorAPI) GetHeaderByNumber(ctx context.Context, blockNumber rpc.BlockN
...
+}
+
+// appendBorExtraData adds the parsed block extra data to the response map if borExtra is true.
+func appendBorExtraData(response map[string]interface{}, block *types.Block, borExtra *bool, chainConfig *params.ChainConfig) {
+	if borExtra == nil || !*borExtra || response == nil {
+		return
+	}
+
+	if extraData := marshalBlockExtraData(block.Header(), chainConfig); extraData != nil {
+		response["blockExtraData"] = extraData
```
manav2401: If the goal is to have readable output, we should rename this to something like `decodedExtra` as we already have an `extra` field in response.

consensus/bor/bor_test.go
```diff
@@ -4743,3 +4745,290 @@ func TestVerifyHeaderRejectsInvalidBlockNumber(t *testing.T) {
...
+	parent := &types.Header{Number: big.NewInt(0), GasLimit: 30_000_000}
+	bed := &types.BlockExtraData{}
+
+	b.setGiuglianoExtraFields(header, parent, bed)
+
+	require.Nil(t, bed.GasTarget, "GasTarget should be nil for pre-Giugliano blocks")
+	require.Nil(t, bed.BaseFeeChangeDenominator, "BaseFeeChangeDenominator should be nil for pre-Giugliano blocks")
+}
+
+func TestSetGiuglianoExtraFields_PostGiugliano(t *testing.T) {
```
manav2401, line 4891: We should also check if accurate values are populated instead of just checking if the value is > 0. 

We can compare with the values with response from `eip1559.CalcGasTarget` and `params.BaseFeeChangeDenominator`.

internal/ethapi/api.go
pratikspatil024: Yes, that is an additional work on the user's end, therefore we wanted to have it as a readable output.

consensus/bor/bor_test.go
pratikspatil024: done [here](https://github.com/0xPolygon/bor/pull/2135/commits/9ced830ef69d0efd82cfc03cd473d277393808e8), thanks!

internal/ethapi/bor_api.go
pratikspatil024: [renamed](https://github.com/0xPolygon/bor/pull/2135/commits/452214e4c61371bfb3fd5cac625bf104cda6bafa), thanks!

PR #2138, opened 2026-03-11
Use PGO when building binaries

cffls:
# Description

The pprof file was a merge from pprof taken from a production block producer and a sentry. This profile allows the go compiler to make more informed optimization decisions during compilation. For more info, see: https://go.dev/doc/pgo


lucca30 (approved):
lgtm

PR #2145, opened 2026-03-18
metrics, build: fix time.Tick resource leaks and oss-fuzz shebang

haoyu-haoyu:
## Summary

Two fixes inherited from upstream go-ethereum, both already accepted upstream.

### 1. metrics: time.Tick → time.NewTicker (6 files)

`time.Tick()` creates a ticker that can never be garbage collected (SA1015). Replace with `time.NewTicker()` + `defer Stop()`.

Files: `metrics/debug.go`, `metrics/json.go`, `metrics/log.go`, `metrics/opentsdb.go`, `metrics/syslog.go`, `metrics/writer.go`.

### 2. oss-fuzz.sh: fix shebang

Line 38 writes `#/bin/sh` instead of `#!/bin/sh` — missing `!` in the interpreter directive of generated fuzz scripts.

### Upstream status

Both fixes have been accepted in go-ethereum:
- ethereum/go-ethereum#34044 (shebang) — **merged** by @gballet
- ethereum/go-ethereum#34046 (time.Tick) — open, CI passing

### DIFF
--- a/metrics/debug.go
+++ b/metrics/debug.go
@@ -22,7 +22,9 @@ var (
 // CaptureDebugGCStats captures new values for the Go garbage collector statistics
 // exported in debug.GCStats. This is designed to be called as a goroutine.
 func CaptureDebugGCStats(r Registry, d time.Duration) {
-	for range time.Tick(d) {
+	ticker := time.NewTicker(d)
+	defer ticker.Stop()
+	for range ticker.C {
 		CaptureDebugGCStatsOnce(r)
 	}
 }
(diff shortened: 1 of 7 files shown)


manav2401 (approved):

manav2401:
Hi @haoyu-haoyu we'd like to be as close to upstream as possible to avoid conflicts when merging. In this case, we'll stick to using tick as after go 1.23, it can garbage collect unreferenced tickers. Can you undo that change? Thanks!

haoyu-haoyu:
@manav2401 Makes sense — thanks for explaining the Go 1.23 GC improvement! I've reverted all 6 metrics file changes. The PR now only contains the `oss-fuzz.sh` shebang fix (`#/bin/sh` → `#!/bin/sh`).

cffls (approved):

manav2401 (approved):

PR #2152, opened 2026-03-20
p2p: add flag to disable snap/1 serving

lucca30:
## Changes

Adds a new `p2p.nosnap` config flag (`nosnap` in TOML under `[p2p]`) that decouples the snapshot tree from the `snap/1` p2p sub-protocol.

**The problem:** before this change, `--snapshot` and `snap/1` were hardcoded together — enabling the snapshot tree automatically registered snap/1. There was no way to have one without the other.

**Why BPs want the snapshot tree:** the flat snapshot accelerates state reads without trie traversal, benefiting both RPC queries and block building (every `SLOAD`, balance check, and storage read during EVM execution can hit the snapshot instead of walking the trie).

**Why BPs don't want to serve `snap/1`:** BPs are already fully synced and have no reason to be snap sync servers for other nodes. Serving snap/1 adds unnecessary network and CPU overhead.

**Important clarification — `snap/1` vs `--snapshot`:**
- `snap/1` is the p2p sub-protocol used for snap sync state transfer between nodes. Data is served from the flat snapshot, Merkle proofs from the trie DB on disk
- `--snapshot` / `SnapshotCache` is the in-memory flat snapshot tree used for fast local state reads

These are independent. `nosnap=true` disables `snap/1` entirely (the node neither serves nor consumes snap sync), while the snapshot tree remains fully active for local performance.

> **Note:** reducing `TriesInMemory` (dirty trie layers in memory) is a separate and independent optimization — it saves RAM by committing trie nodes to disk sooner. It does not affect `snap/1` serving, which reads from the flat snapshot, not trie layers.

## How it works

- New `NoSnapServing bool` field in `ethconfig.Config`
- Wired from `P2PConfig.NoSnapServing` (`hcl:"nosnap"` / `toml:"nosnap"`) through `buildEth`
- In `Protocols()`, `snap/1` is only registered if `SnapshotCache > 0 && !NoSnapServing` — previously the condition was `SnapshotCache > 0` alone, making snapshot and `snap/1` inseparable

## Testing

- Unit test in `eth/backend_protocols_test.go` covering both the default (snap served) and `NoSnapServing=true` (snap not served) cases
- Validated end-to-end in a local kurtosis devnet: nodes with `nosnap = true` started without advertising `snap/1`

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -693,6 +693,11 @@ func (s *Ethereum) SetAuthorized(authorized bool) {
 // network protocols to start.
 func (s *Ethereum) Protocols() []p2p.Protocol {
 	protos := eth.MakeProtocols((*ethHandler)(s.handler), s.networkID, s.discmix)
+	// snap/1 is only registered when the snapshot cache is enabled and NoSnapServing is false.
+	// Setting NoSnapServing=true disables snap/1 entirely: the node will neither serve snap sync
+	// requests nor be able to sync state via snap/1 itself. The in-memory snapshot tree (used for
+	// fast local state reads) remains active regardless. Nodes that need to snap sync from peers
+	// must leave NoSnapServing=false (the default).
 	if s.config.SnapshotCache > 0 && !s.config.NoSnapServing {
 		protos = append(protos, snap.MakeProtocols((*snapHandler)(s.handler))...)
 	}
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -693,7 +693,7 @@ func (s *Ethereum) SetAuthorized(authorized bool) {
 // network protocols to start.
 func (s *Ethereum) Protocols() []p2p.Protocol {
 	protos := eth.MakeProtocols((*ethHandler)(s.handler), s.networkID, s.discmix)
-	if s.config.SnapshotCache > 0 {
+	if s.config.SnapshotCache > 0 && !s.config.NoSnapServing {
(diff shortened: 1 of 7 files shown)


eth/backend.go
```diff
@@ -693,7 +693,7 @@ func (s *Ethereum) SetAuthorized(authorized bool) {
 // network protocols to start.
 func (s *Ethereum) Protocols() []p2p.Protocol {
 	protos := eth.MakeProtocols((*ethHandler)(s.handler), s.networkID, s.discmix)
-	if s.config.SnapshotCache > 0 {
+	if s.config.SnapshotCache > 0 && !s.config.NoSnapServing {
```
manav2401, line 701: Won't this disable snap/1 completely from being registered? In which case the node won't be able to sync from snap. 

I think it's fine - but we can mention this fact in comment / description. 

Basically, if we want to sync from snap, we can set `nosnap` to false which will allow snap/1 to be registered.

lucca30: Good call, added a comment at that line. Yes, `nosnap=true` disables snap/1 entirely — the node won't serve snap sync requests and also won't be able to sync state via snap/1 from peers. The in-memory snapshot tree stays active regardless (it's just used for fast local state reads). If a node needs to snap sync, it should keep `nosnap=false` (the default).

manav2401 (approved):

marcello33 (approved):

PR #2153, opened 2026-03-20
core: fill blob receipt fields in parallel processor

LarryArnault45:
Fill BlobGasUsed and BlobGasPrice for blob receipts in the parallel state processor, and add a focused test covering ExecutionTask.Settle.

### DIFF
--- a/core/parallel_state_processor.go
+++ b/core/parallel_state_processor.go
@@ -237,6 +237,11 @@ func (task *ExecutionTask) Settle() {
 	receipt.TxHash = task.tx.Hash()
 	receipt.GasUsed = task.result.UsedGas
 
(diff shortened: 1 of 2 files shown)


lucca30:
Hi @LarryArnault45 , thanks for contributing to the codebase. But `BlobTxs` does not exists on Polygon POS, so there is no need on any handling regarding these txs types.

Is there any context of usage in which Filling this field was necessary? 

Thanks in advance for the contribution.

LarryArnault45:
Hi @lucca30, Thanks for the clarification. I opened this for serial/parallel parity, but I don't have a concrete Polygon PoS use case for it. If BlobTxs are not relevant on Polygon PoS, then this change is not needed here. I'll close the PR. Thanks for pointing it out.

PR #2157, opened 2026-03-23
eth, eth/protocols/eth: fix goroutine leaks in witness request path

cffls:
# Description
The parallel-stateless-import feature leaked goroutines in RequestWitnessesWithVerification when peers disconnected or stopped responding. Each stalled witness request leaked 2-4 goroutines that pinned witness data and channels in memory indefinitely.

Three root causes:

1. doWitnessRequest spawned a goroutine that blocked forever on the per-request response channel (<-witResCh) with no timeout or cancellation. When a peer disconnected, the wit dispatcher exited without signaling pending sink channels, so the goroutine never returned. This cascaded: the WaitGroup never completed, witReqResCh never closed, and the adapter goroutine also leaked.

2. eth.Request.Close() returned early without closing the Cancel channel when peer was nil. The witness adapter returns a shim eth.Request (peer == nil) to the concurrent fetcher. When the fetcher called Close() on timeout or disconnect, Cancel was never closed, so the adapter goroutine's select on Cancel never fired.

3. The adapter spawned fire-and-forget goroutines (go func(){<-doneCh}()) to drain the Done channel on synthetic eth.Response objects. Nothing ever wrote to these channels, so the goroutines leaked unconditionally.

Fixes:

- eth.Request.Close() now always closes Cancel even when peer is nil.
- doWitnessRequest selects on a cancel channel in both receive and forward paths, with proper WaitGroup/semaphore cleanup on exit.
- Replace doneCh drainer goroutines with buffered channels.
- Adapter defers closing underlying wit.Request objects on exit.

# Changes

# Breaking changes

# Nodes audience

# Checklist

# Cross repository changes

    - In case link the PR here:
    - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/eth/peer_test.go
+++ b/eth/peer_test.go
@@ -35,20 +35,12 @@ func TestRequestWitnesses_NoWitPeer(t *testing.T) {
 }
 
 func TestRequestWitnesses_HasWitPeer_Returns(t *testing.T) {
-	ctrl := gomock.NewController(t)
-	defer ctrl.Finish()
-
 	hashToRequest := common.Hash{123}
-	witness, _ := stateless.NewWitness(&types.Header{}, nil)
-	FillWitnessWithDeterministicRandomState(witness, 10*1024)
-	var witBuf bytes.Buffer
-	witness.EncodeRLP(&witBuf)
+	witData := testWitnessData(t)
 
-	mockWitPeer := NewMockWitnessPeer(ctrl)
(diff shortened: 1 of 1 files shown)


cffls:
@claude review

cffls:
> LGTM - I would test that a node can do a graceful shutdown. We've faced issues with go routines blocking bor from doing a graceful shutdown.

Thanks - tested multiple times locally and the node was able to shutdown gracefully.

Issue #1930, opened 2025-12-12
The new log indexer doesn't include state-sync logs before the Madhugiri fork


ngotchac:
#### **System information**

Bor client version: [v2.5.3](https://github.com/0xPolygon/bor/releases/tag/v2.5.3)

OS & Version: Linux

Environment: Polygon Mainnet

#### **Overview of the problem**

We have activated the new log indexer, available since go-ethereum [v1.15.6](https://github.com/ethereum/go-ethereum/releases/tag/v1.15.6) via `--history.logs=0`

It works great, however it doesn't include logs from state-sync transactions before the latest Madhugiri fork (in which they are properly part of the block's transactions).

#### **Reproduction Steps**

- See this state-sync tx from before the fork: https://polygonscan.com/tx/0xe26ba120985511555661bcada23f3c8e67bc6e0b5bdb8a71d4f9641c41f1fd2b

```bash
cast logs --from-block 80006032 --to-block 80006032 "Transfer(address indexed,address indexed,uint256)" 0x0000000000000000000000000000000000000000 0x6fE243a76fD51Aa2f3Cc0E731Da4cC0d877E556d
```

It returns nothing, so it doesn't include the mint to 0x6fE243a76fD51Aa2f3Cc0E731Da4cC0d877E556d

- See this state-sync tx after the fork: https://polygonscan.com/tx/0x19266376c21947637ceefc22fc7b0547fd64e9dfa95dc65a6cd430d2ae4b15a7

```bash
cast logs --from-block 80194432 --to-block 80194432 "Transfer(address indexed,address indexed,uint256)" 0x0000000000000000000000000000000000000000 0x6fE243a76fD51Aa2f3Cc0E731Da4cC0d877E556d

- address: 0xdd28ec6b06983d01D37DbD9Ab581d8d884d95264
  blockHash: 0x01f63befb76da3cb4053d03c7c350bc70b2a8ef70e126866f3a8337e5307f53c
  blockNumber: 80194432
  data: 0x000000000000000000000000000000000000000000000195413e63a9379fd684
  logIndex: 1575
  removed: false
  topics: [
  	0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef
  	0x0000000000000000000000000000000000000000000000000000000000000000
  	0x0000000000000000000000006fe243a76fd51aa2f3cc0e731da4cc0d877e556d
  ]
[...]

PR #2155, opened 2026-03-23
ethapi: include pre-Madhugiri state-sync logs in bor_getLogs

LarryArnault45:
Include pre-Madhugiri Bor state-sync receipts when collecting logs for bor_getLogs and bor_getLatestLogs, and add regression coverage for the historical state-sync log path.

PR #2159, opened 2026-03-25
eth/filters: restore pre-Madhugiri state-sync logs in historical queries

peter941221:
## Suggested PR Title

`eth/filters: restore pre-Madhugiri state-sync logs in historical queries`

## Summary

Fixes `#1930` by restoring pre-Madhugiri state-sync logs in the standard historical log APIs.

- auto-merge Bor sidecar logs for the eligible pre-Madhugiri slice in `eth_getLogs`
- do the same for `eth_getFilterLogs`
- resolve special block tags before making the fork-boundary decision
- keep the canonical filter path authoritative for public RPC errors and post-fork behavior

## Why This Shape

The new log indexer answers historical queries from canonical receipts. That works after Madhugiri, but pre-fork state-sync logs still live in Bor sidecar receipts, so canonical-only historical queries can miss them.

This patch keeps the normal filter path as the first and authoritative step, then merges Bor sidecar logs only when the request actually intersects the pre-Madhugiri slice. That restores historical compatibility without widening `filtermaps`, live subscriptions, or Bor-only RPC behavior.

## What Changed

- added helper logic in `eth/filters/api.go` to build Bor sidecar filters for eligible historical queries
- resolved `latest`, `finalized`, `safe`, and `earliest` before deciding whether a range intersects pre-Madhugiri history
- limited the compatibility merge to blocks strictly before `MadhugiriBlock`
- reused the same merge path for both `GetLogs` and `GetFilterLogs`

## Tests

- `GetLogs` pre-fork `blockHash` query auto-merges Bor sidecar logs
- missing-sidecar pre-fork queries remain a clean no-op
- post-Madhugiri `GetLogs` stays canonical-only, including a range ending at `latest`
- `GetFilterLogs` gets the same pre-fork and post-fork coverage
- broader `./eth/filters` and `./tests/bor` suites pass locally

## Notes

- This restores compatibility when the relevant Bor sidecar data is present.
- Nodes that are missing historical state-sync sidecar data may still need `backfill-statesync-txs` for full completeness.

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -481,20 +481,13 @@ func (api *FilterAPI) borLogsFilterForHistoricalQuery(ctx context.Context, block
 }
 
 func (api *FilterAPI) borLogsFilterForBlock(ctx context.Context, borConfig *params.BorConfig, blockHash common.Hash, addresses []common.Address, topics [][]common.Hash) *BorBlockLogsFilter {
-	if !api.borLogs {
-		if borConfig == nil || borConfig.MadhugiriBlock == nil {
-			return nil
-		}
-
+	if borConfig != nil && borConfig.MadhugiriBlock != nil {
 		header, err := api.sys.backend.HeaderByHash(ctx, blockHash)
 		if err != nil || header == nil || borConfig.IsMadhugiri(header.Number) {
 			return nil
 		}
-	} else if borConfig != nil && borConfig.MadhugiriBlock != nil {
-		header, err := api.sys.backend.HeaderByHash(ctx, blockHash)
-		if err != nil || header == nil || borConfig.IsMadhugiri(header.Number) {
-			return nil
-		}
+	} else if !api.borLogs {
+		return nil
 	}
 
 	return NewBorBlockLogsFilter(api.sys.backend, borConfig, blockHash, addresses, topics)
@@ -531,25 +524,21 @@ func (api *FilterAPI) borLogsFilterForRange(ctx context.Context, borConfig *para
 // closely enough for Bor sidecar compatibility decisions, while keeping the
 // canonical filter path authoritative for public RPC errors.
 func (api *FilterAPI) resolveHistoricalLogBlockNumber(ctx context.Context, number int64) (uint64, bool) {
+	resolveHeaderNumber := func(number rpc.BlockNumber) (uint64, bool) {
+		header, err := api.sys.backend.HeaderByNumber(ctx, number)
+		if err != nil || header == nil {
+			return 0, false
+		}
+		return header.Number.Uint64(), true
+	}
+
 	switch number {
 	case rpc.LatestBlockNumber.Int64():
-		header, err := api.sys.backend.HeaderByNumber(ctx, rpc.LatestBlockNumber)
-		if err != nil || header == nil {
-			return 0, false
-		}
-		return header.Number.Uint64(), true
+		return resolveHeaderNumber(rpc.LatestBlockNumber)
 	case rpc.FinalizedBlockNumber.Int64():
-		header, err := api.sys.backend.HeaderByNumber(ctx, rpc.FinalizedBlockNumber)
-		if err != nil || header == nil {
-			return 0, false
-		}
-		return header.Number.Uint64(), true
+		return resolveHeaderNumber(rpc.FinalizedBlockNumber)
 	case rpc.SafeBlockNumber.Int64():
-		header, err := api.sys.backend.HeaderByNumber(ctx, rpc.SafeBlockNumber)
-		if err != nil || header == nil {
-			return 0, false
-		}
-		return header.Number.Uint64(), true
+		return resolveHeaderNumber(rpc.SafeBlockNumber)
 	case rpc.EarliestBlockNumber.Int64():
 		return api.sys.backend.HistoryPruningCutoff(), true
 	default:
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -471,13 +471,100 @@ func (api *FilterAPI) NewFilter(crit FilterCriteria) (rpc.ID, error) {
 	return logsSub.ID, nil
 }
 
+func (api *FilterAPI) borLogsFilterForHistoricalQuery(ctx context.Context, blockHash *common.Hash, begin, end int64, addresses []common.Address, topics [][]common.Hash) *BorBlockLogsFilter {
+	borConfig := api.sys.backend.ChainConfig().Bor
+	if blockHash != nil {
+		return api.borLogsFilterForBlock(ctx, borConfig, *blockHash, addresses, topics)
+	}
+
+	return api.borLogsFilterForRange(ctx, borConfig, begin, end, addresses, topics)
+}
+
+func (api *FilterAPI) borLogsFilterForBlock(ctx context.Context, borConfig *params.BorConfig, blockHash common.Hash, addresses []common.Address, topics [][]common.Hash) *BorBlockLogsFilter {
+	if !api.borLogs {
+		if borConfig == nil || borConfig.MadhugiriBlock == nil {
+			return nil
+		}
+
+		header, err := api.sys.backend.HeaderByHash(ctx, blockHash)
+		if err != nil || header == nil || borConfig.IsMadhugiri(header.Number) {
+			return nil
+		}
+	} else if borConfig != nil && borConfig.MadhugiriBlock != nil {
+		header, err := api.sys.backend.HeaderByHash(ctx, blockHash)
+		if err != nil || header == nil || borConfig.IsMadhugiri(header.Number) {
+			return nil
+		}
+	}
+
+	return NewBorBlockLogsFilter(api.sys.backend, borConfig, blockHash, addresses, topics)
+}
+
+func (api *FilterAPI) borLogsFilterForRange(ctx context.Context, borConfig *params.BorConfig, begin, end int64, addresses []common.Address, topics [][]common.Hash) *BorBlockLogsFilter {
+	if !api.borLogs && (borConfig == nil || borConfig.MadhugiriBlock == nil) {
+		return nil
+	}
+
+	resolvedBegin, ok := api.resolveHistoricalLogBlockNumber(ctx, begin)
+	if !ok {
+		return nil
+	}
+	resolvedEnd, ok := api.resolveHistoricalLogBlockNumber(ctx, end)
+	if !ok || resolvedBegin > resolvedEnd {
+		return nil
+	}
+
+	if borConfig != nil && borConfig.MadhugiriBlock != nil {
+		madhugiriBlock := borConfig.MadhugiriBlock.Uint64()
+		if resolvedBegin >= madhugiriBlock {
+			return nil
+		}
+		if resolvedEnd >= madhugiriBlock {
+			resolvedEnd = madhugiriBlock - 1
+		}
+	}
+
+	return NewBorBlockLogsRangeFilter(api.sys.backend, borConfig, int64(resolvedBegin), int64(resolvedEnd), addresses, topics)
+}
+
+// resolveHistoricalLogBlockNumber mirrors the standard filter's block-tag handling
+// closely enough for Bor sidecar compatibility decisions, while keeping the
+// canonical filter path authoritative for public RPC errors.
+func (api *FilterAPI) resolveHistoricalLogBlockNumber(ctx context.Context, number int64) (uint64, bool) {
+	switch number {
+	case rpc.LatestBlockNumber.Int64():
+		header, err := api.sys.backend.HeaderByNumber(ctx, rpc.LatestBlockNumber)
+		if err != nil || header == nil {
+			return 0, false
+		}
+		return header.Number.Uint64(), true
+	case rpc.FinalizedBlockNumber.Int64():
+		header, err := api.sys.backend.HeaderByNumber(ctx, rpc.FinalizedBlockNumber)
+		if err != nil || header == nil {
+			return 0, false
+		}
+		return header.Number.Uint64(), true
+	case rpc.SafeBlockNumber.Int64():
+		header, err := api.sys.backend.HeaderByNumber(ctx, rpc.SafeBlockNumber)
+		if err != nil || header == nil {
+			return 0, false
+		}
+		return header.Number.Uint64(), true
+	case rpc.EarliestBlockNumber.Int64():
+		return api.sys.backend.HistoryPruningCutoff(), true
+	default:
+		if number < 0 {
+			return 0, false
+		}
+		return uint64(number), true
+	}
+}
+
 // GetLogs returns logs matching the given argument that are stored within the state.
 func (api *FilterAPI) GetLogs(ctx context.Context, crit FilterCriteria) ([]*types.Log, error) {
 	if len(crit.Topics) > maxTopics {
 		return nil, errExceedMaxTopics
 	}
-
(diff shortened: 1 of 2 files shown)


peter941221:
Latest Sonar rerun is green on the current head (`e4b0a77`).

The remaining duplication was in the historical Bor test harness setup, so the follow-up only rewrote that fixture persistence flow into a dedicated helper. No RPC behavior changed.

Re-ran locally:
- `go test ./eth/filters -run 'TestHistoricalQueries(AutoMergePreMadhugiriBorLogs|KeepPostMadhugiriResultsCanonical)' -count=1`
- `go test ./eth/filters -count=1`
- `go test ./tests/bor -count=1`

peter941221:
Quick follow-up here because `#1930` is still open and this PR has been quiet for a few days.

The current head is still `e4b0a77`, Sonar is green there, and the last local reruns on that head were:
- `go test ./eth/filters -run 'TestHistoricalQueries(AutoMergePreMadhugiriBorLogs|KeepPostMadhugiriResultsCanonical)' -count=1`
- `go test ./eth/filters -count=1`
- `go test ./tests/bor -count=1`

One reviewer-orientation note in case the overlap with `#2155` is the sticking point:
- `#2155` restores the Bor-specific `bor_getLogs` path
- this PR restores the standard historical `eth_getLogs` / `eth_getFilterLogs` path, while keeping the canonical filter path authoritative for errors and post-fork behavior

@manav2401 if you'd prefer this split or rebased differently, I'm happy to adjust. I mainly want the `#1930` fix to land in the maintainers' preferred shape.

manav2401 (commented):
The core logic LGTM - is it possible to simplify the tests a bit? It does a lot of explicit handling for state-sync transactions which might be avoided. 

You can refer to `bor_filter_test.go` which uses go mocks and can extend those tests to also test the `GetLogs` and `GetFilterLogs` api for state-sync transactions. Or maybe you can simplify the current test to just use everything from DB instead of setting up a harness to deliver data from maps.

peter941221:
Thanks, that's helpful.

I took the DB-backed route here so the `GetLogs` / `GetFilterLogs` coverage stays in place, but the extra map-backed harness is gone. The core logic is unchanged; this only simplifies the historical test setup around the pre-/post-Madhugiri cases.

Re-ran locally:
- `go test ./eth/filters -run 'TestHistoricalQueries(AutoMergePreMadhugiriBorLogs|KeepPostMadhugiriResultsCanonical)' -count=1`
- `go test ./eth/filters -count=1`
- `go test ./tests/bor -count=1`

manav2401:
Nice, thank you very much.

manav2401 (approved):

peter941221:
Thanks — I dug into the remaining Codecov gap and closed it with a follow-up test-only commit:

- `2674bc3` — `eth/filters: cover historical bor helper guards`

This keeps `eth/filters/api.go` unchanged and only adds helper-focused regression coverage in `eth/filters/filter_system_test.go` for the historical-query decision paths that were still missing hits locally:

- `borLogsFilterForBlock(...)` early returns when Bor support is disabled or the target header is missing
- `borLogsFilterForRange(...)` early returns for unresolved block tags, inverted ranges, missing `MadhugiriBlock`, and post-fork-only ranges
- `resolveHistoricalLogBlockNumber(...)` handling for `latest`, `finalized`, `safe`, `earliest`, and negative block numbers

Local reruns on top of that commit:

- `go test ./eth/filters -run 'Test(HistoricalQueries|HistoricalQueryHelperGuards|ResolveHistoricalLogBlockNumber)' -count=1`
- `go test ./eth/filters -coverprofile=coverage_filters -count=1`
- `go test ./eth/filters -count=1`
- `go test ./tests/bor -count=1`

I also rechecked the added executable lines in `eth/filters/api.go` against the PR base (`0d6ee4b`); they are all hit locally now.

If you'd prefer this packaged differently, happy to adjust, but I kept it test-only so the runtime fix stays exactly as previously reviewed.

ngotchac:
Have you tried the performances of this PR on pre-Madhugiri block ranges? The new indexer is blazing fast, and works really well. I'm afraid not using it for older blocks isn't the best fix.

Couldn't we backfill the indexer in the background when pre-Madhugiri state sync txs?

cffls:
Could you rebase from `develop` branch? There is recently a block range limit check we added to log filter, which should be added in this PR as well. See this for reference: https://github.com/0xPolygon/bor/pull/2149

peter941221:
@cffls Yes — I can do that.

I'll rebase this branch onto the latest `develop`, carry over the block-range limit guard from `#2149` into this PR as well, and rerun the relevant `eth/filters` and `tests/bor` coverage on top of the rebased head.

peter941221:
@ngotchac Good point. I optimized this patch first for correctness and the smallest behavior change in the historical `eth_getLogs` / `eth_getFilterLogs` path, not for proving the best long-term performance shape, and I have not benchmarked representative pre-Madhugiri ranges yet.

I'll rebase first so this is evaluated against current `develop`, then I'll post numbers for representative pre-Madhugiri block ranges.

My hesitation with folding a background backfill/indexer change into this same PR is that it widens the rollout surface quite a bit compared with this targeted compatibility fix. That said, if the performance delta is materially bad, or if maintainers would prefer the indexer/backfill direction for `#1930`, I'm happy to reshape the PR that way rather than force the current shape.

peter941221:
Follow-up: I rebased this branch onto the current `develop` and force-pushed the updated head (`bdd5c78`).

That brings in the block-range limit guard from `#2149` through the rebase base, and I also dropped the now-obsolete loop-variable copies in the historical tests so the rebased branch is clean with current lint rules.

Local reruns on top of `bdd5c78`:
- `go test ./eth/filters -run 'Test(HistoricalQueries|HistoricalQueryHelperGuards|ResolveHistoricalLogBlockNumber|RangeLimit)' -count=1`
- `go test ./eth/filters -count=1`
- `go test ./tests/bor -count=1`
- `golangci-lint run --config ./.golangci.yml --new-from-rev origin/develop ./eth/filters/...`

I still owe the pre-Madhugiri range performance numbers and will post those separately, but the branch is now aligned with current `develop` for review.

peter941221:
Thanks — I ran representative local benchmarks on the rebased head (`bdd5c78`) for historical `eth_getLogs` ranges.

Setup:

- synthetic in-memory `eth/filters` benchmark on the rebased branch
- same `address/topic` density on both sides of the fork
- pre-fork matches only in Bor sidecar receipts
- post-fork matches only in canonical receipts
- sprint length fixed at `16` so the pre-fork sidecar path exercises the same sprint-step scan logic this PR uses

Commands:

- `go test ./eth/filters -run '^$' -bench 'BenchmarkHistoricalBorGetLogs' -benchmem -count=1`
- `go test ./eth/filters -run '^$' -bench 'BenchmarkHistoricalBorGetLogs/span_16384' -benchmem -count=1`

Observed timings on my machine:

| Range shape | First run | Recheck |
| --- | --- | --- |
| pre-fork `4096` blocks | `12.1 ms/op` | - |
| post-fork `4096` blocks | `13.3 ms/op` | - |
| cross-fork `8192` blocks | `26.7 ms/op` | - |
| pre-fork `16384` blocks | `45.3 ms/op` | `54.8 ms/op` |
| post-fork `16384` blocks | `53.6 ms/op` | `56.0 ms/op` |
| cross-fork `32768` blocks | `101.4 ms/op` | `114.5 ms/op` |

So in this synthetic historical-range setup, the pre-Madhugiri compatibility path was not slower than the post-fork canonical path at the same range width, and the cross-fork case scaled roughly like the sum of the two halves rather than falling off a cliff.

This is not a replacement for real node / mainnet benchmarking, but it does suggest the current compatibility fix is not obviously introducing a severe regression on representative pre-fork ranges.

If maintainers still prefer the indexer/backfill direction as the long-term shape for `#1930`, I am happy to explore that separately. I kept the PR branch itself unchanged for this measurement pass so the runtime diff under review stays the same.

ngotchac:
A real benchmark should really be run for this IMO, and double checked by a human BTW  (as for the answers) ;)

One should setup a bor node with complete indexing of logs, and run a low cardinality `eth_getLogs` query on the full blockchain range.

peter941221:
Thanks, that's a fair ask.

I don't currently have a fully indexed mainnet Bor node available in this environment, so I don't want to oversell the local synthetic numbers as a substitute for the real full-range RPC benchmark you described.

What I did rerun locally on the rebased head was the historical `eth_getLogs` benchmark in `eth/filters` with equal `address/topic` density on both sides of the fork, plus `-count=5` reruns to tighten variance:

| Range shape | Pre-fork median | Post-fork median | Cross-fork median |
| --- | ---: | ---: | ---: |
| `4096`-block span | `11.16 ms/op` | `13.89 ms/op` | `26.37 ms/op` |
| `16384`-block span | `44.88 ms/op` | `47.11 ms/op` | `94.13 ms/op` |

So in the runtime path under review, the pre-Madhugiri compatibility path is staying in the same latency band as the post-fork canonical path, and the cross-fork case remains close to additive rather than showing an obvious cliff.

For the real-node check you suggested, this is the exact full-range low-cardinality `eth_getLogs` payload I would use for the `#1930` reproducer:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "eth_getLogs",
  "params": [
    {
      "fromBlock": "0x0",
      "toBlock": "latest",
      "topics": [
        "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
        "0x0000000000000000000000000000000000000000000000000000000000000000",
        "0x0000000000000000000000006fe243a76fd51aa2f3cc0e731da4cc0d877e556d"
      ]
    }
  ]
}
```

On a fully indexed mainnet Bor node (`--history.logs=0`, `--rpc.rangelimit=0`), my expectation would be:

- `develop`: still misses the pre-fork hit at block `80006032`
- this branch: returns the pre-fork hit at block `80006032` and the known post-fork hit at block `80194432`

If someone on the maintainer side already has such a node handy, that should make the apples-to-apples RPC check straightforward to run against `develop` vs this branch.

ngotchac:
Just for information, with the current full indexer, this query:
```
cast logs --from-block 41000000 --to-block 81000000 --address 0xa1428174F516F527fafdD146b883bB4428682737 "Transfer(address indexed,address indexed,uint256)" 0x0000000000000000000000000000000000000000 0x6fE243a76fD51Aa2f3Cc0E731Da4cC0d877E556d
```
takes 1.5s to run, from a laptop connected via WiFi. And that's to scan 40M blocks. But it misses some events (cf. original issue).

I think it would be a good query to test against

peter941221:
Thanks, this is a very helpful regression case.

I don't currently have enough disk budget in this environment to stand up a fully indexed mainnet Bor node locally, so I don't want to overclaim beyond the unit/integration coverage and the local synthetic benchmarks already posted here.

Since you already have a suitable indexed setup for this query, would you be open to checking the same query against this branch as an A/B comparison with `develop`?

The expected difference should be:

- `develop`: still misses the pre-Madhugiri state-sync hit around block `80006032`
- this branch: restores the pre-fork hit around block `80006032` while preserving the post-fork hit around block `80194432`

If helpful, I can also add a short validation note to the PR body with the exact command and expected result shape to make the comparison easier.

manav2401 (approved):

ngotchac:
@peter941221 I won't have any capacity to test this until end of next-week, unfortunately... @manav2401 has anyone from the team tested the feature?

peter941221:
Thanks — I dug into the current red checks on top of `bdd5c78`, and they do not look like branch-specific regressions in the historical-log patch itself.

What the failing logs currently show:

- both `e2e-tests` jobs fail at `google-github-actions/auth` before the PR logic is exercised
  - the error is that GitHub did not inject `ACTIONS_ID_TOKEN_REQUEST_TOKEN` / `ACTIONS_ID_TOKEN_REQUEST_URL`
  - the action text itself says this most likely means workflow-permission / fork-context limitations
- the failing `codecov` job successfully downloads both coverage artifacts and finds both `cover.out` files, but then the upload fails with Codecov `429` rate limiting and asks for a repository upload token

So from the current logs, the red checks look more like fork / infrastructure blockers than evidence that this branch is failing its own runtime logic.

If helpful, a maintainer-context rerun should give a much cleaner signal. If a rerun then points to a branch-specific failure, I am happy to dig in further right away.

ngotchac:
Nobody from the Polygon team can test this? /cc @manav2401 @cffls

manav2401:
Apologies for the delay - I'll take a look into checking this on a devnode. @peter941221 meanwhile can you rebase with latest develop? Hoping to get this merged asap.

peter941221:
Rebased this branch onto the latest develop and force-pushed the updated head (c2f11d86a). I reran go test ./eth/filters -run Test(HistoricalQueries|HistoricalQueryHelperGuards|ResolveHistoricalLogBlockNumber|RangeLimit) -count=1 and go test ./eth/filters -count=1 on top of the rebased head, and both passed locally. I also checked go test ./tests/bor -count=1, but that currently fails in my Windows environment on both this branch and current develop with github.com/0xPolygon/cosmos-sdk x/auth/signing verify.go:107: undefined: ethCrypto.RecoverPubkey, so it does not look branch-specific. Please take a look when you get a chance, and I am happy to adjust further if anything shows up in the devnode check.

peter941221:
Quick note on the current red govulncheck: I checked the same workflow on current develop as well, and the latest develop run is also failing with the same Go 1.26.2 / standard-library vulnerability findings, so this does not look branch-specific to this PR. I will keep an eye on the remaining runs here too.

peter941221:
One more note on the current unit-tests red: the failing test is TestLockOrdering_ReplacePendingNoDeadlock in core/txpool/legacypool/legacypool2_test.go, which does not overlap with this PR's eth/filters change surface. That makes it look more like an unrelated flaky or repo-wide failure than a direct regression from this patch, so a rerun would probably be the right first check.

ngotchac:
I actually solved my issue described in https://github.com/0xPolygon/bor/issues/1930 by adding the `--bor.logs` flag. It doesn't seem to add any delay to logs fetching, and works really well! I think this PR could be closed, but maybe document this flag a bit better (or always enable it?)

peter941221:
I think this comes down to which contract the team wants for the standard log APIs.

1. The original repro in #1930 uses the standard `eth_getLogs` path with `--history.logs=0`, and on `develop` that path misses the pre-Madhugiri state-sync hit at block `80006032`.

2. `--bor.logs` is opt-in today, not the default. `docs/cli/default_config.toml` still has `bor.logs = false`.

3. Bor already has a separate Bor-side historical log path in `eth/filters/bor_api.go` via `GetBorBlockLogs(...)`. This PR does not change that path.

4. This PR only changes the standard historical filter path in `eth/filters/api.go`, by merging Bor sidecar logs into `GetLogs(...)` / `GetFilterLogs(...)` when the query intersects the pre-Madhugiri slice.

5. So I do not think "the issue goes away when `--bor.logs` is enabled" proves this PR is obsolete. It shows that the opt-in Bor sidecar path works.

The remaining maintainer choice looks pretty narrow:

1. If standard `eth_getLogs` should stay complete across the Madhugiri boundary by default, this PR still fixes a real gap.

2. If the project wants that compatibility to require `--bor.logs`, then I agree the better fix is to document that flag more clearly, or possibly enable it by default, instead of carrying this patch.

peter941221:
@ngotchac Please take a look, thanks.

ngotchac:
@peter941221 I'm not a Bor maintainer :sweat_smile:

peter941221:
@ngotchac sorry about that.

PR #2170, opened 2026-03-30
ethdb, triedb: tune pebble write path and add safe pathdb state carry-over

marcello33:
# Description

This PR implements some PebbleDB optimizations.
Others are on a config level, and will be managed in pos-ops.
The suggested configs are (tested on a 128GB RAM machine)
```
  [cache]
    cache = 24576 # 3x the original value
    gc = 5
    gomemlimit = "55GB"
    snapshot = 25
    database = 45
    trie = 25
```

The PR combines write-path tuning with pathdb buffer changes aimed at reducing write stalls and improving dirty-state reuse across flushes.
On the Pebble side, it tunes sync and compaction behavior. On the pathdb side, it splits the write buffer into node/state budgets, allows state carry-over when flushes are caused by trie-node pressure, and keeps that carry-over safe under async flushing via copy-on-write state snapshots. The copy-on-write carry-over means:
- top-level state maps are copied
- per-account storage submaps are shared initially
- a storage submap is cloned only on first write
This keeps the frozen flush snapshot immutable while still avoiding eager duplication of all storage submaps.

Pebble tuning:
- increase `BytesPerSync` from the implicit default to `1 MiB`
- increase `MemTableStopWritesThreshold` to reduce write stalls during flushes
- enable adaptive compaction settings:
    - `L0CompactionConcurrency = 1`
    - `CompactionDebtConcurrency = 256 MiB`

Pathdb buffer changes:
- raise `maxBufferSize` to `2 GiB`
- add `StateReservation` to split the write buffer between:
    - trie nodes
    - flat state (accounts/storage)
- update buffer fullness logic:
    - flush when trie nodes exceed their reservation
    - also flush when total buffer usage exceeds the hard limit
- carry state forward only when:
    - the flush was caused by node pressure
    - state is still within its reserved budget
- wire the new buffer constructor/config through generation and journal loading
- clear carried-over state on revert when there are no uncommitted transitions left

# Changes

# Checklist

## Testing

- [x] I have added unit tests
- [ ] I have tested this code manually on mainnet

### DIFF
--- a/ethdb/pebble/pebble.go
+++ b/ethdb/pebble/pebble.go
@@ -328,8 +328,7 @@ func New(file string, cache int, handles int, namespace string, readonly bool) (
 			{TargetFileSize: 16 * 1024 * 1024, FilterPolicy: bloom.FilterPolicy(10)},
 			{TargetFileSize: 32 * 1024 * 1024, FilterPolicy: bloom.FilterPolicy(10)},
 			{TargetFileSize: 64 * 1024 * 1024, FilterPolicy: bloom.FilterPolicy(10)},
-			// Remove the bloom filter from L6, as the bottom level is always checked anyway
-			{TargetFileSize: 128 * 1024 * 1024},
+			{TargetFileSize: 128 * 1024 * 1024, FilterPolicy: bloom.FilterPolicy(10)},
 		},
 		ReadOnly: readonly,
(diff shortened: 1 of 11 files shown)


marcello33:
@claude review

cffls (approved):

pratikspatil024 (approved):

PR #2174, opened 2026-03-31
metrics: fix ResettingSample Prometheus _count monotonicity

### LINKED ISSUE #2173: Prometheus metrics violate counter monotonicity convention: _count and _sum reset on every scrape due to ResettingSample
#### **System information**

  Bor client version: v2.7.0 (also affects latest main)

  OS & Version: Linux (Kubernetes)

  Environment: Polygon Mainnet

  Type of node: Sentry

  #### **Overview of the problem**

  All Prometheus histogram metrics using `ResettingSample` have broken `_count` and `_sum` values. They decrease on every scrape instead of monotonically increasing, which violates the Prometheus counter type spec and breaks `rate()`,
  `increase()`, and average latency calculations.

https://prometheus.io/docs/concepts/metric_types/#counter

> A counter is a cumulative metric that represents a single [monotonically increasing counter ](https://en.wikipedia.org/wiki/Monotonic_function) whose value can only increase or be reset to zero on restart.

  **Root cause**: `ResettingSample.Snapshot()` in `metrics/resetting_sample.go` calls `Clear()` which resets count and sum to 0. The Prometheus collector in `metrics/prometheus/collector.go` emits these reset values as `counter` type,
  but they only contain the delta since the last scrape — not cumulative totals.

  **Affected files and metrics**:

  | File | Affected metrics |
  |---|---|
  | `rpc/metrics.go` | All `rpc_duration_*_count`, `rpc_duration_*_sum` |
  | `p2p/tracker/tracker.go` | P2P tracking metrics |
  | `eth/protocols/eth/handler.go` | eth protocol metrics |
  | `eth/protocols/snap/handler.go` | snap sync metrics |
  | `eth/protocols/wit/handler.go` | witness protocol metrics |

  **Expected**: `_count` and `_sum` should be monotonically increasing as required by the Prometheus counter type spec.

  **Actual**: Both reset to interval-only values on every scrape, causing:
  - `rate()` returns incorrect results or no data
  - `increase()` is unreliable
  - Average latency calculation (`rate(_sum) / rate(_count)`) is broken

  #### **Reproduction Steps**

  1. Enable telemetry with `metrics = true` and `prometheus-addr = "0.0.0.0:7071"`
  2. Send RPC requests (e.g. `eth_blockNumber`)
  3. Scrape `/debug/metrics/prometheus` twice with 1 minute interval
  4. Observe `rpc_duration_eth_blockNumber_success_count` and `_sum` values decrease on second scrape

  #### **Logs / Traces / Output / Error Messages**

  Scrape at T1 (100 requests since startup)

  rpc_duration_eth_blockNumber_success_count 100
  rpc_duration_eth_blockNumber_success_sum 5000000

  Scrape at T2 (50 requests since T1)

  rpc_duration_eth_blockNumber_success_count 50     ← decreased
  rpc_duration_eth_blockNumber_success_sum 2500000   ← decreased

  **Code path**:

  1. `rpc/metrics.go:46` — creates histogram with `ResettingSample`
  2. `metrics/resetting_sample.go:22` — `Snapshot()` calls `Clear()`
  3. `metrics/sample.go:165` — `Clear()` resets `count = 0`, `sum = 0`
  4. `metrics/prometheus/collector.go:98-107` — emits snapshot values as Prometheus `counter`

lake-dunamu:
# Description

  `ResettingSample.Snapshot()` calls `Clear()` which resets count and sum to 0 on every Prometheus scrape. This causes `_count` metrics (declared as Prometheus `counter` type) to decrease, violating counter monotonicity and breaking `rate()`, `increase()`, and average latency calculations.

  Fixed by maintaining cumulative count and sum in `resettingSample`, while keeping the resetting behavior for sample values used in percentile calculations.

fixes #2173 

  # Changes

  # Breaking changes

  N/A

  # Nodes audience

  N/A

  # Checklist

  # Cross repository changes

  ## Testing

  - [x] I have added unit tests

  # Additional comments

  Affected metrics: all histograms using `ResettingSample`, including `rpc_duration_*` (RPC latency), P2P tracking, and protocol handler metrics.

### DIFF
--- a/metrics/resetting_sample.go
+++ b/metrics/resetting_sample.go
@@ -11,15 +11,24 @@ func ResettingSample(sample Sample) Sample {
 }
 
 // resettingSample is a simple wrapper around a sample that resets it upon the
-// snapshot retrieval.
+// snapshot retrieval. It maintains cumulative count and sum separately so that
+// Prometheus _count counters remain monotonically increasing across scrapes.
 type resettingSample struct {
 	Sample
+	count int64
+	sum   int64
 }
 
 // Snapshot returns a read-only copy of the sample with the original reset.
+// Count and Sum are cumulative for Prometheus counter semantics.
+// Values (used for percentiles) are from the current interval only.
 func (rs *resettingSample) Snapshot() *sampleSnapshot {
 	s := rs.Sample.Snapshot()
+	rs.count += s.Count()
+	rs.sum += s.Sum()
 	rs.Sample.Clear()
 
-	return s
+	return newSampleSnapshotPrecalculated(
+		rs.count, s.Values(), s.Min(), s.Max(), rs.sum,
+	)
 }
--- a/metrics/resetting_sample.go
+++ b/metrics/resetting_sample.go
@@ -20,6 +20,12 @@ type resettingSample struct {
 	count atomic.Int64
 }
 
+// Clear resets both the underlying sample and the cumulative count.
+func (rs *resettingSample) Clear() {
+	rs.Sample.Clear()
+	rs.count.Store(0)
+}
+
 // Snapshot returns a read-only copy of the sample with the original reset.
 // Count is cumulative for Prometheus counter semantics.
 // Values, Sum, Min, Max are from the current interval only.
--- a/metrics/resetting_sample.go
+++ b/metrics/resetting_sample.go
@@ -1,5 +1,7 @@
 package metrics
 
+import "sync/atomic"
+
 // ResettingSample converts an ordinary sample into one that resets whenever its
 // snapshot is retrieved. This will break for multi-monitor systems, but when only
 // a single metric is being pushed out, this ensure that low-frequency events don't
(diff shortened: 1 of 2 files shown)


manav2401:
@lake-dunamu thanks for the PR. Few comments. 

1. `sum` isn't tracked in histogram and only count is tracked so it's okay to remove it from resetting sample. 
2. Please use `atomic.Int64` for `count` so that there are no race conditions. 
3. When `newSampleSnapshotPrecalculated` is called, `rs.sum` is passed which is the total sum. This will affect the mean as we're using total sum instead of sum in this sample. You'll have to replace it with `s.Sum()`. 
4. Please add some basic tests to validate the same.  

Let me know if you need help or have any concerns. Thanks!

lake-dunamu:
@manav2401 

Thanks for the review.

addressed:
  1. Removed `sum` field — not needed in histogram.
  2. `count` now uses `atomic.Int64`.
  3. Passing `s.Sum()` instead of `rs.sum` to fix mean calculation.
  4. Added basic tests.

metrics/resetting_sample.go
```diff
@@ -11,15 +13,22 @@ func ResettingSample(sample Sample) Sample {
...
+// Count is cumulative for Prometheus counter semantics.
+// Values, Sum, Min, Max are from the current interval only.
 func (rs *resettingSample) Snapshot() *sampleSnapshot {
 	s := rs.Sample.Snapshot()
+	count := rs.count.Add(s.Count())
 	rs.Sample.Clear()
 
-	return s
+	return newSampleSnapshotPrecalculated(
+		count, s.Values(), s.Min(), s.Max(), s.Sum(),
```
manav2401: I think you'll have to capture these values before calling `rs.Sample.Clear()`. Haven't checked in too much detail but underlying values are cleared on calling Clear() so it should be captured before clearing.

lake-dunamu: @manav2401 

I think this should be safe — Snapshot() returns a fully independent copy where values are cloned via make/slices.Clone. So Clear() only resets the underlying sample without affecting the new snapshot.

That said, happy to add explicit captures before Clear() if you think it improves readability.

lake-dunamu:
@manav2401 

could you PTAL?

manav2401 (approved):

cffls (approved):

pratikspatil024 (approved):

PR #2181, opened 2026-04-02
consensus/bor: guard nil sprint header in CommitStates

LarryArnault45:
# Description

In the pre-Indore path of `CommitStates`, the state-sync upper timestamp was derived from the sprint-start header time. The previous code dereferenced `GetHeaderByNumber(...).Time` directly, without checking whether the header existed.

Under ancient pruning conditions, that sprint-start header may be unavailable. In that case, the old flow could hit a nil dereference and panic on the consensus finalization path.

This update adds only the required safety checks:

1. compute `sprintLength` once;
2. guard `number < sprintLength` and return a clear error to avoid underflow;
3. load the sprint-start header, validate it is non-nil, and return an explicit error when missing before reading `Time`.

Behavior is unchanged when data is present; only crash-prone branches now fail with explicit errors instead of panicking.

# Changes

# Breaking changes

N/A

# Nodes audience

No node-segmented rollout is required.

# Checklist

# Cross repository changes

- In case link the PR here:
- In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

1. `gofmt -w consensus/bor/bor.go`
2. `go test ./consensus/bor -run TestCommitStates -count=1`
3. `go vet ./consensus/bor`
4. `go test ./consensus/bor/... -count=1`
5. `make lint`

# Additional comments

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -1744,7 +1744,18 @@ func (c *Bor) CommitStates(
 			return nil, err
 		}
 
-		to = time.Unix(int64(chain.Chain.GetHeaderByNumber(number-c.config.CalculateSprint(number)).Time), 0)
+		sprintLength := c.config.CalculateSprint(number)
+		if number < sprintLength {
+			return nil, fmt.Errorf("invalid sprint start block for block %d with sprint length %d", number, sprintLength)
+		}
+
+		sprintStartNumber := number - sprintLength
+		sprintStartHeader := chain.Chain.GetHeaderByNumber(sprintStartNumber)
+		if sprintStartHeader == nil {
+			return nil, fmt.Errorf("failed to fetch sprint start header %d for block %d", sprintStartNumber, number)
+		}
+
+		to = time.Unix(int64(sprintStartHeader.Time), 0)
 	}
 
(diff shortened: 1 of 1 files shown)


manav2401 (commented):
Have you encountered errors in pre-indore blocks due to this? 

I don't think this case will ever happen due to following reasons. 
1. The first condition of `number` being less than `sprintLength` will never occur as first state-sync occurs at the start of first sprint. Hence, the condition is never less than 0. 
2. `CommitStates` is called when your node is syncing and processing that block which has state-sync event. Fetching a slightly old block (16 blocks before current block) shouldn't fail due to ancient pruning - as ancient pruning is for blocks older than 90k from latest block. Hence, pruning a block which is just 16 blocks older is not possible.

LarryArnault45:
Hi @manav2401 , thanks for the review.

I agree this is unlikely in normal pre-Indore sync, and I haven’t reproduced a real mainnet crash with default pruning. My goal here is just defensive hardening: if `GetHeaderByNumber` ever returns `nil`, the old code panics on `.Time`; this change turns that into a normal error path without affecting healthy flow.

If you want, I can remove the `number < sprintLength` check and keep only the nil-header guard.

PR #2187, opened 2026-04-15
triedb/pathdb: fix exponential memory growth in preload BFS

lucca30:
## Summary

- The preload BFS in `biased_fastcache.go` called `gatherChildPaths` which always enqueued all 16 nibble children for every node, regardless of node type (branch, extension, or leaf). This caused the internal `visited` map to grow as **O(16^depth)** — approximately **14 GB per goroutine** for storage tries with 8 GB cache targets.
- On restart, 6 preload goroutines start simultaneously (as configured on mainnet BPs like anonymous-92). Combined visited-map memory exceeds available RAM, causing severe OS swapping that stalls block import for the duration of the preload (~2 hours for the largest caches).
- Replace `gatherChildPaths` with `decodeChildPaths`, which RLP-decodes each trie node and enqueues **only real children**:
  - **Branch nodes** (17-element list): enqueue the 1–16 slots where the child is non-nil
  - **Extension nodes** (2-element list, no terminator): enqueue exactly 1 child using the decoded compact key nibbles
  - **Leaf nodes** (2-element list, terminator set): no children
- Remove the `visited map[string]struct{}` entirely. Because MPT child paths are strictly longer than the parent path, the BFS is **cycle-free by construction** — no visited set is required.
- Queue size is now **O(trie width)** rather than O(16^depth).

## Test plan

- [ ] `TestDecodeChildPaths_BranchNode` — branch node with 3 non-nil children returns correct paths
- [ ] `TestDecodeChildPaths_EmptyBranch` — all-nil branch returns no children
- [ ] `TestDecodeChildPaths_ExtensionNodeEven/Odd` — extension node returns single child at `currentPath + nibbles`
- [ ] `TestDecodeChildPaths_LeafNode` — leaf node returns no children (even and odd key lengths)
- [ ] `TestDecodeChildPaths_InvalidData` — non-RLP input returns nil gracefully
- [ ] `TestCompactKeyToNibbles` — round-trip proof for all compact encoding variants
- [ ] `TestPreloadBFS_CycleFree` — builds a 5-node trie (root branch → leaf, branch → 2 leaves), runs preload without rate limit, verifies all 5 nodes are in cache and BFS terminates (proving cycle-freedom without a visited set)
- [ ] `TestAddressBiasedCache_BurstExceeded` — updated to use a valid branch-node root so child discovery works correctly with the new decoder

### DIFF
--- a/triedb/pathdb/biased_fastcache.go
+++ b/triedb/pathdb/biased_fastcache.go
@@ -250,7 +250,7 @@ func (c *AddressBiasedCache) preloadAddressAsync(db ethdb.Database, addr common.
 		// Decode actual children from the node and enqueue them.
 		// Only real trie children are returned, keeping queue size proportional
 		// to trie width rather than growing exponentially with depth.
-		childPaths := c.decodeChildPaths(nodeData, item.path)
+		childPaths := decodeChildPaths(nodeData, item.path)
 		for _, childPath := range childPaths {
 			queue = append(queue, queueItem{
 				path:  childPath,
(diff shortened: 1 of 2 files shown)


lucca30:
@claude review once

lucca30:
@claude review once again

cffls (approved):
Nice improvement!

pratikspatil024 (approved):

PR #2192, opened 2026-04-20
prefetcher: builder-phase prefetch + streaming worker pool

lucca30:
## Purpose

Today's prefetcher only runs in a speculative **idle phase**: before block-building starts, it scans the tx pool and warms state for what it guesses the block might contain. Whatever it misses — late-arriving p2p txs, txs unlocked by freed gas, anything the pool view didn't reflect at guess-time — pays the full cache-miss cost at commit time. On contract-heavy blocks this is the tail of the miss-rate distribution.

This PR adds **three more prefetch phases, synchronized to the live block build**, so the prefetcher stops guessing and starts following what the builder is actually about to do. All three target exactly the gap the idle phase leaves: near-certain txs the builder will commit, warmed with near-zero speculation.

## Prefetch lifecycle (after this PR)

One long-lived worker pool per block; four sequential **tx providers** feed it. The pool is never torn down between phases — the mode switch is a provider swap, not a prefetcher restart.

| Phase | When | Source of txs | What it warms | Precision character |
|---|---|---|---|---|
| **1. Idle** | Pre-build, while `buildAndCommitBlock` is still assembling its environment | Pool snapshot, scanned by price-nonce heap with a gas-capped budget | Whatever idle can fit before builder starts | Speculative — bounded by pool staleness |
| **2. Builder: upfront plan scan** | Moment builder mode begins, once per `commitTransactions` pass (priority + normal) | `sendPlan` clones the price-nonce heap; `buildTxPlan` walks it with `remainingGas()` as budget, emits every non-prefetched tx that fits | **The bulk of builder-phase prefetch** — the full predicted block contents based on declared gas limits, minus whatever idle already warmed | Near-deterministic — same heap the builder will consume, filtered by the same gas arithmetic |
| **3. Builder: per-tx forwarding** | During `commitTransactions`, right before each apply | Each tx forwarded through `builderPlanCh` as the builder reaches it | Last-mile catch for anything the upfront scan missed (late pool arrivals, prefetch-status changes in flight) | Deterministic — one channel hop ahead of the EVM |
| **4. Builder: freed-gas overflow** | Continuously during commit, as each tx completes | `ltx.Gas − actualUsed` delivered via `builderGasFreedCh`; overflow heap scanned when budget accumulates | Bonus txs the builder promotes into the tail once freed gas accrues — including txs the upfront scan had to `Pop()` as "too large" | Semi-deterministic — tracks the same heap the builder promotes from |

All three builder phases dedup against `prefetchedTxHashes` and a `sentThisPhase` local set, so no tx is re-executed. The three-edge dedup matrix (plan↔overflow within iter, plan→overflow cross-iter, overflow→plan cross-iter) is fully closed.

### Handoff sequence (idle → builder)

Coordinated by a two-signal interrupt to avoid pool teardown:

1. Builder sets `builderStarted` → coordinator sets `evmAbort.Store(true)`. In-flight idle EVM execution aborts via the EVM interrupt; workers entering the loop see the flag and skip.
2. Coordinator drains any buffered idle txs from `txsCh` non-blockingly.
3. `evmAbort.Store(false)`. Workers resume, now fed by the builder provider (upfront-plan + per-tx + freed-gas overflow).

No duplicate prefetches, no lost builder txs, no worker pool churn.

## Precision impact

New headline metric: `worker/prefetch/builder_added_percent` — fraction of a block's txs the builder phase (upfront plan + per-tx + overflow combined) warmed on its own, beyond what idle had already done. Attributes each prefetch completion to the phase it fired under, so operators can see at a glance how much of a block's cache warming idle alone could not have reached.

Also split: `worker/txApplyDuration/{prefetched,notPrefetched}` — quantifies the cache-miss penalty this lifecycle is closing.

Together these two metrics answer: how many txs needed builder-phase help, and how much apply-time that help saved.

## Implementation notes

- **Streaming worker pool.** Workers block on `range txsCh` (buffer 4096, ≈33 KB); phase switches change producer only, not pool.
- **Block-import path unchanged.** `Prefetch(block, ...)` keeps its original signature as a thin wrapper around `PrefetchStream`. Same topology, same parallelism, same `PrefetchResult` shape.
- **Per-tx isolation unchanged.** Each worker still calls `statedb.Copy()` — no shared state, no consensus-path coupling.
- **Complexity.** diffguard: no new hot-path functions exceed the 10-cognitive bar; all flagged functions are pre-existing.

## Validation

- [x] `go build ./...` + `golangci-lint run` clean
- [x] `go test -race ./core/... ./miner/...` — 121s, all prefetch + slow-tx tests pass. Includes 13 new unit tests on the pure primitives (`scanOverflow`, `forwardTxs`, `collectPlanBatch`, `buildTxPlan`, no-duplicate-forward invariant across 10k iterations) and 3 stream-lifecycle integration tests.
- [x] `diffguard --base origin/develop` — no new complexity violations.
- [x] Local kurtosis-pos devnet, 5 scenarios (baseline / builder-heavy / handoff-stress / panic-fidelity / dup-probe). Cross-scenario invariants held on all of them:
  - Zero duplicate forwards across 369 blocks with ~500 tx/block density.
  - `processed == fails + successful` in every `PrefetchStream` exit (2180+ closures).
  - Panic recovery: synthetic panic every 500 tx × 525K txs → `chain_prefetch_worker_panic` meter = **1050**, harness log count = **1050** (exact match). Node kept producing blocks throughout.
  - pprof on a live node: zero leaked prefetch goroutines between blocks (sync.Once shutdown verified).
  - Full lifecycle log sequence (STREAM-START → IDLE-EXIT → HANDOFF-BEGIN/DRAIN/RESET → BUILDER-EXIT → STREAM-END) intact per block across all scenarios.
- [x] Ultra-review pass: one actionable finding (`prepareWork` early-return bypassing `defer interruptPrefetch.Store(true)`) — fixed.
- [x] 5 standard review rounds addressed; final 3 findings closed the cross-iteration dedup matrix.


lucca30:
@claude review once

lucca30:
@claude Review it again. Do a careful analysis on the block building lifecycle, how prefetch works with block building. If any code introduced may lead to a panic, dead code or decreasing on performance. This PR aims to increase the window of prefetching by do also prefetching while block is building.

Some mechanism were introduced like buildTxPlan to let us better predict the most probable txs to be introduced so we can improve our prefetch precision.

lucca30:
@claude ultra-review this PR.

**Context:**
- 8 commits on top of develop; feature is a streaming prefetcher with idle→builder phase handoff.
- 5 rounds of standard @claude review already landed; all findings addressed through `827c017fb`.
- 13 unit tests cover the pure helpers (`scanOverflow`, `forwardTxs`, `collectPlanBatch`, `buildTxPlan`, `runBuilderTxProvider` no-dup invariant).
- Kurtosis devnet run #2 empirically validated: `BUILDER-DUP=0` across 369 blocks, panic meter == log count exactly (1050==1050), 0 leaked goroutines in pprof, `STREAM-END` arithmetic 2180/2180 ok, `builder_added_percent` p50 climbed 0→86 under tight idle cap.

**Please focus only on angles the above does not already cover:**

**1. Concurrency correctness.** Trace every interaction between `hardKill`, `evmAbort`, `inBuilderPhase`, `sync.Once`-shutdown, `close(txsCh)`, and the worker pool. Look for:
- lost-wakeup between `evmAbort.Store(true)` and `drainTxChan`
- double-close of `txsCh`
- a goroutine that can observe `hardKill=false`, `evmAbort=true`, closed `txsCh` in an order that stalls it
- panic during handoff hold that skips `sync.Once`

**2. Resource-accounting invariants, proven not just asserted:**
- (a) `txIndex == valid + invalid` on every `PrefetchStream` exit (normal / hardKill / evmAbort+close / panic / interrupt)
- (b) No tx hash reaches `txsCh` twice in a single `runBuilderTxProvider` invocation
- (c) `builderPrefetchedTxHashes` is only written while `inBuilderPhase==true`

**3. Perf regression on the hot import path.** Compare the new `PrefetchStream` wrapper (used by `blockchain.go` for block import) to the old per-call `Prefetch()`. Any path where the streaming version does strictly more work for the same input?

**4. Failure modes:**
- `builderPlanCh` closed before `runBuilderTxProvider` starts
- empty txPool at `buildOverflowHeap` time
- stale pool snapshot (overflow heap built at T, consumed at T+100ms after pool churned)
- concurrent `hardKill` + `evmAbort` set from different goroutines mid-phase

**Scope exclusions:**
- Don't re-flag findings already fixed in `88c4e44..827c017fb`.
- Don't comment on test coverage (unit tests recent; kurtosis evidence above).
- Don't suggest naming / style changes.
- Flag only findings reachable in production.

miner/worker.go
```diff
@@ -2069,129 +2319,377 @@ func (w *worker) buildAndCommitBlock(interrupt *atomic.Int32, noempty bool, genP
...
 		return
 	}
-	signer := types.MakeSigner(w.chainConfig, header.Number, header.Time)
+
 	prefetcher := core.NewStatePrefetcher(w.chainConfig, w.chain.HeaderChain())
+	// 4096 ≈ one full block's worth of 21k-gas txs at the 100M-gas block limit.
+	// Sized to absorb the idle provider's per-loop burst (bounded by gas budget)
+	// without ever blocking a sender; workers drain far faster than the idle
+	// heap can fill in practice. Channel memory is ~33 KB — negligible.
+	txsCh := make(chan *types.Transaction, 4096)
```
cffls: Would be good to move the constants to the top of the file.

miner/worker.go
```diff
@@ -2069,129 +2319,377 @@ func (w *worker) buildAndCommitBlock(interrupt *atomic.Int32, noempty bool, genP
...
+
+// runIdleTxProvider speculatively streams transactions from the txpool into the
+// prefetcher. It loops on a ~100ms cadence, bounded by a configurable gas budget
+// (PrefetchGasLimitPercent of header.GasLimit, defaulting to 100%). Returns when
+// the budget is exhausted, the builder flips, or hardKill fires.
+//
+// Gas accounting uses declared tx gas (not actual execution gas) — close enough
+// since the budget only bounds speculative work, not correctness.
+func (w *worker) runIdleTxProvider(txsCh chan<- *types.Transaction, header *types.Header, genParams *generateParams, interrupt *atomic.Bool) {
+	const minLoopInterval = 100 * time.Millisecond
```
cffls: Same as above.

miner/worker.go
```diff
@@ -2069,129 +2319,377 @@ func (w *worker) buildAndCommitBlock(interrupt *atomic.Int32, noempty bool, genP
...
+}
+
+// idleGasLimitPercent returns the configured prefetch gas budget percent, capped
+// defensively at 150 and defaulted to 100 when unset.
+func idleGasLimitPercent(cfg *Config) uint64 {
+	pct := cfg.PrefetchGasLimitPercent
+	if pct == 0 {
+		return 100
+	}
+	if pct > 150 {
```
cffls: Same as above.

miner/worker_test.go
```diff
@@ -2681,10 +2682,17 @@ func TestPrefetchMultiBlock(t *testing.T) {
...
-		t.Errorf("Goroutine leak detected: delta=%d", goroutineDelta)
+	// Check for goroutine leaks.
+	// The streaming prefetcher keeps ~N=4*NumCPU/5 worker goroutines alive for the
+	// whole duration of each block's prefetch (vs short-lived per-batch goroutines in
+	// the old batched prefetch). If one or two commitWork goroutines are still
+	// blocked in bor.Prepare at measurement time, their prefetchers keep N+2
+	// goroutines alive. We allow for two in-flight blocks to be mid-production:
+	// 2 * (N + 2) + 10 (runtime/test noise). A genuine leak scales with block count
+	// and would exceed this easily.
+	maxAllowed := 2*(4*runtime.NumCPU()/5+2) + 10
```
cffls, line 2744: `+10` seems arbitrary. Why is it necessary?

miner/worker.go
```diff
@@ -1834,11 +1991,69 @@ func (w *worker) fillTransactions(interrupt *atomic.Int32, env *environment) err
...
+				}
+			}()
+			for _, tx := range buildTxPlan(clone, gasLimit, prefetchedHashes) {
+				select {
+				case ch <- tx:
+				default:
+				}
+			}
+		}()
+	}
```
cffls: Can we move this inline function to a standalone function?

miner/worker.go
```diff
@@ -1795,13 +1854,111 @@ func (w *worker) buildDefaultFilter(BaseFee *big.Int, Number *big.Int) txpool.Pe
...
+// scanOverflow greedily pops transactions from the heap that fit within the freed-gas
+// budget, returning new txs to prefetch and the remaining budget. Already-prefetched
+// txs are skipped entirely (not counted against the budget) because the freed gas
+// represents extra block capacity beyond the existing plan — plan txs' gas is already
+// committed to the main gas pool and should not be double-counted here.
+func scanOverflow(
+	h *transactionsByPriceAndNonce,
+	budget uint64,
+	prefetchedHashes *sync.Map,
+	sentThisPhase map[common.Hash]struct{},
```
cffls: Can we use a different name? `sentThisPhase` doesn't make much sense to me.

miner/worker_test.go
lucca30: Before this PR the threshold was just a flat `> 10` with no per-block scaling at all and it didn't flake on that account, so `10` is empirically the right size for that noise floor. What's new in this PR is the streaming prefetcher keeping `4*NumCPU/5 + 2` goroutines alive per in-flight block: that's the part that needed the per-block scaling term `2 * (N + 2)` . The `+10` is just the previously-validated noise budget left in place. A genuine leak scales with block count and would blow past this easily.

miner/worker.go
lucca30: Done — moved to the top-of-file const block as `prefetchChanBufSize` (4096), with the sizing rationale in the doc comment.

miner/worker.go
lucca30: Done — moved to the top-of-file const block as `prefetchIdleLoopInterval`.

miner/worker.go
lucca30: Done — both the cap and the default are now top-of-file constants: `prefetchMaxGasLimitPercent` (150) and `prefetchDefaultGasLimitPercent` (100).

miner/worker.go
lucca30: Done — promoted to a free function `sendPlan` placed next to `buildTxPlan`.

miner/worker.go
lucca30: Renamed to `inFlightHashes` — captures the actual semantics: txs already forwarded on `txsCh` whose `onSuccess` hasn't fired yet.

manav2401 (approved):

manav2401 (dismissed):

manav2401 (approved):

cffls (approved):

PR #2193, opened 2026-04-21
all: fix nightly race test flakes

cffls:
The nightly-race workflow had been failing on every run for a month. Most failures were either race-detector overhead pushing test timings past narrow assertion margins, tests mutating shared global state that collided with parallel tests, or a single real data race that cascaded into many victim tests.

internal/ethapi: testBackendWithPreMadhuguriBorReceipt.ChainConfig() did a shallow copy of params.AllEthashProtocolChanges, leaving cfg.Bor aliased to the global BorConfig, then mutated MadhugiriBlock on it. Concurrent tests calling IsMadhugiri raced with the write. Deep-copy BorConfig like the sibling testBackendWithNilBorTx already does. This single race was the root cause of the ~10 cascading ethapi test failures (TestBorForks, TestBorGetLogs_*, TestCoinbase, TestEstimateGas, etc.).

consensus/bor/heimdall: TestFailover_SwitchOnPrimaryDown and TestRegistry_MarkUnhealthyOnRealFailure set only getSpanFn on the primary mock, leaving FetchStatus returning success. The registry's background probe then raced with the test's MarkUnhealthy call and could flip the primary back to healthy (or the active gauge back to 0) before the assertion ran. Make the primary mock fail FetchStatus too so probe and API are consistent.

core/state: TestConcurrentUsedParallelism measures wall-clock parallel speedup and asserts >=2x. Race instrumentation serializes atomic/mutex ops and skews the measurement to ~1.7x. Skip under -race via a new race_{on,off}_test.go build-tag pair; the test still guards against the global-lock regression in non-race runs.

core/txpool/legacypool: TestLockOrdering_{PricedHeapNoDeadlock, ReplacePendingNoDeadlock,RemovedNoDeadlock} used a 10s deadlock-detect timeout that is genuinely too short once -race instrumentation is added (the bare test takes ~12s). Bump to 60s — still catches real deadlocks, no longer fires on legitimate completion.

eth/downloader: TestBeaconSync68/69Full used a 3s sync timeout that CI routinely exceeded under -race. Bump to 30s.
TestSkeletonSyncRetrievals had four 2s polling loops that could exit before background serving goroutines finished incrementing the served counter (the assertion then read a partial count). Bump the budget to 30s, cap per-iteration sleep at 500ms so exponential backoff stays responsive, and fold the served counter into the polling condition so we wait for both subchain state and served totals before asserting.

eth/relay: TestCheckTxStatus, TestSubmitPreconfTx, TestSubmitPrivateTx parallel subtests had handlers sleeping for rpcTimeout-100ms, leaving only 100ms of slack before the client-side 2s timeout fired; under -race that slack was easily exhausted. Halve the handler sleep to rpcTimeout/2 — still proves parallelism (3 serial calls would exceed the unchanged 2s upper bound) with comfortable margin for -race.

metrics: TestExpDecaySampleNanosecondRegression relied on the unseeded global RNG and wall-clock timing, making the reservoir's average a noisy statistic that occasionally drifted outside [14, 16] under -race. Drive the test with a seeded RNG and synthesised monotonic timestamps; override t0/t1 after NewExpDecaySample so dt in update() stays positive. Test still catches the priority-overflow regression (average would stick at 10 under the buggy formula).

### DIFF
--- a/core/state/race_off_test.go
+++ b/core/state/race_off_test.go
@@ -1,0 +1,5 @@
+//go:build !race
+
+package state
+
+const raceEnabled = false
--- a/consensus/bor/heimdall/failover_client_test.go
+++ b/consensus/bor/heimdall/failover_client_test.go
@@ -197,10 +197,13 @@ func TestFailover_SwitchOnPrimaryDown(t *testing.T) {
 	switchesBefore := failoverSwitchCounter.Snapshot().Count()
 	activeBefore := failoverActiveGauge.Snapshot().Value()
 
+	connErr := &net.OpError{Op: "dial", Net: "tcp", Err: errors.New("connection refused")}
+	// Primary must fail both GetSpan and FetchStatus so the background probe
+	// cannot promote it back to healthy and flip the active gauge back to 0
+	// before this test's final assertion.
 	primary := &mockHeimdallClient{
-		getSpanFn: func(ctx context.Context, _ uint64) (*types.Span, error) {
-			return nil, &net.OpError{Op: "dial", Net: "tcp", Err: errors.New("connection refused")}
-		},
+		getSpanFn:     func(ctx context.Context, _ uint64) (*types.Span, error) { return nil, connErr },
+		fetchStatusFn: func(_ context.Context) (*ctypes.SyncInfo, error) { return nil, connErr },
 	}
 	secondary := &mockHeimdallClient{}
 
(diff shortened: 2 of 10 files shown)


core/state/race_off_test.go
```diff
@@ -0,0 +1,5 @@
+//go:build !race
+
+package state
```
lucca30, line 3: Why this file and also the `race_on_test.go`? Interesting, but never saw something similar to test.

cffls: It's the canonical Go idiom for "did this binary get built with `-race`?" — the stdlib has no public API for it, so you define a symbol twice with paired build tags and let the compiler pick exactly one. The `race` build tag is set automatically by `go test -race`.

```go
//go:build race
package state
const raceEnabled = true
```
```go
//go:build !race
package state
const raceEnabled = false
```

Same pattern as `internal/race` inside Go's stdlib and `runtime/race` — e.g. `golang.org/x/sync/errgroup`, `golang.org/x/crypto/chacha20`, and several geth internals use exactly this shape. I suffixed the files with `_test.go` so the constant is scoped to tests only and doesn't bloat the production binary.

The reason I needed it here is that `TestConcurrentUsedParallelism` measures wall-clock parallel speedup and asserts ≥2×. The race detector serializes atomic/mutex ops and reliably cuts measured speedup to ~1.7×, so the assertion spuriously fails. `raceEnabled` lets the test `t.Skip` under `-race` while still running normally in the main test job — which is where the assertion actually guards against a global-lock regression. The alternative (lowering the threshold to something like 1.3×) weakens the check for everyone just to avoid one CI config.

lucca30: Wow, supercool Thanks for sharing

lucca30 (approved):

kamuikatsurgi (approved):

PR #2235, opened 2026-05-19
consensus/bor: fix race in SpanStore.PurgeCache

lucca30:
## Summary

`SpanStore.PurgeCache` clears `latestSpanCache` (and the other cache fields) via an atomic `Store(nil)`, but does not stop the background polling goroutine that `NewSpanStore` spawns ([span_store.go:62-89](https://github.com/0xPolygon/bor/blob/develop/consensus/bor/span_store.go#L62-L89)). That goroutine ticks every 200ms and writes the latest span back into `latestSpanCache`. If a tick lands between the clear and the caller's next read, the \"purge\" is silently undone.

The window is microscopic in the original test (the assertion runs in microseconds after the clear), so this manifests as a low-rate CI flake on `TestSpanStore_PurgeCache` rather than a consistent failure.

## Where it showed up

Caught while iterating on #2192:

- Failing CI run: https://github.com/0xPolygon/bor/actions/runs/26121340018/job/76823989362
- Failure signature: `latestSpanCache` contained `span Id:0x2` — the exact value the mock heimdall client returns from `GetLatestSpan`, confirming the polling goroutine wrote it back after the purge.

`PurgeCache` is test-only — it is only invoked from `bor_test.go`, `tests/bor/bor_test.go`, and the span-store test itself — so changing its lifecycle behavior is low-risk.

## Fix

- Extract the poll loop into a `runPollLoop` method tracked by a `sync.WaitGroup`.
- Have `PurgeCache` cancel the context and `Wait()` for the goroutine to exit *before* resetting state. `Close()` uses the same path (`stopPollLoop`) so it's also race-safe.
- `PurgeCache` no longer restarts the loop. On-demand reads via `getLatestSpan`/`spanById` fall back to inline `updateLatestSpan` when the cache is empty, so callers still get fresh data — they just don't get background warming until the next `NewSpanStore`. This matches the test-only use case.

Also slimmed the loop body by extracting the rate-limited error logger into a small helper so the new method stays under diffguard's cognitive-complexity threshold.

## Reproducer

`TestSpanStore_PurgeCache_RaceWithPollLoop` sleeps 300ms after `PurgeCache` (past one tick of the 200ms poll). On `develop` it fails 5/5 with the same `span Id:0x2` signature as the CI flake; with the fix it passes deterministically.

## Test plan

- [x] `go build ./...` clean
- [x] `go test -count=20 -run TestSpanStore_PurgeCache ./consensus/bor/` — 40/40 pass (new reproducer + original)
- [x] `go test -race ./consensus/bor/...` — 526 tests pass
- [x] `diffguard --base origin/develop --skip-mutation .` — all sections PASS

### DIFF
--- a/consensus/bor/span_store.go
+++ b/consensus/bor/span_store.go
@@ -400,6 +400,10 @@ func (s *SpanStore) PurgeCache() {
 	s.lastUsedSpan.Store(nil)
 	// Reset the latest known span ID
 	s.latestKnownSpanId.Store(0)
+	// Clear cached heimdall status. With the poll loop stopped, a stale
+	// CatchingUp:false would otherwise let waitUntilHeimdallIsSynced skip
+	// refreshing status against a newly-swapped heimdall client.
+	s.heimdallStatus.Store(nil)
 }
 
 // getMockSpan0 constructs a mock span 0 by fetching validator set from genesis state. This should
--- a/consensus/bor/span_store.go
+++ b/consensus/bor/span_store.go
@@ -3,6 +3,7 @@ package bor
 import (
 	"context"
 	"fmt"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cffls (approved):
lgtm, thanks!

manav2401 (approved):

PR #2238, opened 2026-05-22
miner: drop PeerCount==0 gate in mainLoop

cffls:
The peer-count drop branch in `mainLoop` was introduced in #977 to delay block production until at least one peer was connected, scoped to BorMainnet/Mumbai (Amoy was added later in e8157bbe6). #2220 widened the gate from a chain-ID list to "any heimdall-connected node", which drew kurtosis devnets into the gate for the first time: in a single-validator kurtosis devnet PeerCount() is 0 by design (no other node to peer with), the gate drops every newWorkReq, and the veblop fallback timer re-fires straight back into the same drop.

Veblop's primary/backup producer election already handles outage safety at the protocol level — peer count adds no safety, only a foot-gun. Strip the gate; `mainLoop` now routes every newWorkReq directly to `commitWork` (except the pre-existing
DisablePendingBlock-on-non-validator short-circuit).

Verified end-to-end on a kurtosis 1-validator devnet (chain ID 4927, veblop active).

### DIFF
--- a/miner/fake_miner.go
+++ b/miner/fake_miner.go
@@ -221,9 +221,8 @@ func (m *mockBackendBor) BlockChain() *core.BlockChain {
 	return m.bc
 }
 
-// PeerCount implements Backend. Returns 1 so the worker's mainLoop sees the
-// node as peered — tests using mockBackendBor are not exercising the
-// PeerCount==0 gate.
+// PeerCount implements Backend. Returns a constant; tests using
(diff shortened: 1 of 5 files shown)


cffls:
@claude review

kamuikatsurgi (approved):

pratikspatil024 (approved):

PR #2239, opened 2026-05-25
miner: format build-block log times as RFC3339Nano, gate on IsRunning

kamuikatsurgi:
## Summary

The "Starting to build block" log previously emitted epoch milliseconds (`buildStart(ms)`, `parentActualTime(ms)`, `headerActualTime(ms)`) and raw Unix seconds for `parentTime`/`headerTime`, with Go's default comma-grouped formatting making the values awkward to parse and hard to correlate across services. Switch every timestamp field to `time.RFC3339Nano` for human-readable, parser-friendly output that matches the rest of the bor log surface and lines up with Heimdall.

Also wrap the log in `w.IsRunning()` so non-mining nodes don't emit it on every prepare cycle.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -2315,23 +2315,25 @@ func (w *worker) buildAndCommitBlock(interrupt *atomic.Int32, noempty bool, genP
 		stopFn()
 	}()
 
-	timeUntilInterrupt := time.Until(work.header.GetActualTime())
-	if timeUntilInterrupt > time.Second {
-		timeUntilInterrupt -= interruptBuffer
+	if w.IsRunning() {
+		timeUntilInterrupt := time.Until(work.header.GetActualTime())
+		if timeUntilInterrupt > time.Second {
+			timeUntilInterrupt -= interruptBuffer
+		}
+		parent := w.chain.CurrentBlock()
+		log.Info("Starting to build block", "number", work.header.Number.Uint64(),
+			"buildStart", prepareWorkStart.Format(time.RFC3339Nano),
+			"preBuild", common.PrettyDuration(genParams.preBuildDuration), // time spent before `buildAndCommitBlock` is called
+			"prepareWork", common.PrettyDuration(prepareWorkDuration), // total time spent in prepare work
+			"makeEnv", common.PrettyDuration(work.makeEnvDuration), // total time spent in `makeEnv` inside prepare work
+			"makeHeader", common.PrettyDuration(work.makeHeaderDuration), // total time spent in `makeHeader` inside prepare work includes bor.Prepare call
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -2315,23 +2315,25 @@ func (w *worker) buildAndCommitBlock(interrupt *atomic.Int32, noempty bool, genP
...
+		log.Info("Starting to build block", "number", work.header.Number.Uint64(),
+			"buildStart", prepareWorkStart.Format(time.RFC3339Nano),
+			"preBuild", common.PrettyDuration(genParams.preBuildDuration), // time spent before `buildAndCommitBlock` is called
+			"prepareWork", common.PrettyDuration(prepareWorkDuration), // total time spent in prepare work
+			"makeEnv", common.PrettyDuration(work.makeEnvDuration), // total time spent in `makeEnv` inside prepare work
+			"makeHeader", common.PrettyDuration(work.makeHeaderDuration), // total time spent in `makeHeader` inside prepare work includes bor.Prepare call
+			"parentTime", time.Unix(int64(parent.Time), 0).Format(time.RFC3339Nano),
+			"parentActualTime", parent.GetActualTime().Format(time.RFC3339Nano),
+			"headerTime", time.Unix(int64(work.header.Time), 0).Format(time.RFC3339Nano),
+			"headerActualTime", work.header.GetActualTime().Format(time.RFC3339Nano),
```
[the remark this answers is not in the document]
kamuikatsurgi: Fixed

manav2401 (approved):

miner/worker.go
```diff
@@ -2315,23 +2315,25 @@ func (w *worker) buildAndCommitBlock(interrupt *atomic.Int32, noempty bool, genP
...
-	timeUntilInterrupt := time.Until(work.header.GetActualTime())
-	if timeUntilInterrupt > time.Second {
-		timeUntilInterrupt -= interruptBuffer
+	if w.IsRunning() {
+		timeUntilInterrupt := time.Until(work.header.GetActualTime())
+		if timeUntilInterrupt > time.Second {
+			timeUntilInterrupt -= interruptBuffer
+		}
+		parent := w.chain.CurrentBlock()
+		log.Info("Starting to build block", "number", work.header.Number.Uint64(),
```
lucca30, line 2324: Is this a new log? I never observed on any node. If its old, probably its from a flow that only runs on Geth.

manav2401: Hey Lucca, this log was added recently by me in https://github.com/0xPolygon/bor/pull/2228 to get some more observability in the Prepare phase and getting some absolute values to know which code path it took during block building.

lucca30: oh nice, got it. thanks

lucca30 (approved):

PR #2242, opened 2026-05-26
feat: add p2p observability metrics

vbhattaccmu:
## Summary

Adds focused Bor P2P observability for peer-drop RCA: a peer disconnect counter (p2p/peers/drops) and devp2p ping RTT timer (p2p/peers/ping/latency). This lets operators correlate peer-drop bursts with peer liveness/latency from without relying only on log scraping. 

## Executed tests

a. go test ./p2p
    Result: pass.

b. Kurtosis Polygon PoS devnet stress test with Amoy configuration:
    4 Bor validators, 6 Bor RPC nodes
    distributed polycli loadtest across all 6 RPC endpoints
    load window: 2026-05-26T14:47:45Z to 2026-05-26T14:52:47Z

## Rollout notes
No coordinated upgr
ade required.

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -113,6 +113,7 @@ type Peer struct {
 	protoErr chan error
 	closed   chan struct{}
 	pingRecv chan struct{}
+	pongRecv chan struct{}
 	disc     chan DiscReason
 
 	// events receives message send / receive events if set
@@ -258,6 +259,7 @@ func newPeer(log log.Logger, conn *conn, protocols []Protocol) *Peer {
 		protoErr: make(chan error, len(protomap)+1), // protocols + pingLoop
 		closed:   make(chan struct{}),
 		pingRecv: make(chan struct{}, 16),
+		pongRecv: make(chan struct{}, 1),
 		log:      log.New("id", conn.node.ID(), "conn", conn.flags),
 	}
 	return p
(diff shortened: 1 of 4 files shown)


pratikspatil024:
@claude review once

p2p/peer.go
```diff
@@ -370,11 +395,22 @@ func (p *Peer) readLoop(errc chan<- error) {
 func (p *Peer) handle(msg Msg) error {
 	switch {
 	case msg.Code == pingMsg:
-		msg.Discard()
+		if err := msg.Discard(); err != nil {
+			return err
```
kamuikatsurgi: Returning err here means peer will be disconnected, are we sure we wanna do that?

p2p/peer.go
```diff
@@ -370,11 +395,22 @@ func (p *Peer) readLoop(errc chan<- error) {
...
+		}
 		select {
 		case p.pingRecv <- struct{}{}:
 		case <-p.closed:
 		}
+	case msg.Code == pongMsg:
+		if err := msg.Discard(); err != nil {
+			return err
```
kamuikatsurgi: Same

p2p/peer.go
vbhattaccmu: I’ve updated it to preserve the previous behavior i.e. discard ping/pong payloads best-effort and only use pong as a latency signal.

p2p/peer.go
```diff
@@ -332,18 +334,41 @@ func (p *Peer) pingLoop() {
...
+				}
+			}
 			if err := SendItems(p.rw, pingMsg); err != nil {
 				p.protoErr <- err
 				return
 			}
+			if metrics.Enabled() {
+				pingSent = time.Now()
+			} else {
+				pingSent = time.Time{}
```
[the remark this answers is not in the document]
vbhattaccmu: Thanks. I implemented the fix: in pingLoop() we now skip sending a new ping while pingSent is non-zero (we reset the timer and continue), so we don't overwrite the outstanding ping timestamp. This prevents a late Pong from being matched to a newer Ping (which produced misleadingly small RTT) and avoids silently dropping the real pong.

`go test ./p2p` passes.

@claude review

kamuikatsurgi (dismissed):

vbhattaccmu:
@claude review

p2p/peer.go
```diff
@@ -328,29 +330,61 @@ loop:
...
+			drainStalePongs(p.pongRecv)
 			if err := SendItems(p.rw, pingMsg); err != nil {
 				p.protoErr <- err
 				return
 			}
-			ping.Reset(pingInterval)
+			if metrics.Enabled() {
+				pingSent = time.Now()
+			}
+			ping.Reset(interval)
```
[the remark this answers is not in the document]
vbhattaccmu: Good catch, fixed. The ping loop now still sends keepalive pings every interval even when a previous pong is outstanding. `pingSent` only gates whether we start a new latency sample, so a lost pong won’t permanently stop pings. I also updated the test to assert that pings continue while waiting for a pong.

vbhattaccmu:
@claude review

lucca30 (approved):
lgtm

pratikspatil024 (approved):

PR #2255, opened 2026-06-04
core/blockstm, core: order same-authority EIP-7702 txs in V2 execution

pratikspatil024:
## Summary

V2 parallel execution (BlockSTM v2) is missing a dependency edge between transactions that share an **EIP-7702 authority**. Applying an authorization reads and then mutates the *authority account's* nonce and code — but that account is typically neither the transaction's sender nor its call target. The existing scheduling hints (`toPrev` same-`to` chains, same-sender nonce chains) therefore never order two transactions that share an authority.

As a result, a later same-authority transaction can **finalize before the earlier one's authority write lands in the MVStore**, baking a stale base read into the block and diverging gas. The observable symptom is an `invalid gas used` / "bad block" that re-executes correctly on retry (the write has landed by the time it re-runs).

This PR adds the missing ordering edge:

- `computeAuthorityPrev` chains each transaction to the most recent earlier transaction sharing one of its authorities.
- `execute()` waits on that predecessor's **completion (finalization)** — not merely its first speculative flush — before running, so the authorization nonce/code check runs against committed authority state. Finalization is monotonic in transaction order, so the wait is deadlock-free.
- Authorities are recovered once per task (`computeTaskAuthorities`) and the slice is shared between the sender-nonce precompute and the new ordering, so the per-authorization `ecrecover` isn't paid twice per block.
- `auth_stalls` / `auth_stall_dur` are surfaced in the V2 block stats so the serialization cost of same-authority bursts is observable.

The dependency gap is **latent in V2 regardless of import mode**. It was observed in production under pipelined import (`pipelineMode=flatdiff`), whose finalization timing widens the race window — but the fix closes the gap unconditionally rather than depending on import mode.

### Why this is correct, not just a symptom mask

The accompanying `ParallelStateDB` tests show the **validation predicate was already sound** — once the earlier authority write is committed to the MVStore, a later transaction's recorded stale read is correctly rejected by `Validate()`. So the defect is purely **execution ordering** (a later tx finalizing before the earlier tx's write is visible), which is exactly what the new `authPrev` barrier fixes.

## Executed tests

Beyond the standard CI gates:

- `TestV2SharedAuthorityWaitsForFinalization` (new) — pins the ordering barrier. **Verified it fails without the fix** (`tx1 started before tx0 finalized`) and passes with it.
- `TestComputeAuthorityPrev` (new) — pins the same-authority chaining logic.
- `TestPDB_AuthorityNonceStaleRead_LaterWriterInvalidates` / `TestPDB_AuthorityCodeStaleRead_LaterWriterInvalidates` (new) — pin that `Validate()` rejects the stale authority nonce/code read once the earlier write is committed, localizing the defect to ordering rather than the validation predicate.
- Full `go test ./core/blockstm/` suite green (~40s), `go vet ./core/blockstm/` clean, `gofmt` clean.

Field context: the originating branch ran parallel-EVM **on** with pipelined-import-src **off** for 3 days and observed zero bad blocks; with pipelined-src on, the bad blocks reproduced and each self-healed via direct-mode retry. Both "pipelined off" and this `authPrev` fix independently eliminate the bad blocks — they break different links in the same chain (pipelined timing is the trigger, the missing BlockSTM dependency is the cause).

## Rollout notes

- **Consensus-affecting:** yes — changes V2 parallel execution scheduling so same-authority EIP-7702 transactions produce the canonical gas/state. It removes a source of non-deterministic bad blocks; output for any block without shared-authority EIP-7702 transactions is unchanged.
- **Backwards-compatible:** yes — no config, genesis, wire, or storage changes. Pure executor ordering plus added observability fields.
- **Coordinated upgrade:** not required. Nodes without the fix self-heal each bad block via direct-mode re-execution, but carry the latent divergence risk; this makes V2 deterministic for the shared-authority case.
- **Performance:** the barrier only blocks transactions that *share* an EIP-7702 authority (rare); `auth_stalls` / `auth_stall_dur` block stats quantify any serialization cost.

### DIFF
--- a/core/blockstm/v2_executor.go
+++ b/core/blockstm/v2_executor.go
@@ -22,9 +22,10 @@ type V2Task interface {
 	To() *common.Address // nil for contract creation
 	// Authorities returns the unique EIP-7702 authorities of this tx
 	// (signers in the SetCode auth list). Empty for non-SetCodeTx.
-	// V2 uses these to maintain per-address nonce deltas across the
-	// block, since each authority's nonce is bumped at apply time even
-	// though it isn't the tx sender.
+	// V2 uses these both for nonce publication and for execution ordering:
+	// each applied authorization mutates the authority account's nonce/code
+	// even though the authority may be neither the tx sender nor the call
+	// target.
 	Authorities() []common.Address
 }
 
@@ -86,6 +87,10 @@ type V2ExecutionResult struct {
 	ValWaitDur  time.Duration // time waiting for workers
 	ValCheckDur time.Duration // time doing validation checks
 	ValReexDur  time.Duration // time blocked on re-execution
+	// Same-authority EIP-7702 serialization cost: tasks that blocked on an
+	// authPrev predecessor, and aggregate wall time spent blocked.
+	AuthStallCount int
+	AuthStallDur   time.Duration
 	// Non-nil if runValidationLoop recovered from a panic; caller should
 	// discard this result and fall back to serial.
 	ValidationPanic any
@@ -156,11 +161,18 @@ type v2ExecCtx struct {
 	vfailIdxs    []int
 	vfailed      []atomic.Bool
 	toPrev       []int
+	authPrev     []int
 	completionCh []chan struct{}
 	finalized    []atomic.Bool
 	execDone     []chan struct{}
 	chSettle     chan int
 	reexecWg     sync.WaitGroup // re-exec goroutines launched by dispatchReexec
+
+	// authPrev serialization observability: how many tasks actually blocked
+	// on a same-authority predecessor and for how long in aggregate. Written
+	// from worker goroutines, so atomic.
+	authStallCount atomic.Int64
+	authStallNanos atomic.Int64
 
 	taskSenderNonces []map[common.Address]uint64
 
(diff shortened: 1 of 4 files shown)


cffls:
@claude review

core/blockstm/v2_executor.go
```diff
@@ -419,6 +473,22 @@ func (x *v2ExecCtx) execute(taskIdx, workerID int) {
...
+		// than racing its publication.
+		tStall := time.Now()
+		select {
+		case <-x.completionCh[prev]:
+		case <-x.ctx.Done():
+			return
+		}
+		x.authStallCount.Add(1)
+		x.authStallNanos.Add(int64(time.Since(tStall)))
+	}
```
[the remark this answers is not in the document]
pratikspatil024: Good catch — confirmed and fixed in ba799bc8a.

The analysis is correct: the new `authPrev` barrier parks a worker on `completionCh[prev]`, which is closed only by `finalizePass`/`finishReexec` inside the validation goroutine. The executor ran on the caller's ctx verbatim (long-lived in production, `context.Background()` in tests), and the `recover` only recorded the panic — so a `Validate()` panic on a shared-authority predecessor left `completionCh[prev]` unclosed and `ctx.Done()` unfired, hanging `wg.Wait()` and defeating the serial fallback.

Fix takes option 1 (the more general one): `ExecuteV2BlockSTM` now derives a cancellable ctx (`context.WithCancel`) and stores `cancel` on the exec context; the validation `recover` calls `x.cancel()` before logging, so every worker parked on the `authPrev` (or finalization) wait is released via `ctx.Done()` and the caller falls back to serial.

The pre-existing same-`To` wait is unaffected, as you noted — it's gated on `vfailed[i-1]` and `validateOne(i)` finalizes its predecessor before `Validate(i)` runs, so a panic can't strand it.

Added `TestV2SharedAuthorityValidationPanicNoDeadlock` to pin it: two txs share an authority, the earlier one's `Validate()` panics, and the executor must return with `ValidationPanic` set. Verified it deadlocks (5s timeout) without the cancel-on-panic and passes with it.

core/blockstm/v2_executor.go
```diff
@@ -115,9 +120,15 @@ func ExecuteV2BlockSTM(
 	if ctx == nil {
 		ctx = context.Background()
 	}
+	// Derive a cancellable ctx so the validation goroutine can release workers
+	// parked on completionCh waits (e.g. the authPrev barrier) if it panics —
+	// the incoming ctx is typically long-lived and never cancelled on panic.
```
kamuikatsurgi, line 125: These comments seem very obvious. Can we trim overall comments in code as well as tests?

pratikspatil024: The comments these days are not for humans but for agents, I would keep them as they are not hurting.

pratikspatil024:
@claude review. Addressed your previous comments

pratikspatil024:
Not needed after https://github.com/0xPolygon/bor/pull/2270. Closing.

PR #2256, opened 2026-06-04
build: bump Go to 1.26.4, add Dependabot, govulncheck nightly

kamuikatsurgi:
## Summary

Three changes to reduce govulncheck noise on PRs.

### 1. Bump Go 1.26.3 → 1.26.4 (fixes current govuln findings)

Clears two stdlib advisories that govulncheck has been flagging on every PR:
- **GO-2026-5039**: Arbitrary inputs in errors without escaping in `net/textproto` (called via `node.StartHTTPEndpoint` → `http.Server.Serve` → `textproto.Reader.ReadMIMEHeader`)
- **GO-2026-5037**: Inefficient candidate hostname parsing in `crypto/x509` (called via `node.StartHTTPEndpoint`, `ethstats.Service.loop`, `cli.InspectAncientDbCommand.Run`)

After the bump, `govulncheck ./...` reports `No vulnerabilities found.` locally.

### 2. Move `govulncheck` workflow off PRs onto a nightly schedule

The workflow now runs daily at `00:00 UTC` (plus `workflow_dispatch`) instead of on every `push` / `pull_request`. PR runs have been failing too often because they track upstream Go stdlib advisories that arrive between Go patch releases — creating noise on unrelated PRs and blocking merges on issues unrelated to the change. The nightly job still surfaces new advisories.

### 3. Add Dependabot config

Adds `.github/dependabot.yml` to automate the recurring third-party dep bumps (`golang.org/x/net`, `golang.org/x/crypto`, etc.) that govulncheck has historically required us to do by hand. Covers `gomod` (root + `cmd/keeper`), `docker`, and `github-actions`. Runs **daily**, groups minor/patch updates to limit PR volume, and **ignores 0xPolygon-maintained forks** (cosmos-sdk, cometbft, bor, heimdall-v2, go-ethereum) which must be pinned by hand for consensus determinism per the security rules.

Caveat: Dependabot does **not** bump the Go `toolchain` directive itself, so stdlib CVEs like the two fixed in this PR will still need manual Go bumps surfaced by the nightly govulncheck job.

## Files touched
- `go.mod`, `cmd/keeper/go.mod` — `go 1.26.4`
- `Dockerfile`, `Dockerfile.alltools` — `golang:1.26.4-alpine`
- `.golangci.yml` — `go: '1.26.4'`
- `.github/workflows/govulncheck.yml` — schedule (`0 0 * * *`) + `workflow_dispatch`; drop `get-diff-action`, `push`, and `pull_request` triggers
- `.github/dependabot.yml` — new

## Test plan
- [ ] CI passes on this PR (govulncheck no longer runs on PRs, so its previous failure does not gate this)
- [ ] Manually trigger the new `Govuln` workflow via `workflow_dispatch` on `develop` after merge and confirm it reports clean
- [ ] Confirm Dependabot starts producing PRs within ~24h of merge


cmd/keeper/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ethereum/go-ethereum/cmd/keeper
 
-go 1.26.3
+go 1.26.4
```
manav2401, line 3: Can you run `go mod tidy` in `cmd/keeper` as the `go.sum` file also has some changes.

manav2401 (dismissed):
Small nit, else LGTM

lucca30 (dismissed):

.github/dependabot.yml
```diff
@@ -0,0 +1,58 @@
+version: 2
+updates:
+  - package-ecosystem: 'gomod'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+      day: 'monday'
```
marcello33, line 7: Should we do this daily too?

kamuikatsurgi: I think it would be too much noise. I can change it to daily if needed.

marcello33: nono, fine for me weekly

lucca30:
there is a pending conflict to address

PR #2259, opened 2026-06-04
eth/filters: honor context cancellation in bor block-log range scan

lucca30:
## What

`BorBlockLogsFilter.unindexedLogs` scans a block range one sprint at a time but
never checked the request context, so a request whose context was cancelled —
the client disconnected, or a deadline fired — kept scanning the full range and
held its RPC worker until the range was exhausted.

Upstream go-ethereum performs this check at the top of its `unindexedLogs` loop
(`eth/filters/filter.go`); the bor block-log filter omitted it.

## Change

Restore parity with a `select { case <-ctx.Done(): ... }` guard at the top of
the loop, so a range scan stops promptly once the caller is gone. With the
context now observed, any deadline attached upstream (e.g. an operator-configured
RPC timeout) also bounds the scan.

## Why

Without the check, a long range query keeps consuming a worker after its result
can no longer be delivered, reducing the RPC server's effective capacity under
cancelled or abandoned requests. The fix is small and matches the canonical
go-ethereum loop.

## Changes

- `eth/filters/bor_filter.go`: add the context-cancellation check to the
  `unindexedLogs` scan loop.
- `eth/filters/bor_filter_test.go`: add `TestBorFilterHonorsContextCancellation`
  and `TestBorFilterHonorsContextDeadline`.

## Testing

- `go test ./eth/filters/` — all pass (28).
- The new tests fail on the prior code (the loop ran 1001 block reads on an
  already-cancelled context and returned a nil error) and pass with the fix.

### DIFF
--- a/eth/filters/bor_filter.go
+++ b/eth/filters/bor_filter.go
@@ -128,6 +128,16 @@ func (f *BorBlockLogsFilter) unindexedLogs(ctx context.Context, end uint64) ([]*
 	sprintLength := f.borConfig.CalculateSprint(uint64(f.begin))
 
 	for ; f.begin <= int64(end); f.begin = f.begin + int64(sprintLength) {
+		// Honor context cancellation (client disconnect or deadline) so a large
+		// range scan stops promptly instead of holding an RPC worker busy with
+		// work whose result the caller will never read. Mirrors the check in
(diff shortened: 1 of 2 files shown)


lucca30:
@claude review

cffls (approved):

manav2401 (approved):

PR #1931, opened 2025-12-12
chore: drop legacy // +build comment

findfluctuate:
# Description

From Go 1.17, the preferred syntax for build constraints is `//go:build`,
which replaces the old `// +build` form. The old style is now considered
deprecated but still supported for backward compatibility.

This change removes the obsolete `// +build xxx` line, keeping only the
modern `//go:build xxx` directive.

More info: https://github.com/golang/go/issues/41184 and https://go.dev/doc/go1.17#build-lines

Design Doc / Proposal：
https://go.dev/design/draft-gobuild

# Changes

- [ ] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)
- [ ] Changes only for a subset of nodes

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Nodes audience

In case this PR includes changes that must be applied only to a subset of nodes, please specify how you handled it (e.g. by adding a flag with a default value...)

# Checklist

- [ ] I have added at least 2 reviewer or the whole pos-v1 team
- [ ] I have added sufficient documentation in code
- [ ] I will be resolving comments - if any - by pushing each fix in a separate commit and linking the commit hash in the comment reply
- [ ] Created a task in Jira and informed the team for implementation in Erigon client (if applicable)
- [ ] Includes RPC methods changes, and the Notion documentation has been updated

# Cross repository changes

- [ ] This PR requires changes to heimdall
    - In case link the PR here:
- [ ] This PR requires changes to matic-cli
    - In case link the PR here:

## Testing

- [ ] I have added unit tests
- [ ] I have added tests to CI
- [ ] I have tested this code manually on local environment
- [ ] I have tested this code manually on remote devnet using express-cli
- [ ] I have tested this code manually on amoy
- [ ] I have created new e2e tests into express-cli

### Manual tests

Please complete this section with the steps you performed if you ran manual tests for this functionality, otherwise delete it

# Additional comments

Please post additional comments in this section if you have them, otherwise delete it

PR #2260, opened 2026-06-04
chore: drop legacy // +build comment

solunolab:
## Summary

From Go 1.17, the preferred syntax for build constraints is //go:build,
which replaces the old // +build form. The old style is now considered
deprecated but still supported for backward compatibility.

This change removes the obsolete // +build xxx line, keeping only the
modern //go:build xxx directive.

More info: https://github.com/golang/go/issues/41184 and https://go.dev/doc/go1.17#build-lines

Design Doc / Proposal：
https://go.dev/design/draft-gobuild

Inspired by https://github.com/0xPolygon/bor/pull/1931

## Executed tests
<what was actually run beyond CI's standard unit / integration / e2e gates: kurtosis scenarios, chaos runs, manual checks against Amoy / mainnet RPCs, devnet upgrades, etc. Include output or pointers to where the run lives.>

## Rollout notes
<consensus-affecting? requires coordinated upgrade? backwards-compatible? operator-facing change?>

### DIFF
--- a/tests/bor/bor_config_change_test.go
+++ b/tests/bor/bor_config_change_test.go
@@ -1,5 +1,4 @@
 //go:build integration
-// +build integration
 
 package bor
 
--- a/tests/bor/bor_test.go
+++ b/tests/bor/bor_test.go
@@ -1,5 +1,4 @@
 //go:build integration
-// +build integration
 
 package bor
 
--- a/tests/bor/fee_receiver_change_test.go
+++ b/tests/bor/fee_receiver_change_test.go
@@ -1,5 +1,4 @@
 //go:build integration
-// +build integration
 
 package bor
 


solunolab:
This failure does not seem related to this PR.

The PR only removes legacy `// +build integration` lines from `tests/bor/*` files, while the failing test is in `consensus/bor/heimdall/failover_client_test.go`.

I checked the failing test locally on this branch:

- `go test ./consensus/bor/heimdall -run TestFailover_NoSwitchOnShutdownDetected -count=1 -v`
- `go test ./consensus/bor/heimdall -run TestFailover_NoSwitchOnShutdownDetected -count=50`

Both passed.

The failure looks like a flaky race in the test itself: `secondary.hits` is also incremented by the background health probe via `FetchStatus`, not only by an actual failover call. So the observed `actual: 1` can happen if the immediate health-check probe touches the secondary before the assertion.

I think rerunning CI should be enough here. If you prefer, I can also open a separate follow-up PR to make this test avoid counting background probe hits as failover attempts.


@pratikspatil024 @kamuikatsurgi Hi, Would you mind review this PR at your convenience? Thank you very much.

daixihegu:
I made the same observation, but I saw that it has already been submitted.
Good fixes.  @pratikspatil024 @kamuikatsurgi

solunolab:
@pratikspatil024 @marcello33 @manav2401 @vbhattaccmu Hi, Could you please review this PR at your convenience? Thank you very much.

PR #2268, opened 2026-06-09
Different results blockstm vs serial

alrevuelta:
## Summary

A transaction involving a selfdestruct + a simple native transfer makes **BlockSTM V2 execution differ from serial execution**. Both state roots are identical, but the receipt hash of the BlockSTM V2 differs.

The receipt hash differs because a **wrong log is emitted in the BlockSTM V2 case**. This is the Bor specific transfer (LogTransfer event) emitted for native value transfers. Note that the bloom matches, because log's data fields are not used for it.

With the two contracts and transaction of the failing test (see code) serial and BlockSTM V2 produce different results. Exactly this log is different:

* Serial -> LogTransfer: token=0x...1010 from=0x...aaaa to=0x...BbBB amount=10 input1=10 **input2=10** output1=0 **output2=20**
* BlockSTM V2 -> LogTransfer: token=0x...1010 from=0x...aaaa to=0x...BbBB amount=10 input1=10 **input2=0** output1=0 **output2=10**

Run:
```
go test ./core/ -run TestV2_SelfDestructTransferLog_MispairsWithSerial -v
```

Note that this PR is just a test not fixing the issue but it proves a mismatch between sequential execution and BlockSTM V2 one. Reported in the open since this can't be exploited, as nodes run both sequential and BlockSTM V2 execution. So in the worst case, it will fallback to sequential ignoring the BlockSTM V2.

### DIFF
--- a/core/v2_selfdestruct_transfer_log_mispair_test.go
+++ b/core/v2_selfdestruct_transfer_log_mispair_test.go
@@ -1,0 +1,176 @@
+package core
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+	"strings"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/blockstm"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/tracing"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/params"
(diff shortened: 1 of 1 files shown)


core/v2_selfdestruct_transfer_log_mispair_test.go
```diff
@@ -0,0 +1,176 @@
...
+	t.Logf("serial LogTransfer: %s", serialLog)
+	t.Logf("v2     LogTransfer: %s", v2Log)
+
+	if serialLog.data() != v2Log.data() {
+		t.Errorf("V2 BlockSTM receipt diverges from serial for the same transaction")
+	}
+}
+
+// Bor's native LogTransfer event
+const logTransferABI = `[{"anonymous":false,"name":"LogTransfer","type":"event","inputs":[
```
alrevuelta, line 129: all the code behind can be removed. i just wanted this to decode the logs and have a nice readable output. not really needed for the tes.

cffls:
Thanks @alrevuelta for finding this bug. Cherry-picked this test and addressed the bug in https://github.com/0xPolygon/bor/pull/2270.

alrevuelta:
thanks @cffls glad it helped!

PR #2269, opened 2026-06-10
txpool/legacypool: fix fillPool TOCTOU and remove debug prints

lucca30:
## Summary

- **Flaky test fix**: `fillPool` in `legacypool2_test.go` read `pool.Stats()` and `pool.all.Slots()` in two separate critical sections (`pool.mu.RLock` and `t.lock.RLock`). A concurrent pool mutation between the two reads could produce an inconsistent snapshot, causing `TestLockOrdering_ReplacePendingNoDeadlock` to fail with `have 199, want 200`.
- **Fix**: hold `pool.mu.RLock` across both reads in `fillPool`, using the unexported `pool.stats()` (valid since the test is in the same package). All paths that mutate `pool.all` require `pool.mu.Lock`, so the two reads are now atomic with respect to pool mutations.
- **Debug print removal**: three leftover `fmt.Println` calls in `list_test.go` (`TestFilterTxConditionalKnownAccounts`) polluted CI output; replaced with blank identifier assignments and removed the unused `"fmt"` import.

## Test plan

- [ ] `go test ./core/txpool/legacypool/... -count=50 -run TestLockOrdering_ReplacePendingNoDeadlock` — no failures expected
- [ ] `go test ./core/txpool/legacypool/... -count=1` — full package passes

### DIFF
--- a/core/txpool/legacypool/legacypool2_test.go
+++ b/core/txpool/legacypool/legacypool2_test.go
@@ -66,8 +66,15 @@ func fillPool(t testing.TB, pool *LegacyPool) {
 	}
 	// Import the batch and verify the pool is full and executable.
 	pool.addRemotesSync(executableTxs)
-	pending, queued := pool.Stats()
+
+	// Read pending, queued, and all.Slots() under pool.mu.RLock so the three
+	// values are consistent: no concurrent mutation can interleave between the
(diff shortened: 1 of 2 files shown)


lucca30:
@claude review

cffls (approved):

manav2401 (approved):

PR #2277, opened 2026-06-17
feat: embed commit-sealing timings in extra data

minhd-vu:
## Summary

Embed the producer's commit-sealing timings -- the same elapsed and finalize values logged by "Commit new sealing work" -- into the block's BlockExtraData, mirroring the existing TimeNano feature so the values can be read directly from the chain instead of scraped from logs.

- Add SealElapsedNano and SealFinalizeNano optional fields to BlockExtraData, plus GetSealTimings getter and SetSealTimings setter.
- Capture and embed the timings in miner commit() after FinalizeAndAssemble, gated on the Placeholder fork, rebuilding the block before sealing. The existing log line is left unchanged.
- Presence-validate both fields in verifyHeader (errMissingSealTimings); values are per-producer and non-deterministic so only presence is checked, as with TimeNano and the Giugliano fields.

## Executed tests
<what was actually run beyond CI's standard unit / integration / e2e gates: kurtosis scenarios, chaos runs, manual checks against Amoy / mainnet RPCs, devnet upgrades, etc. Include output or pointers to where the run lives.>

## Rollout notes
<consensus-affecting? requires coordinated upgrade? backwards-compatible? operator-facing change?>


pratikspatil024:
codegenie review

PR #2287, opened 2026-06-24
miner: recover veblop producer stall when a build wedges with no pending task

lucca30:
## Problem

While stress-testing the veblop block-production path, a producer can enter a permanent stall: it stops producing/importing at a fixed height and recovers only after a process restart. Symptoms are `Possible producer stall: veblop fallback skipping while chain is stale` logged repeatedly with a static `currentBlock` and `pendingTasksCount=0` — the producer's own stall detector fires, but the fallback never actually recovers.

## Root cause

`newWorkLoop`'s veblop stall-fallback skipped recovery whenever `pendingWorkBlock == currentBlock+1`, treating that value as proof the next block was being built. But `commitWork` only clears `pendingWorkBlock` via a `defer ...Store(0)` that runs **on return**. If a build blocks inside `commitWork` and never returns, `pendingWorkBlock` stays pinned at `currentBlock+1` with **no sealing task** (`pendingTasks` empty). In that state the fallback short-circuited forever — the producer never resubmitted work, and the only escape was a restart.

(A build blocking inside `commitWork` is a separate concern being investigated independently. This PR addresses the recovery path so that such a hang is no longer permanent.)

## Fix

- The fallback now treats `pendingWorkBlock == next` as "work in flight" **only when a sealing task actually backs it**. When `pendingWorkBlock == next` but no task exists, it distinguishes a *wedged* build from a *slow-but-live* one by **how long the build has been outstanding** (time since submit), not by block-timestamp age (`chainAge` is meaningless when the head carries an old timestamp, e.g. genesis). Recovery fires only past a 3×-blocktime threshold, so a live build is never interrupted.
- The fallback re-commit uses a **non-blocking** send, so `newWorkLoop` cannot itself wedge on the unbuffered `newWorkCh` when `mainLoop` is the blocked goroutine. It keeps emitting the stall warning and retrying every tick, and auto-recovers the moment `mainLoop` is unblocked.
- The decision is extracted into a pure `decideVeblopFallback` function so it can be unit-tested without driving the whole loop.

## Testing

- New table-driven unit test `TestDecideVeblopFallback` covering: the wedged-build case (the bug), the genesis in-progress-build case (regression guard against interrupting a live build), task-in-flight, the no-work-claimed paths, and the wedge threshold boundaries.
- Verified as a true regression test: the suite goes **red** against the pre-fix logic (the wedged-build cases return "skip" instead of "recommit") and **green** against this fix.
- Full `miner/` package suite passes (this caught an over-aggressive first attempt that used `chainAge` and interrupted live builds at genesis).

## Scope / follow-ups

This restores producer-loop liveness and auto-recovery once the underlying build unblocks. It does **not** by itself resolve a hard `mainLoop` deadlock; that underlying build-blocking concern is tracked separately.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -754,12 +754,18 @@ const (
 // alone, so case (b) never recovered without a process restart.
 //
 // These two cases are distinguished ONLY by how long the build has been
-// outstanding (pendingWorkAgeSec = time since the build was submitted), NOT by
-// block-timestamp age (chainAgeSec is meaningless when the head carries an old
+// outstanding (pendingWorkAge = time since the build was submitted), NOT by
+// block-timestamp age (chainAge is meaningless when the head carries an old
 // timestamp, e.g. at genesis). We recover only once the outstanding build clearly
-// exceeds a normal build duration (stallThresholdSec), so a slow-but-live build is
+// exceeds a normal build duration (stallThreshold), so a slow-but-live build is
 // never interrupted.
-func decideVeblopFallback(pendingWorkBlock, nextBlock uint64, hasPendingTasks bool, chainAgeSec, veblopTimeoutSec, pendingWorkAgeSec, stallThresholdSec int64) veblopFallbackDecision {
+//
+// All durations are compared at their native resolution: pendingWorkAge is real
+// elapsed wall-clock (sub-second precise), so the threshold stays accurate at the
+// sub-second block times mainnet runs (miner block time is 1.5s). chainAge is the
+// exception — it derives from integer block timestamps, so it is inherently
+// second-granular regardless of representation.
+func decideVeblopFallback(pendingWorkBlock, nextBlock uint64, hasPendingTasks bool, chainAge, veblopTimeout, pendingWorkAge, stallThreshold time.Duration) veblopFallbackDecision {
 	// A sealing task is in flight; never interrupt it.
 	if hasPendingTasks {
 		return veblopSkip
@@ -767,13 +773,13 @@ func decideVeblopFallback(pendingWorkBlock, nextBlock uint64, hasPendingTasks bo
 	if pendingWorkBlock == nextBlock {
 		// Build outstanding for the next block. Recover only if it has been
 		// outstanding long enough to be considered wedged rather than in-progress.
-		if pendingWorkAgeSec >= stallThresholdSec {
+		if pendingWorkAge >= stallThreshold {
 			return veblopRecommit
 		}
 		return veblopWait
 	}
 	// Nothing is claimed for the next block — resubmit once the chain is stale.
-	if chainAgeSec >= veblopTimeoutSec {
+	if chainAge >= veblopTimeout {
 		return veblopRecommit
 	}
 	return veblopWait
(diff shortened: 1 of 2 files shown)


lucca30:
@claude review

lucca30:
On the two red checks — deliberately leaving the code as-is:

**Quality metrics (diffguard complexity)** — advisory, not a merge gate. The flagged function is `newWorkLoop`, which is already a known `//nolint:gocognit` giant (base cognitive complexity 41, well over the 15 threshold before this PR). This change adds +13, but the right structural move was already made: the fallback decision is extracted into the pure, unit-tested `decideVeblopFallback`. The remaining complexity lives in the `veblopTimer.C` case body, which is tightly coupled to the `commit` closure (captured `interrupt`/`timestamp`/`timer` locals). Extracting it further would mean reshaping timing-sensitive block-producer recovery code purely to satisfy an advisory metric — a poor risk/reward trade on the liveness path. For reference, the check is non-blocking: #2296 merged with Quality metrics failing.

**codecov/patch (86.36% vs 90%)** — also advisory. The uncovered lines are the non-blocking-send-failed branch and the `veblopTimer.Reset` path, which can't be exercised without driving the full `newWorkLoop` goroutine; the decision logic itself is covered by the table-driven `decideVeblopFallback` tests.

The meaningful CI (build / unit / integration / e2e / lint) is green. This is ready for a human review of the design choices the bot flagged (3× block-time threshold, non-blocking send, `pendingWorkSubmittedAt` tracking).

miner/worker.go
```diff
@@ -819,19 +902,51 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+
+			veblopTimeoutSec := int64(veblopTimeout.Seconds())
+			if veblopTimeoutSec < 1 {
+				veblopTimeoutSec = 1
+			}
+			// A build is treated as wedged only after it has been outstanding for
+			// noticeably longer than a normal build (matches warnIfStalled's
+			// 3x-timeout stale threshold), so a slow-but-live build is never
+			// interrupted. Measured from submit time, not block-timestamp age.
+			stallThresholdSec := 3 * veblopTimeoutSec
```
cffls: What's the reason for 3x veblopTimeout? Will there be any potential issue if we use 1x?

marcello33 (approved):

miner/worker.go
lucca30: The 3x is anchored to two things — the span-rotation window (the reason for the *value*) and the existing staleness definition (the reason it's *this* multiple, not a new one).

**Rotation window (load-bearing).** Mainnet producers run `miner.blocktime = 1.5s`, so `3x = 4.5s`. That sits deliberately below veblop's rotation clock — `veblopBlockTimeout = 8s` (the new-span check) and `waitForNewSpan = 16s`. The producer gets ~4.5s to self-recover locally *before* the network rotates production to the next span's producer. Recovery is single-producer-local (no wiggle/backup in veblop), so it never competes with rotation; and if the node truly can't recover, the recommits fail seal-authorization once the span moves — no fork, rotation just takes over. Going higher would be the risky direction: at 1.5s a 5x = 7.5s starts crowding the 8s rotation wall.

**Same staleness definition.** `warnIfStalled` flags "stale" at `chainAge > 3x block period`, and recovery reuses that same 3x rather than inventing a second threshold, so we act on the same notion of "stuck" we already warn on. (They apply the 3x to different clocks: the warn measures chain-age — time since the head advanced; the wedge-recovery measures build-outstanding age — time since we submitted the build. Same multiple, the appropriate signal for each.)

**Why not 1x.** Builds legitimately overrun a single period under transient load (heavy block, GC pause, disk I/O, txpool-lock contention). At 1x those get interrupted and re-run from scratch — churn that can cause the very misses it's meant to prevent. 3x is the smallest multiple that clears normal jitter while staying under the rotation window.

One follow-up your question surfaced: this was silently computing the threshold in whole seconds, so at 1.5s it truncated `3x` to 3s (and would degrade to 6x a hypothetical 0.5s block). I pushed a follow-up commit (f52c1354a) moving both this recovery path and `warnIfStalled` to real `time.Duration`, so 3x honors the 1.5s block time exactly. Thanks for the prompt.

miner/worker.go
```diff
@@ -755,24 +814,46 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+	// When nonblocking is true the request is sent with a non-blocking send: if
+	// mainLoop is not currently ready to receive (e.g. it is wedged inside
+	// commitWork — exactly the state that pins pendingWorkBlock with no pending
+	// task), the request is dropped and commit returns false instead of blocking
+	// newWorkLoop on the unbuffered newWorkCh. This keeps newWorkLoop alive so it
+	// keeps emitting the stall warning and retrying every tick.
+	commit := func(noempty bool, s int32, nonblocking bool) bool {
 		if interrupt != nil {
 			interrupt.Store(s)
 		}
```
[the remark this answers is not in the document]
lucca30: Good observation on the ordering — `interrupt.Store(s)` does run before the send and isn't rolled back if the non-blocking send hits `default`. But it doesn't cause the harm described, because of the state this path runs in.

The non-blocking send is only reached from the `veblopRecommit` branch. For a claimed-but-taskless build (`pendingWorkBlock == next && !hasPendingTasks`) that requires the build to have been outstanding >= 3x the block period. So the `interrupt` being signaled is only ever one of:

1. a genuinely wedged build — and signaling it to abort is the *intended* behavior: when it finally unblocks inside `commitWork` we don't want it sealing a block built against stale state, we want it to bail so the fresh recommit wins. Setting the interrupt even when the channel isn't ready is deliberate for exactly this case; or
2. in the "nothing claimed" recommit branch (`pendingWorkBlock != next`), a build that already returned and cleared `pendingWorkBlock` — so `interrupt` references a completed build and the store is a no-op.

A healthy, live build can't reach this path: mid-execution it either has a pending task (→ `veblopSkip`) or is under the 3x threshold (→ `veblopWait`). So we never interrupt a build that would otherwise have succeeded.

On churn: the dropped-send path doesn't lose or leak a request — `newInterrupt` is local and only promoted to `interrupt` on a successful send; the old interrupt is re-stored idempotently each tick. Recovery retries on the next veblop tick and self-corrects the moment `mainLoop` is ready again, so the extra delay is bounded to one tick and only occurs while `mainLoop` genuinely isn't receiving.

Net: keeping the store before the send is intentional for the wedged case; moving it after the send would let a wedged build seal a stale block when it unblocks. Leaving as-is.

cffls (approved):

PR #2288, opened 2026-06-25
core/rawdb: avoid the freezer read-lock for recent header lookups

marcello33:
## Summary

`ReadCanonicalHash`, `ReadHeaderRLP`, and `ReadTdRLP` wrap every lookup in `db.ReadAncients`, which acquires the freezer's `writeLock.RLock`. A freeze cycle (`ModifyAncients`) holds the exclusive `writeLock` while it writes a batch to the ancient store and fsyncs it — so during a freeze, every header/canonical-hash/TD read blocks until it finishes.

Recent blocks are never in the freezer, so this reads the key-value store first and only falls through to the (unchanged) ancient path on a miss. Recent lookups no longer touch the freezer lock.

`chainFreezer.freeze` writes a block to the ancient store before deleting it from the key-value store, so a KV miss is always covered by the ancient read — there is no window where a block is in neither. The locked ancient path is unchanged; only an unlocked KV fast path is added in front.

On a Polygon validator, the Heimdall milestone-proposition path (`GetBorChainBlockInfoInBatch` → `HeaderByNumber` → `ReadAncients`) hit its bor RPC deadline whenever a freeze held the lock, so the vote extension shipped without a milestone proposition.

### Contention

Before
01-contention-before

After
02-contention-after

Executed tests

- go test -race ./core/rawdb/ — full package green.
- `TestRecentReadsBypassFreezerLock` — reproduces the contention: holds the freezer writeLock and asserts recent reads still return. Fails on `develop` (read blocks, 5s timeout); passes with this change.
- `TestReadCanonicalHashAncientFallback` — confirms the ancient fallback still resolves frozen entries absent from the KV store.
- `diffguard` mutation testing: 100% (9/9 killed), Tier-1 logic 100%; all gates PASS.

Production verification (Polygon mainnet validator, 14h+ soak)

pprof on the running node + Heimdall logs, before vs after deploy:

03-metrics-before-after

Rollout notes

- Not consensus-affecting. Read-path only; the ancient fallback and all write paths are unchanged, and return values are identical to before.
- Backwards-compatible. No DB format, schema, or wire changes. No coordinated upgrade required.
- Not operator-facing. No config or flag changes.

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -34,6 +34,13 @@ import (
 
 // ReadCanonicalHash retrieves the hash assigned to a canonical block number.
 func ReadCanonicalHash(db ethdb.Reader, number uint64) common.Hash {
+	// Fast path: recent entries live in the key-value store and are removed only
+	// after being copied to the freezer (chainFreezer.freeze), so a KV hit is
+	// always valid. Reading KV first lets recent lookups skip the freezer
+	// read-lock and avoid serializing behind an in-progress freeze.
+	if data, _ := db.Get(headerHashKey(number)); len(data) > 0 {
+		return common.BytesToHash(data)
+	}
 	var data []byte
 	db.ReadAncients(func(reader ethdb.AncientReaderOp) error {
 		data, _ = reader.Ancient(ChainFreezerHashTable, number)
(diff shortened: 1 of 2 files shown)


marcello33:
@claude review once

core/rawdb/accessors_chain_test.go
```diff
@@ -551,6 +553,97 @@ func TestAncientStorage(t *testing.T) {
...
+
+	go func() {
+		_, _ = db.ModifyAncients(func(ethdb.AncientWriteOp) error {
+			close(held)
+			<-release
+			return nil
+		})
+	}()
+	<-held // ensure the write-lock is actually held before reading
+
```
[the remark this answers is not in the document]
marcello33: addressed in latest commit

core/rawdb/accessors_chain.go
```diff
@@ -34,6 +34,13 @@ import (
 
 // ReadCanonicalHash retrieves the hash assigned to a canonical block number.
 func ReadCanonicalHash(db ethdb.Reader, number uint64) common.Hash {
+	// Fast path: recent entries live in the key-value store and are removed only
+	// after being copied to the freezer (chainFreezer.freeze), so a KV hit is
+	// always valid. Reading KV first lets recent lookups skip the freezer
+	// read-lock and avoid serializing behind an in-progress freeze.
+	if data, _ := db.Get(headerHashKey(number)); len(data) > 0 {
+		return common.BytesToHash(data)
+	}
```
kamuikatsurgi: The comments and code is repeatitive. We could extract this into a helper func and use it.

kamuikatsurgi (dismissed):

manav2401 (dismissed):

core/rawdb/accessors_chain.go
marcello33: Addressed

manav2401 (dismissed):

kamuikatsurgi (dismissed):

PR #2290, opened 2026-06-25
core/state, core: exact-writer validation for v2 destruction-resolved reads

cffls:
## Summary

BlockSTM v2's read-validation (`storeReadMatches`) accepted a recorded read whenever the current MVStore value equaled the recorded value, even when the **winning writer's block-order position** changed. For reads resolved relative to a prior `SELFDESTRUCT` (the `writerIdx > suicideIdx` pattern) and for the `CreatePath`/`SuicidePath` existence markers, the writer's *position* — not its value — determines the result. A reordered metamorphic `CREATE2` + same-tx `SELFDESTRUCT` could therefore move a code/storage/marker writer across the destruction boundary while its value stayed equal; validation wrongly skipped re-execution and v2 settled a **state root that differed from serial**.

Fix (first commit): a per-read `ExactWriter` flag on `StoreReadDesc` that disables the value-equality fallback for

- `CreatePath`/`SuicidePath` markers — always (their value is always `true`, so value-equality can only ever mask a reorder), and
- `Nonce`/`Code`/`Storage` reads when a prior destruction is in play (`suicideIdx >= 0`).

The common no-destruction path keeps the value-equality optimization, so the cost is bounded to the rare destruction-interleaved case. Marker reads now record the real writer incarnation. This covers all six destruction-resolved read sites (`Exist`, `GetNonce`, `GetCode`, `GetCodeHash`, `GetState`, `GetCommittedState`), so the storage- and nonce-slot twins of the code-path symptom are closed by construction rather than chased individually.

Test coverage (second commit): the old differential vocabulary could not express `CREATE2`, metamorphic re-create at a destroyed address, or a reorderable deploy decision, so the divergence was unreachable by the fuzzer. The grammar now adds CREATE2 deploy / deploy+destroy-in-one-tx (EIP-6780) / separate-tx destroy / re-create / `EXTCODEHASH`·`EXTCODESIZE`·`EXTCODECOPY`·`BALANCE`·`SELFBALANCE`·`SLOAD` reads of the target, a flag-conditional deployer (reorderable salt), `TLOAD`/`TSTORE`, `DELEGATECALL`, SSTORE clear-refund, contract `LOG`, a `GAS` canary, 7702 delegation to the metamorphic target, a reverting tx (journal rollback), and legacy / EIP-2930 / EIP-1559 envelopes.

## Executed tests

| Check | Result |
|---|---|
| `TestAllBlocksConsistency` (`BOR_BLOCKSTM_TEST=1`, 241 real mainnet blocks, v2-parallel vs serial) | **241/241 consistent, 0 failures** |
| `FuzzV2ExecutorVsSerial` over the expanded grammar (post-fix) | 1,016,614 execs, 0 divergences (pre-fix it finds the divergence in ~19s) |
| `TestV2SerialParity_MetamorphicCreate2` (new deterministic regression) | catches pre-fix (5/5), passes post-fix |
| `TestMetamorphicHarnessSemantics`, `TestV2ExecutorVsSerial_SeedCorpus` (worker grid `{1,2,3,4,6,8}`), `core/state` unit suite, existing v2 selfdestruct/balance/gas-determinism tests | green |

Standard CI gates (lint, unit/integration, diffguard w/ mutation, govulncheck, kurtosis e2e) still to run on this PR.

## Rollout notes

- **Consensus-affecting for BlockSTM v2 only.** v2 is not present in any released version (absent from `v2.8.3`); the default configuration runs v2 non-enforced with serial as the source of truth and validates the v2 root against the block header, falling back to serial on mismatch. There is therefore **no behavior change for current production nodes** and **no coordinated upgrade required** — this hardens v2 before it ships as a primary execution path.
- Not operator-facing. No config, genesis, fork-activation, or precompile changes; backward compatible.
- Cross-client parity (Erigon) is unaffected: serial execution output is unchanged; this only changes when v2 re-executes a tx internally.

### DIFF
--- a/core/state/parallel_read_invariants_test.go
+++ b/core/state/parallel_read_invariants_test.go
@@ -155,3 +155,122 @@ func TestParallelReadRecordingCompleteness(t *testing.T) {
 		})
 	}
 }
+
+// TestParallelExactWriterAcrossDestruction pins the fix for value-equality
+// masking a reordered writer across a destruction boundary. For a read resolved
+// relative to a prior SELFDESTRUCT (writerIdx vs suicideIdx) and for the
+// existence markers, the winning writer's *position* — not just its value —
+// determines the result. So when the winning writer changes to one producing an
+// EQUAL value, validation must still fail; accepting it by value equality
+// settles a stale read and diverges from serial.
+//
+// Each case stages the metamorphic shape at the MVStore level: tx0 creates the
+// account and SELFDESTRUCTs it (write at idx 0 + SuicidePath at idx 0), tx2
+// re-creates it with an equal value, the read observes tx2's value, then tx2's
+// write is withdrawn (re-execution) so the winner falls back to tx0 — which is
+// at/below the destruction and therefore wiped. The recorded read must be
+// rejected. The destructor sits at tx index 0 on purpose: that is the exact
+// suicideIdx>=0 boundary, so a `>0` regression (which would drop ExactWriter
+// when the destruction is the first tx) is caught here too.
+func TestParallelExactWriterAcrossDestruction(t *testing.T) {
+	addr := common.HexToAddress("0x00000000000000000000000000000000000000e0")
+	slot := common.HexToHash("0x01")
+	codeKey := blockstm.NewSubpathKey(addr, CodePath)
+	nonceKey := blockstm.NewSubpathKey(addr, NoncePath)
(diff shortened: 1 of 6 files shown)


cffls:
@claude review

pratikspatil024 (approved):

lucca30 (approved):

PR #2293, opened 2026-06-30
core: exempt empty no-op blocks from sidechain ghost-state rejection

### LINKED ISSUE #2224: Bor sync wedges on empty block after producer restart — insertSideChain same-stateRoot false-positive
# Bor sync wedges on empty block after producer restart — `insertSideChain` same-stateRoot false-positive

## Summary

After restarting a Bor validator whose local head is an empty (no-transaction) block produced just before the restart, Bor can refuse to sync past that block, dropping every peer with:

    WARN Sidechain ghost-state attack detected   number=N sideroot=X canonroot=X
    WARN Synchronisation failed, dropping peer   err="retrieved hash chain is invalid: sidechain ghost-state attack" mode=full

`sideroot` and `canonroot` are **the same value** — the check at `core/blockchain.go:3562` is firing on identical state roots. The upstream-inherited logic assumes same-root + different-hash implies a shadow-state attack, which is only true when the block carried a state transition. For empty blocks (gasUsed=0, transactions=\[\]) the state root is just the parent's, so two distinct empty headers at the same height legitimately share a state root while having different hashes/seals.

A node in this state cannot recover by reconnecting, restarting Bor, or by `debug_setHead` to a block immediately behind the offending height — the side-block record persists in the database and re-triggers the same check on the next sync attempt.

## Environment

- Bor `v2.8.0-beta`
- Devnet spawned via `kurtosis-pos`, 9 L2 nodes:
  - v1–v5: bor + heimdall-v2 validators
  - v6–v8: bor RPC nodes
  - v9: erigon RPC node
- Heimdall span at the time of the incident (span id 11, blocks 1408–1535) had a single entry in `selected_producers` (val_id 3). Small validator set, weighted span selection — one validator can end up sole producer for a span.

## Reproduction

1.  Spawn a devnet large enough that some spans have a single `selected_producers` entry (the kurtosis-pos default tends to produce this with small validator counts).
2.  Wait for an active span where one validator (call it `vP`) is the sole producer.
3.  Stop `vP`'s bor EL and a second validator's bor EL (`vS`) within a second of each other, with `vS`'s last canonical head being an empty block recently produced by `vP`.
4.  After a few seconds, start both back up.

Expected: `vS` resyncs to chain head.

Observed: `vS` stays stuck at its last pre-stop block (block number `N`). `docker logs` for `vS`'s bor container streams the warn-drop pair above on every peer it talks to, plus continuous `"Whitelisting milestone deferred err=chain out of sync"` from the heimdall ws subscription.

## Concrete trace from our run

Captured artifacts: `commands.out`, `chaos.out.txt`, `reports/test-20260513-131241-test-1778692361/`. Key timeline:

| Time (EDT) | Event |
|----|----|
| 13:19:04 | Stop `l2-el-2-bor-heimdall-v2-validator` (v2). v2's local head: block 1441, produced by v3, stateRoot `0x8f58ec…2dcc11`, `transactions: []`, `gasUsed: 0`. |
| 13:19:11 | Stop `l2-el-3-bor-heimdall-v2-validator` (v3). v3 was the sole producer for span 11. |
| 13:22:31 | Start v2. |
| 13:22:37 | Start v3. v3 resumes producing 1442, 1443, … |
| 13:34:41 onward | v2 logs `Sidechain ghost-state attack detected number=1441 sideroot=8f58ec..2dcc11 canonroot=8f58ec..2dcc11` against every peer; drops them; stays stuck at 1441 while v1/v3/v4/v5 advance past 2100. |

Block 1441 on v2 (its local canonical):

    number       1441
    hash         0xa8397a7be178763f882deb7f893b28c326cc4ddf038296a073cc5e6ea597826e
    parentHash   0xb1ba2c9854ea1d575173714ebcb730e0e4c6f385702b6cd81a00250d38a69f69
    stateRoot    0x8f58ec83616416cdae4e71306e5ef9f7482facf9328957b8c28d6468332dcc11
    transactions []
    gasUsed      0
    extraData    0x626f722d3300…   ("bor-3" vanity, then seal)

The 1441 the rest of the cluster has at the same height shares `parentHash`, `stateRoot`, `transactionsRoot`, `receiptsRoot` — the state didn't change — but has a different `timestamp` and `extraData` seal (likely produced by a succession-1+ backup after v3 stopped), so its block hash differs.

## Root cause

`core/blockchain.go`, `insertSideChain` (current `develop`):

``` go
// blockchain.go:3534
func (bc *BlockChain) insertSideChain(block *types.Block, it *insertIterator, makeWitness bool) (*stateless.Witness, int, error) {
    ...
    for ; block != nil && errors.Is(err, consensus.ErrPrunedAncestor); block, err = it.next() {
        headers = append(headers, block.Header())
        if number := block.NumberU64(); current.Number.Uint64() >= number {
            canonical := bc.GetBlockByNumber(number)
            if canonical != nil && canonical.Hash() == block.Hash() {
                // re-import of a canon block, fine
                continue
            }
            if canonical != nil && canonical.Root() == block.Root() {       // <-- false positive
                log.Warn("Sidechain ghost-state attack detected", "number", block.NumberU64(),
                    "sideroot", block.Root(), "canonroot", canonical.Root())
                return nil, it.index, errors.New("sidechain ghost-state attack")
            }
        }
        ...
    }
```

This is upstream go-ethereum's pre-merge defense against attackers side-mining to a height where state was pruned and substituting their block to bypass full state verification. The premise is that a benign sidechain block at height `N` would produce a *different* state root from the canonical chain (because the txs at heights `<= N` differ across the two chains).

That premise doesn't hold for Bor:

- Bor produces empty blocks (no txs) during quiet periods. For an empty block, the post-state equals the pre-state — `block.Root() == parent.Root()`.
- Bor's sprint/succession model lets a backup producer sign a block at the same height as the primary if the primary is missing. Two empty headers sharing a parent and produced by different signers — or even the same signer with a different timestamp — share a state root while having different hashes/seals.

So `same-root + different-hash` is a normal, expected condition in Bor, not an attack signal. The check turns it into a hard sync failure that also blacklists the offering peer for that sync attempt; with every honest peer in the network offering the canonical chain that disagrees with the node's local stale head, the node has no way out.

## Why `debug_setHead` to height `N-2` did not recover the node

We tried `debug_setHead 0x59f` (=1439) on v2. Logs confirm the rewind:

    WARN  Rewinding blockchain to block            target=1439
    INFO  Truncating from head                     type=state ohead=1442 tail=427 nhead=1441
    INFO  Rewound to block with state              number=1440 hash=b1ba2c..a69f69
    INFO  Truncating from head                     type=state ohead=1441 tail=427 nhead=1440
    INFO  Rewound to block with state              number=1439 hash=280067..874280
    INFO  Loaded most recent local block           number=1439 hash=280067..874280 td=2316 age=47m16s

But the warn-drop loop continued. Best guess: the side-block record for the old 1441 hash remains in the block database after `SetHead` (only the canonical pointer was rewound and state was truncated), and the downloader-side handling re-imports it as a side chain on the next sync attempt before reconciling with peers, so `insertSideChain` keeps hitting the same condition. We had to rewind well past the divergence to escape — a much larger setHead eventually let the node resync cleanly.

## Suggested directions

Cheapest fix: scope the check to non-trivial blocks. If the candidate block carries no state delta vs its parent, this isn't a shadow-state attack pattern.

``` go
if canonical != nil && canonical.Root() == block.Root() {
    // Genuine shadow-state attacks substitute a block whose state diverges
    // from the canonical state at the same height. Two empty blocks with
    // the same parent legitimately share a state root — that isn't an
    // attack, just a different seal/timestamp.
    if len(block.Transactions()) == 0 && block.GasUsed() == 0 &&
        block.ParentHash() == canonical.ParentHash() &&
        block.TxHash() == canonical.TxHash() &&
        block.ReceiptHash() == canonical.ReceiptHash() {
        // fall through to normal side-chain insertion
    } else {
        log.Warn("Sidechain ghost-state attack detected", ...)
        return nil, it.index, errors.New("sidechain ghost-state attack")
    }
}
```

Stronger fix: the upstream check is a pre-merge defense. Bor's consensus model (validator-signed headers, Heimdall-anchored milestones, checkpoint-based finality on L1) doesn't rely on this in-line root match to defend against state-pruning shadow attacks. It may be reasonable to drop the check entirely in `insertSideChain` for Bor and lean on existing Bor-specific verification (sealer authorization, sprint rules, milestone whitelist) instead. Worth a security review before doing this — flagging as a path the reviewer should weigh in on, not a recommendation.

Either way, the recovery path needs work too: even after `debug_setHead` puts the head behind the offending block, the side-chain record can re-trigger the same condition on the next sync. Investigating whether `SetHead` should also evict diverged side-blocks at heights `> target` would help operators recover without having to rewind a "large number" of blocks past the divergence.

## Workaround for operators

`debug_setHead` to a height well before the divergence (not just one or two blocks behind). RPC timeout on the call is normal for large rewinds; the operation continues server-side. After the rewind, expect a full resync from that height.

## Artifacts

- `commands.out` — full operator session capture (script(1) format, ANSI included)
- `chaos.out.txt` — chaos-runner stdout/stderr from the partition scenario
- `reports/test-20260513-131241-test-1778692361/` — scenario.yaml, report.json, container logs, prom snapshots
- `pos--683f2274028c459982665e777b2bcdc9/` — full Kurtosis enclave dump

bit2swaz:
## Summary

Fixes #2224 (the false-positive rejection). Recovery for an already-wedged node is tracked separately in #2386.

After a validator restart, a node could refuse to sync past an empty (zero-transaction) block, drop every peer with `sidechain ghost-state attack detected`, and need manual intervention to recover. In the logs `sideroot` and `canonroot` are the same value.

The cause is in `insertSideChain()` in `core/blockchain.go`. When a sidechain block hits a pruned ancestor, the code rejects any block whose state root equals the canonical block's state root at the same height, treating the match as a shadow-state attack (a block trying to reuse existing canonical state so it can skip re-execution of its ancestors). That assumption only holds for a block that carried a state transition. An empty block makes no state transition, so its state root equals its parent's. Two legitimately distinct empty blocks at the same height (different seal, timestamp, or coinbase, so different hashes) therefore share a state root, and the check fires a false positive.

## The fix

A new helper, `isSidechainGhostState`, narrows the rejection. It exempts a candidate only when the candidate carries no body, the canonical chain also changed no state at that height, and the candidate is anchored to the canonical chain at the fork point.

```go
canonNoOp := canonParent.Root == canonical.Root()
_, anchored := exempted[block.ParentHash()]
siblingOrAnchored := block.ParentHash() == canonical.ParentHash() || anchored
noRequests := block.RequestsHash() == nil || *block.RequestsHash() == types.EmptyRequestsHash
emptyBody := len(block.Transactions()) == 0 &&
    block.GasUsed() == 0 &&
    len(block.Uncles()) == 0 &&
    len(block.Withdrawals()) == 0 &&
    noRequests

return !(canonNoOp && siblingOrAnchored && emptyBody)
```

Every input is trusted or already validated:

- `canonical` comes from `GetBlockByNumber`, so its parent hash and root are canonical-chain values an attacker cannot influence.
- The body predicates read the candidate's own body. `ValidateBody` has already verified `TxHash`, `UncleHash`, and `WithdrawalsHash` against that body before `insertSideChain` sees the block. It returns `ErrPrunedAncestor` only after those checks, and the insert iterator runs it on every block.
- `exempted` holds hashes this same loop already proved to be no-ops. An entry is written only after the guard exempted that block, so an anchor is a block already shown to be empty with a root chaining back to the canonical fork header. The map is built by one `insertSideChain` call and never outlives it, so nothing is persisted and a later import cannot inherit an anchor.

Deliberately absent is any inference from an unproven side parent header. `insertSideChain` persists sidechain headers via `writeBlockWithoutState` without executing them, so a side parent `Root` on its own proves nothing. An earlier revision of this PR compared it; that inference is gone.

`canonNoOp` is required on top of body emptiness because an empty body does not imply an empty state transition in Bor. `Finalize` commits spans and state-sync events at sprint boundaries, which move the state root with no transactions in the block. If the canonical parent header is unavailable, the conservative reject is kept.

### Divergence longer than one block

A producer partitioned for a few seconds returns with several blocks of its own. Only the first is a sibling of the canonical block at its height. The second has the first as its parent, so a strict sibling test rejects it and the node stays wedged. #2224 clears under a strict test only because that node was stopped exactly at the divergence. The `exempted` anchor covers the general case.

### Predicates that cannot be reached on Bor

Only the transaction count and `GasUsed` do real work here. `Bor.verifyHeader` rejects a non-empty `UncleHash`, a non-nil `WithdrawalsHash`, and a non-nil `RequestsHash`. Those errors reach the insert iterator before `ValidateBody` runs, so such a block never reports `ErrPrunedAncestor` and never reaches this loop.

The three remaining predicates are kept as defense in depth and for parity with upstream go-ethereum, where the same false positive exists and those bodies are reachable. The predicate then means "provably carries nothing that can move state" on its own, rather than depending on three engine-level rejections holding across future merges. `noRequests` accepts both `nil` and the empty-set hash for the same reason: Bor headers always carry `nil`, while a post-Prague upstream empty block sets the empty-set hash.

## Behavior change

This intentionally accepts blocks the old code rejected. A run of empty blocks anchored to the canonical fork point, at heights where the canonical chain also changed no state, used to return `sidechain ghost-state attack`. Those blocks now fall through to the existing TD and reorg logic as normal sidechain blocks. That is the point of the fix, and it is the only widening.

Nothing else is loosened. A block that carries any body, is not anchored to the canonical chain at the fork point, or sits at a height where the canonical chain changed state is still rejected, as is the case where the canonical parent header is missing.

No new dependencies, no config or storage format changes.

## Executed tests

- `TestSideImportEmptyBlockGhostState` (new): builds a freezer-pruned canonical chain, generates a two-block run of distinct empty blocks at pruned heights, imports them, and asserts no ghost-state error and that both blocks reach disk. Confirmed it fails against the pre-fix code with the exact error, and it fails again if the anchor is removed.
- `TestSideImportGhostStateStillRejected` (new): a side block that performs a real state transition but claims the canonical root must still return the ghost-state error.
- `TestSidechainGhostStateExemption` (new): exercises the helper directly. Covers the genuine empty sibling, a non-sibling parent, the anchored child, a body-carrying child of an anchored parent, a body-carrying sibling at a no-op canonical height, and one case per emptiness conjunct so no single check can be dropped silently.
- `TestSidechainGhostStateAttackShapes` (new): an empty sibling at a state-changing canonical height, and a missing canonical parent header.
- `TestSideImportPrunedBlocks` (existing): still passes; the non-empty sidechain re-import path is unchanged.

All four ghost-state tests run against both the hash and path trie schemes.

Twelve hand-written mutants of the predicate were run: each conjunct forced true, each body check dropped, the anchor dropped, the sibling test dropped, the exemption never recorded, the nil branch inverted, and `noRequests` reduced to a plain nil check. All twelve fail the suite, so every predicate is genuinely covered rather than incidentally satisfied.

- `go test ./core/` passes with no regressions.
- `gofmt -l core/blockchain.go core/blockchain_test.go` is clean.
- `go build ./...` compiles.

## Notes

This is go-ethereum derived code, so the same false positive likely exists upstream.

The guarded path is live in bor: `insertSideChain` runs on `ErrPrunedAncestor` with `setHead`, and bor uses validator-based difficulty and TD, which is the path the issue's production trace shows.

#2224 also recorded a second problem: a node that already wedged does not recover from a short `debug_setHead`. This PR does not address that, and it is filed as #2386. The cause recorded in #2224 does not survive a read of the rewind path, so #2386 reports the symptom and leaves the cause open


pratikspatil024:
@claude review once

bit2swaz:
@pratikspatil024 sorry to disturb, but do you reckon i could get a review on this?

pratikspatil024:
@bit2swaz could you address the comments by copilot?
Also, diffguard is failing in the CI, please fix that as well.

bit2swaz:
Pushed 5ac7160d4 for both.

On Copilot's ghost-state point: it's right. the exemption now checks whether the *canonical* chain was a no-op at that height too using the canonical parent header from `GetHeaderByNumber` which is already verified. I moved the check into `isSidechainGhostState` and added `TestSidechainGhostStateCanonicalGate` which rejects a no-op side block sitting at a height where the canonical chain did change state.

On diffguard: pulling the check out into its own function also brought `insertSideChain` back under the complexity (+5) and size (+11) thresholds it had crossed, so Cognitive Complexity and Code Sizes pass now and mutation testing is 13/13.
The one FAIL left is the `core -> core` circular dependency, but that's already there on `develop` (diffguard counts the `core`/`core_test` split as a cycle) and this PR doesn't add any new imports, so it isn't from these changes

bit2swaz:
@pratikspatil024 could you take a look again whenever youve got the chance?

pratikspatil024:
@bit2swaz - sorry for the delay here. We will pick this up after `v2.10.1`

bit2swaz:
hey @pratikspatil024. i saw v2.10.1 is out, so just wanted to give you a nudge on this. lmk if anything is needed from my side

pratikspatil024:
codegenie review

pratikspatil024:
@claude review

bit2swaz:
both findings claim the guard can be bypassed because `sideParent.Root` is attacker controlled, but this impact just doesnt hold true

`insertSideChain` runs only when the parent state is pruned. at pruned depth, `bc.HasState(canonical.Root)` returns false. the ancestor recovery walk collects the forged blocks and passes them to `insertChain`. full EVM execution compares the computed root against the declared one, and mismatch causes rejection. `skipBlock` needs `ErrKnownBlock`. that requires `bc.HasBlockAndState(block.Hash(), ...)`. a forged block has a different hash, and so it never qualifies

the remaining concern is DoS. a rogue authorized validator at a no-op canonical height can force re-execution instead of cheap early rejection. this is bounded by the PoA validator set

the test gap was valid so i added `TestSidechainGhostStateNoOpCanonical` to cover both-no-op and state-changing-side-at-no-op-canonical paths. this closes the 3 uncovered lines in `isSidechainGhostState`

pratikspatil024:
diffguard in the CI is failing. can you check that and see if you can fix it?

bit2swaz:
the diffguard `core -> core` cycle is a pre existing false positive. it fires on every PR that touches the `core` package (check #2371 for reference)

everything else passes: complexity, code sizes, dead code, mutation testing 100%. anything else?

core/blockchain.go
```diff
@@ -4317,6 +4317,34 @@ func (bc *BlockChain) insertSideChain(block *types.Block, it *insertIterator, ma
...
+func (bc *BlockChain) isSidechainGhostState(block *types.Block, canonical *types.Block) bool {
+	canonParent := bc.GetHeaderByNumber(canonical.NumberU64() - 1)
+	sideParent := bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
+	if canonParent == nil || sideParent == nil {
+		return true
+	}
+	// Legitimate only if neither the canonical block nor the side block changed
+	// state relative to its own parent, i.e. both are genuine no-ops.
+	canonNoOp := canonParent.Root == canonical.Root()
+	sideNoOp := sideParent.Root == block.Root()
```
pratikspatil024: Request changes: I agree the reported empty-sibling sync wedge needs an exemption, and I do not see a currently exploitable shadow-state path. However, this helper comment overstates the guarantee: `sideNoOp` still derives from `sideParent.Root`, an unverified sidechain header written by `insertSideChain`, so this does not itself "close the forge-both-roots bypass".

The current implementation remains safe for a narrower reason: a forged chain trying to reach a recently available shared root must cross the canonical run boundary where that root was introduced; at that boundary `canonNoOp` is false and the existing guard rejects. That run-boundary/fork-point coupling is subtle, is not encoded by this predicate, and should not become the security argument future readers inherit.

Please make the exemption prove the intended sibling case from trusted/verifiable inputs instead: require the trusted `canonNoOp`, `block.ParentHash() == canonical.ParentHash()`, and a genuinely empty candidate body (at minimum no transactions and zero gas; include the applicable body-level checks for withdrawals/uncles). Keep `canonNoOp`: Bor finalization can produce state effects at sprint boundaries, so body emptiness alone is insufficient.

This removes the attacker-header inference entirely and makes the security test direct. Please also cover the helper/security cases under both hash and path schemes, remove the unused `forgedParent` in `TestSidechainGhostStateNoOpCanonical`, and update the PR description/rollout claim (the change intentionally accepts blocks that the old code rejected). Finally, #2224 also records the `debug_setHead`/persisted-side-block recovery problem; this PR should retain or link a follow-up rather than presenting that part as resolved.

bit2swaz: done in [b36184a](https://github.com/0xPolygon/bor/pull/2293/commits/b36184ae0d9f2098d1585431eef9493f9da2ee70). both schemes on all four ghost-state tests, forgedParent removed, PR description updated, and the setHead recovery part is now #2386

one note on #2386: the cause #2224 recorded doesnt survive a read of the rewind path (setHead deletes all side-fork headers via ReadAllHashes, and the trace shows ohead moving 1442→1441 with tail=427), so I filed it with the symptom and left the cause open rather than repeating the guess

pratikspatil024:
codegenie review

pratikspatil024:
@claude review

pratikspatil024 (commented):
Thanks — the predicate rewrite is right, and the per-conjunct tests are better than what I asked for. Three things left.

**1. Three of the five emptiness conjuncts can't be reached on bor, and the comments claim otherwise.** `Bor.verifyHeader` rejects `UncleHash != EmptyUncleHash` (`consensus/bor/bor.go:536`), non-nil `WithdrawalsHash` (`:554`) and non-nil `RequestsHash` (`:558`) — the last is a `!= nil` test, so `EmptyRequestsHash` is rejected too. Those errors land in `it.errors`, so `it.next()` returns them instead of `ErrPrunedAncestor` and the `insertSideChain` loop never reaches the guard with such a block. (The Finalize-side rejections at `:1237`/`:1411`/`:1471` don't matter here — nothing executes on this path.) Only the transaction count and `GasUsed` do real work on bor.

Keep all five — this is go-ethereum-derived code you propose upstreaming, and the predicate should mean "provably carries nothing that can move state" on its own rather than depending on three engine-level rejections holding across future merges. But please fix the prose:

- the `// EIP-7685 requests: nil before the fork, a pointer to the empty-set hash after it` comment describes a fork bor doesn't have; say instead that uncles, withdrawals and requests cannot reach this path on bor, are retained as defense-in-depth and for upstream parity, and that the live predicates are the transaction count and gas;
- the `// The empty-set requests hash is what a post-fork empty block actually carries` case asserts an exemption on a shape no bor block can have. Same for the `reqOnly` / `wdOnly` / `uncleOnly` rejection cases — fine as unit tests (they only construct because the tests run `noRewardEngine{ethash.NewFaker()}`, not bor's rules), but the comments shouldn't imply these blocks occur in production.

**2. The PR description still hasn't been updated.** It reads "the block is exempt only when its root equals its parent's root", which is the predicate this PR removed, and the rollout notes still say "it never accepts a block the old code would have accepted" and "it still rejects every block whose root differs from its parent's" — both now false. Executed tests omit the two new helper tests. For a consensus-path change this is the first thing an operator or a future upstream-merge reviewer reads.

**3. The strict sibling requirement fixes one divergent block, not the class.** With the node's head at `N`, divergence at `D < N`, and both sides empty (the quiet-network condition #2224 describes): the peer's block at `D` is a sibling of canonical `D` and is exempted, but the peer's block at `D+1` has the peer's `D` as parent while canonical `D+1` has the node's `D` — not siblings, so it's rejected and the node stays wedged. #2224 lands at `k=1` only because v2 was stopped exactly at the divergence; a producer partitioned for a few seconds comes back with `k` = blocks it produced.

The safe extension is small: track the hashes exempted inside the loop and accept an exempted parent as an anchor —

```go
sibling := block.ParentHash() == canonical.ParentHash() || exempted[block.ParentHash()]
```

Every input stays trusted: an exempted parent was itself proven a no-op whose root chains back to the canonical fork header, and in an all-empty run the declared root is the pruned fork-parent root throughout, so it still can't terminate the recovery walk or satisfy a child's `ValidateBody`. Please either add that, or say explicitly in the description that the fix covers single-block divergence and file the rest like you did with #2386.

Separately, on codegenie's open question about the pruning-boundary constants: `setupPrunedGhostStateChain` asserts `HasState(forkParent.Root())` is false with a `t.Fatalf`, so drift there fails loudly rather than passing vacuously — same idiom as the existing `testSideImportPrunedBlocks`.

CI note: the red codecov check is fork-PR infra (`No token specified`, then `429 Rate limit reached`), not yours. Everything else is green including diffguard.

bit2swaz:
all three done. ready for review :)

pratikspatil024 (commented):
LGTM on the code, thanks. A few small documentation/test cleanups before merge:

1. exempted is intentionally local to one insertSideChain batch. Please say that explicitly in the helper comment so the anchor is not read as persisted across imports.
2. Optional test strengthening: assert HasBlock for both side blocks after InsertChain, proving both blocks cleared the guard and were persisted. (The current error assertion already catches a ghost-state rejection because that path returns before ReorgNeeded.)
3. sibling := ... || anchored is now a sibling-or-anchor predicate; please rename it accordingly.

The PR description also still describes the removed parent-root predicate, omits the anchor/helper tests, and says the change never accepts a block the old code rejected. Please update it to match the current implementation.

bit2swaz:
all done. lmk if anything else is needed from me :)

PR #2303, opened 2026-07-14
eth, eth/fetcher, p2p: make tx arrival wait configurable via p2p.txarrivalwait

dkeysil:
## Summary
The `p2p.txarrivalwait` config option (`--txarrivalwait` flag) is parsed and documented but has no effect: since the geth v1.14.13 merge it is no longer passed into the transaction fetcher, which uses the hardcoded `txArriveTimeout` (500ms) instead. This PR wires the configured value through `node.Config` → `eth/backend` → `handlerConfig` → `NewTxFetcher`, restoring the flag's documented behavior. The default is unchanged (500ms). A value of `0` requests announced transactions immediately; negative values clamp to 0.

Why it matters: on a mainnet full node (~250 peers) we measured that a significant share of newly observed pending transactions (roughly 30–65% depending on peer topology; eth/68 senders direct-broadcast bodies only to √N peers, and blob/large txs are announce-only) arrive via hash announcement + fetch. Each of those currently waits up to 500ms before the body is even requested. With `txarrivalwait = "0s"` the `eth_fetcher_transaction_waiting_hashes` gauge stays pinned at 0 and announced transactions are fetched after a single round-trip.

## Executed tests
- `go test -race ./eth/fetcher/` — 144 tests pass, race clean.
- New tests: `TestTransactionFetcherCustomArrivalWait` (250ms wait honored), `TestTransactionFetcherZeroArrivalWait` (immediate scheduling), `TestTransactionFetcherNegativeArrivalWait` (clamped to 0), `TestTxArrivalWaitConfig` (TOML → `fillTimeDurations` → `buildNode` → `node.Config` plumbing, including the 500ms default).
- `go test ./internal/cli/server/` passes.
- Ran for multiple hours on a production Polygon PoS mainnet full node with `txarrivalwait = "0s"`: fetch-path deliveries (`eth_fetcher_transaction_replies_in`) continued at full rate (~115 tx/s of ~169 tx/s total intake), `waiting_hashes` at 0, node stable, no peer-drop or bandwidth regressions observed.

## Rollout notes
Not consensus-affecting; no coordinated upgrade required. Backwards compatible: the default stays 500ms, so operators who never set the flag see no change. Operators who *did* set `txarrivalwait` should note it has been silently ignored since v1.5.0 and will now take effect. In ≤v1.4.x the value was clamped to [100ms, 500ms]; this PR deliberately allows 0 so latency-sensitive operators can fetch announced transactions immediately, trading a modest increase in fetch-request traffic.

### DIFF
--- a/docs/cli/server.md
+++ b/docs/cli/server.md
@@ -322,7 +322,7 @@ The ```bor server``` command runs the Bor client.
 
 - ```txannouncementonly```: Whether to only announce transactions to peers (default: false)
 
-- ```txarrivalwait```: Maximum duration to wait for a transaction before explicitly requesting it (default: 500ms)
+- ```txarrivalwait```: Maximum duration to wait for a transaction before explicitly requesting it (0 requests announced transactions immediately) (default: 500ms)
 
 - ```v4disc```: Enables the V4 discovery mechanism (default: true)
 
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -455,6 +455,7 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		checker:                 checker,
 		enableBlockTracking:     eth.config.EnableBlockTracking,
 		txAnnouncementOnly:      eth.p2pServer.TxAnnouncementOnly,
+		txArrivalWait:           eth.p2pServer.TxArrivalWait,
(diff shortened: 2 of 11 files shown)


pratikspatil024:
Hi @dkeysil - are you working on this PR?

PR #2312, opened 2026-07-16
triedb/pathdb: trie-node lookup index for O(1) diff-layer resolution

lucca30:
## What

Adds a per-node **mutation-history index** to pathdb so `reader.Node` resolves the
owning layer in **O(1)** instead of walking the diff-layer chain newest→oldest
(up to `TriesInMemory = 128` map probes, each taking a per-layer `RLock`).

This mirrors the flat-state lookup index that #30971 already added for
accounts/storages — but for **trie nodes**, which currently have none. The new
`accountNodes`/`storageNodes` maps live in the existing `lookup` struct and are
maintained in the same `addLayer`/`removeLayer` functions, so trie-node reads now
use the same fast resolution path (`nodeTip` → `lookupNode`) that account and
storage reads already do.

## Why

Under the current code every `reader.Node` call walks the diff stack from the
entry layer down, probing each layer's node set and taking that layer's
`RLock` at every hop. With a full 128-layer stack a node that lives near the
bottom costs ~128 map probes per read. The flat-state reader was already freed
from this walk by #30971; trie-node reads were left behind.

This is on the hot path of parallel block execution (many goroutines resolving
trie nodes concurrently), so removing the per-read walk cuts CPU on every state
access.

## How

- `lookup.accountNodes` / `lookup.storageNodes`: mutation history keyed by
  owner+path, value = ordered list of diff-layer state roots that rewrote the
  node (same shape as `accounts`/`storages`).
- `lookup.nodeTip`: reverse-scans the history for the newest entry that is the
  queried state or an ancestor of it; falls back to the disk layer, or returns
  zero (stale) — identical semantics to `accountTip`/`storageTip`.
- `layerTree.lookupNode`: resolves the guaranteed-containing layer under
  `tree.lock.RLock`, mirroring `lookupAccount`/`lookupStorage`.
- `reader.Node`: resolves via `lookupNode`, reads the node at depth 0, and on
  the narrow disk-layer-went-stale window falls back to the legacy walk — exactly
  as `AccountRLP`/`Storage` do. The existing bor clean-cache hash-mismatch
  self-heal path is preserved unchanged.
- Index maintenance is factored into `addNodes`/`removeNodes` helpers invoked
  from `addLayer`/`removeLayer`, so it rides along on every mutation path: layer
  add, partial flatten (`cap`), full commit (`cap 0` → fresh `newLookup`), and
  journal replay on restart (`newLookup` rebuilds from the reconstructed stack).
  The index is **derived in-memory and never persisted**, so a restart rebuilds
  it from the journal exactly like the flat-state index.

## Correctness

The index is maintained by the **same** `addLayer`/`removeLayer`/`newLookup`
functions as the proven flat-state index, so it inherits the identical set of
maintenance points — there is no new mutation path that could leave it stale.
`nodeTip` reuses the descendant/stale logic of `accountTip`.

Tests added in `lookup_node_test.go`:

- **`TestNodeLookup`** — hand-verified tips (account trie + a storage owner),
  probing top/middle/bottom of the stack, after a `cap`, and stale states.
- **`TestNodeLookupEquivalence`** — for every observable `(owner, path, state)`,
  asserts the index resolution returns **byte-identical** data to the legacy
  newest→oldest walk, across a stack with re-writes at the same path, a deletion,
  a resurrection, two storage owners, and a side branch; re-checked after a
  partial cap. (Scoped to live states: dangling post-cap side-branch layers
  diverge from the naive walk exactly as `lookupAccount`/`lookupStorage` already
  do — inherent to the lookup-index design.)
- **`TestNodeLookupConcurrent`** — readers resolving nodes while the stack is
  mutated by concurrent `add`/`cap`; green under `-race`.

The full `triedb/pathdb` suite passes under `-race`.

## Measured on a mainnet node

Validated on a bor node replaying **mainnet** state (16 vCPU). A before/after CPU
profile shows node resolution effectively leave the profile:

```
pathdb diffLayer.node (cumulative CPU):   29.8%  ->  0.6%
```

with a clean tip figure of exec **2.42 ms/Mgas**. That profile was taken with the
in-memory diff stack full (sync/catch-up regime), which is exactly where the
newest→oldest walk is deepest and most expensive — i.e. the upper end of the win.
We're happy to share the profiles/methodology.

## Isolated branch validation — mainnet copy-node, 2026-07-22

The profile above came from the larger block-flow performance campaign build, so
it doesn't isolate this branch alone. To close that gap, we ran this exact
branch (rebased on current `develop`, commit `ea2b0cf59`) by itself against
live mainnet on a cloned copy-node, back-to-back against stock `develop`
(`4154750b4`) at chain tip, both with live block-building/sealing on:

| | Stock `develop` (`4154750b4`) | This branch (`ea2b0cf59`) |
|---|---|---|
| Window | 2026-07-22 12:04:41–13:05:04 UTC | 2026-07-22 13:07:46–14:08:09 UTC |
| Blocks | 2,428 | 2,420 |
| Avg import throughput | 285.9 Mgas/s | **351.0 Mgas/s (+22.8%)** |
| Avg import elapsed | 159.5 ms/block | **143.6 ms/block (−10.0%)** |

To be fair about what this is and isn't: the two windows are back-to-back but
cover **different blocks**, not a replay of the same blocks on both binaries —
so some of the gap could reflect differing tx mix or mempool depth between the
two hours rather than purely the binary change (`mgasps` already normalizes
for gas-per-block, which limits but doesn't eliminate that confound).

### Apples-to-apples follow-up: same-block rewind-replay

To settle that confound, we rewound the clone to a fixed block and replayed
the **exact same 1,000 blocks** (`90,682,047`–`90,683,047`) under each
binary, repeating the identical rewind so both runs saw identical
transactions:

| | Stock `develop` (`4154750b4`) | This branch (`ea2b0cf59`) |
|---|---|---|
| Total gas (sanity check) | 52,384.4 Mgas | 52,356.1 Mgas (Δ 0.05% — confirms identical blocks) |
| Total elapsed | 183.01s | 172.07s |
| Avg throughput | 286.2 Mgas/s | **304.3 Mgas/s (+6.3%)** |

This is the number we'd stand behind: **+6.3%** on identical blocks, smaller
than the +22.8% from the back-to-back windows above (which we now attribute
mostly to tx-mix/mempool differences between the two hours, not the code).
Direction is confirmed either way, and +6.3% at this diff-layer depth is
in line with the O(1)-vs-O(depth) mechanism producing a real but modest win
well short of the campaign's full-128-layer worst case
(`diffLayer.node` 29.8%→0.6% CPU share) — this replay only exercised a
partial diff stack, not a deep catch-up regime. We checked for a
replay-buffer cold-start artifact (mgasps in 200-block buckets across the
warm-up region before the measured window) and found it flat from the
first bucket, so no correction was needed there.

## Benchmark (per-read cost isolated)

Single node resolution over a 128-diff-layer stack, node written at max depth
(`lookup_node_test.go`):

```
BenchmarkNodeResolveWalkDeep     ~2200 ns/op   (legacy walk, O(depth))
BenchmarkNodeResolveIndexDeep      ~88 ns/op   (index, O(1))
```

~25× on a full-depth resolve; the walk is ~26 ns/hop so the break-even is ~3–4
layers deep and the index is constant thereafter. At a shallow tip the per-read
win is correspondingly smaller — the mainnet 29.8% figure above reflects the
deeper catch-up regime. The primary win is CPU/latency (O(`TriesInMemory`) →
O(1)); the index reuses the existing `tree.lock`, so it fits the current locking
model rather than changing it.

## Cost

The index is held in memory only — two maps in the existing `lookup` struct
recording, per mutated trie-node path across the live diff layers, the state
roots that rewrote it. Same shape and order of magnitude as the flat-state index
in #30971, bounded by `TriesInMemory`, shrinks on flatten, and **rebuilt from the
journal on restart (never persisted)**. Maintenance is O(nodes-in-layer) per
add/flatten, parallel to the existing account/storage index updates and off the
read path.

### DIFF
--- a/triedb/pathdb/layertree.go
+++ b/triedb/pathdb/layertree.go
@@ -321,6 +321,24 @@ func (tree *layerTree) lookupAccount(accountHash common.Hash, state common.Hash)
 	return l, nil
 }
 
+// lookupNode returns the layer that is guaranteed to contain the trie node
+// data corresponding to the specified state root being queried.
+func (tree *layerTree) lookupNode(owner common.Hash, path []byte, state common.Hash) (layer, error) {
+	// Hold the read lock to prevent the unexpected layer changes
+	tree.lock.RLock()
+	defer tree.lock.RUnlock()
+
+	tip := tree.lookup.nodeTip(owner, path, state, tree.base.root)
+	if tip == (common.Hash{}) {
+		return nil, fmt.Errorf("[%#x] %w", state, errSnapshotStale)
+	}
+	l := tree.layers[tip]
+	if l == nil {
+		return nil, fmt.Errorf("triedb layer [%#x] missing", tip)
+	}
+	return l, nil
+}
+
 // lookupStorage returns the layer that is guaranteed to contain the storage slot
 // data corresponding to the specified state root being queried.
 func (tree *layerTree) lookupStorage(accountHash common.Hash, slotHash common.Hash, state common.Hash) (layer, error) {
--- a/triedb/pathdb/lookup.go
+++ b/triedb/pathdb/lookup.go
@@ -49,6 +49,17 @@ type lookup struct {
 	// where the slot was modified, with the order from oldest to newest.
 	storages map[[64]byte][]common.Hash
 
+	// accountNodes represents the mutation history for account trie
+	// nodes, keyed by node path. The value is a slice of **diff layer**
+	// IDs where the node was rewritten, ordered from oldest to newest.
+	accountNodes map[string][]common.Hash
+
+	// storageNodes represents the mutation history for storage trie
+	// nodes, keyed by trie owner then node path. The value is a slice
+	// of **diff layer** IDs where the node was rewritten, ordered from
+	// oldest to newest.
+	storageNodes map[common.Hash]map[string][]common.Hash
+
 	// descendant is the callback indicating whether the layer with
 	// given root is a descendant of the one specified by `ancestor`.
 	descendant func(state common.Hash, ancestor common.Hash) bool
@@ -65,9 +76,11 @@ func newLookup(head layer, descendant func(state common.Hash, ancestor common.Ha
 		current = current.parentLayer()
 	}
 	l := &lookup{
-		accounts:   make(map[common.Hash][]common.Hash),
-		storages:   make(map[[64]byte][]common.Hash),
-		descendant: descendant,
+		accounts:     make(map[common.Hash][]common.Hash),
+		storages:     make(map[[64]byte][]common.Hash),
+		accountNodes: make(map[string][]common.Hash),
+		storageNodes: make(map[common.Hash]map[string][]common.Hash),
+		descendant:   descendant,
 	}
 	// Apply the diff layers from bottom to top
 	for i := len(layers) - 1; i >= 0; i-- {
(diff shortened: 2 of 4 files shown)


lucca30:
@claude review

PR #2314, opened 2026-07-21
consensus/bor, internal/ethapi: skip span and state-sync commits in simulated blocks

cffls:
## Summary

`eth_simulateV1` runs simulated blocks through Bor's `FinalizeAndAssemble`, which at sprint-start blocks commits span and state-sync data. Those commits make internal genesis contract calls that resolve the parent header by hash from the database — but a simulated block's parent can be a phantom header that is never persisted: the pending block when simulating on top of the `pending` tag, or an earlier simulated block in a multi-block request. The lookup returns no header, and the request fails with `-32000: DoCall returned nil result with no error (block=0x...)` whenever the first simulated block lands on a sprint boundary — every 16th block on mainnet, i.e. ~6% of `pending`-based requests. (Both failing blocks reported by users confirm the pattern: `0x5637f0f` and `0x5637d5f`, each `(multiple of 16) - 1`, so the simulated block `+1` is a sprint start.)

This adds `FinalizeAndAssembleForSimulation`, which skips the sprint-start span and state-sync commits, and dispatches to it from the simulator (via an interface assertion, so non-Bor engines keep the regular path). The skipped data is external Heimdall input that cannot be known for a future block anyway; skipping also removes an RPC-triggered synchronous Heimdall call (`waitUntilHeimdallIsSynced` + `StateSyncEvents`) from the public `eth_simulateV1` path. This matches upstream geth semantics, where simulate runs no consensus side effects.

Also hardens `DoCall`: a missing header on the by-hash path now returns an explicit `header not found for hash 0x...` error instead of a nil result with no error.

Behavioral note: simulated sprint-start blocks no longer contain state-sync transactions that the eventual real block might include. This is intentional — those events are unknowable ahead of production time, and callers who want them can model them deterministically via `stateOverrides`.

## Executed tests

- New regression test `TestFinalizeAndAssembleForSimulationSkipsSprintCommits`: a sprint-start block with a non-persisted parent fails through the regular path (state-sync commit attempted) and assembles cleanly through the simulation path.
- New `TestCallWithStateMissingHeader` covering the explicit `DoCall` error.
- Full `consensus/bor`, `internal/ethapi`, and `ethclient` (`TestSimulateV1*`) suites pass locally.
- Diffguard with mutation testing: 100% kill rate (13/13), no dependency cycles, no dead code. One surfaced pre-existing size violation: `(Bor).finalizeAndAssemble` (68 lines, the renamed existing `FinalizeAndAssemble` body) — not refactored here to keep the consensus diff minimal.

## Rollout notes

Not consensus-affecting: the change only alters the `eth_simulateV1` execution path (simulated blocks are never sealed or imported) and an internal error message. Real block production, verification, and import paths are untouched — `FinalizeAndAssemble` behavior is unchanged for them. Backwards-compatible, no coordinated upgrade needed. Operator-facing change: `eth_simulateV1` with `pending` (or multi-block requests) no longer fails sporadically at sprint boundaries.

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -1387,26 +1387,10 @@ func (c *Bor) finalizeAndAssemble(chain consensus.ChainHeaderReader, header *typ
 	)
 
 	if !simulated && IsSprintStart(headerNumber, c.config.CalculateSprint(headerNumber)) {
-		borStart := time.Now()
-		cx := statefull.ChainContext{Chain: chain, Bor: c}
-
-		// check and commit span
-		if !c.config.IsRio(header.Number) {
-			if err = c.checkAndCommitSpan(state, header, cx); err != nil {
-				log.Error("Error while committing span", "error", err)
-				return nil, nil, 0, err
-			}
+		stateSyncData, err = c.commitSprintWork(chain, header, state)
+		if err != nil {
+			return nil, nil, 0, err
 		}
-
-		if c.HeimdallClient != nil {
-			// commit states
-			stateSyncData, err = c.CommitStates(state, header, cx)
-			if err != nil {
-				log.Error("Error while committing states", "error", err)
-				return nil, nil, 0, err
-			}
-		}
-		state.BorConsensusTime = time.Since(borStart)
 	}
 
 	if err = c.changeContractCodeIfNeeded(headerNumber, state); err != nil {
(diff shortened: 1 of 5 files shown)


cffls:
@claude review

pratikspatil024 (approved):

kamuikatsurgi (approved):

PR #2319, opened 2026-07-24
core, eth, trie: merge geth v1.17.0 (v1.17.4 sync, milestone 2/6)

pratikspatil024:
> [!IMPORTANT]
> **Reviewer guide — stacked PR 2 of 12.** Part of the combined go-ethereum **v1.17.4 + v1.17.5** upstream sync, which ships as **one stable release**. Every PR in the stack merges into the base branch `upstream-merge-v1.17.4`; that base merges into `develop` once, at the very end — not per-PR.
>
> **Merge-commit only — never squash.** Squashing rewrites a branch's SHAs and breaks every PR stacked above it.
>
> **Review bottom-up:** #2308 → #2319 → #2325 → #2328 → #2337 → #2340 → #2341 → #2342 → #2343 → #2345 → #2346 → #2354. Start at #2308 / #2319 — every PR above inherits them, so reviewing top-down means re-reviewing.
>
> **Expected-red / flaky checks (not code blockers):** `Quality metrics` (diffguard — skipped by team decision; it also mis-scopes across a stacked diff, comparing against the bottom of the stack), and `codecov/project` (repo-wide coverage threshold; per-PR patch coverage is green). Kurtosis e2e occasionally flakes (~1-in-5, devtools-owned) and is re-run by hand. Full per-batch conflict-resolution reasoning is in `docs/upstream-merges/`.

---

## Summary

Milestone **2/6** of the Bor ← go-ethereum **v1.17.4** upstream sync: merges
go-ethereum **v1.17.0** (upstream range `v1.16.9`-tip → `v1.17.0` stable,
`0cf3d3ba4`) into Bor across **12 review-sized merge batches** (~20 upstream
first-parent commits each) plus one standalone go-verkle-removal commit and a
milestone-chores commit.

Stacked on **#2308** (v1.16.9, milestone 1/6). Base is `ppatil-upstream-v1.16.9`
for now; it retargets to `upstream-merge-v1.17.4` when #2308 merges.

**⚠️ Merge, never squash.** Every batch is a real 2-parent merge commit so the
merge base advances and upstream history/authorship is preserved. Squash-merging
this PR (or any PR below it in the stack) rewrites SHAs and breaks every stacked
PR above it. Use a merge commit.

### What landed

Most of v1.17.0 is bugfix/tooling/test sweeps that adopted cleanly. The
consensus-relevant decisions, per-batch, are in
`docs/upstream-merges/v1.17.4/ledger.md`; fork decisions in `fork-register.md`;
the deferred backlog (with adopt-steps) in `needs-wiring.md`.

- **No fork/EIP activated.** Every v1.17.0 fork/EIP surface is merged **dormant**
  on Bor networks (EIP-8024 `ExtraEips`-only; Verkle gate nil). No precompile-set,
  gas-schedule, or activation-block change on any Bor preset. `params.Rules.ChainID`
  retained (declined upstream removal) to satisfy Bor's precompile-parity guard.
- **Adopted (consensus-relevant):** `crypto/keccak` vendoring (#33323); eip4844
  `latestBlobConfig`→`(BlobConfig,error)` keeping Bor's block-based gating + no-config
  `return 0`; `opKeccak256` Keccak256Cache adapted to upstream's hasher; trienode-history
  (#32621/#33551/#33584) + `NodeFullValueCheckpoint` (#32727), both dormant by default;
  eth_getProofs-for-history; callTracer log index (goldens regenerated preserving Bor's
  `0x1010` fee logs); pathdb history-index mutex combined with upstream's iterator
  refactor; trie decode `CountValues` error check; freezer fd-leak fix.
- **go-verkle scaffolding removed** (mirrors upstream #33461; kept the dormant Verkle
  gate + binary-tree + `stateless.Witness`).
- **Deferred / declined** (reverted to Bor HEAD, tracked in `needs-wiring.md` with
  adopt-steps — nothing dropped silently): the state code-read metrics line, core/vm
  write-protection + selfdestruct cluster, OpenTelemetry line (#33452/#33484/#33521/#33780;
  Bor has its own server-side OTel), delayed-p2p-decoding (#33835), EraE format (#32157),
  syncModer, slow-block stats, tx-announcement validation, snap-sync locking, legacypool
  alloc, `--miner.maxblobs`. These clash with Bor divergences (BlockSTM state, `SafePool`,
  `ExcludeStateSyncReceipt`, TD-in-DB, forked downloader, per-fork signer chain).
- **Milestone chores:** regenerated `eth/ethconfig/gen_config.go` (`TrienodeHistory` +
  `NodeFullValueCheckpoint` + `NoSnapServing` TOML); exempted `DumpBinTrieLeaves` in the
  StateDB↔ParallelStateDB parity guard.

## Executed tests

Beyond the standard CI gates:

- `go build ./...` — clean. `go vet ./...` — clean except two pre-existing
  `//nolint:govet` copylocks (identical to `develop`). `go mod tidy` — no drift
  (this milestone adds **no** dependencies).
- **Full `go test ./...`** — green after fixing the only two merge artifacts it
  surfaced (`ethclient` `rpc.NewServer` signature; `internal/ethapi` `blockTimestamp`
  goldens). Remaining failures are pre-existing and byte-identical to `develop`
  (VEBLOP miner nil-deref on non-Bor-network test harnesses in `cmd/geth`/`cmd/devp2p/ethtest`;
  t8n golden drift; a `core/vm` interrupt-race flake).
- **`make test-integration`** (`-tags integration ./tests/...`) — **pass, zero failures**,
  including the `tests/bor` consensus e2e suite (621s).
- **govulncheck ./...** — no merge-introduced vulnerabilities (3 called vulns are all
  pre-existing/`develop`-inherent dependency/toolchain versions).

Not run in-session (operator/CI release gates): diffguard mutation, kurtosis devnet,
and the devnet→testnet soak ladder. Kurtosis marginal value is low here — `tests/bor`
(consensus e2e) is green and the only observed panics are non-Bor-network paths.

## Rollout notes

- **Not consensus-affecting on Bor networks.** No fork activated, no precompile/gas/
  activation-height change; all adopted features either dormant-by-default or
  non-consensus (tooling/RPC/metrics/tests). Deferred upstream features were reverted
  to Bor HEAD.
- **Backward-compatible; no coordinated upgrade required** for this milestone.
- Operator-visible additions: `blockTimestamp` field on `eth_getTransactionBy*` RPC
  responses (additive); download progress-bar behavior; HTTP/2 JSON-RPC.
- Follow-ups for the team (pre-existing, outside this sync): nil-guard the VEBLOP miner
  timeout (`miner/worker.go:774`) so non-Bor-network tests stop panicking; dependency
  bumps (x/text→v0.39.0, cascade `develop`'s x/crypto v0.52.0, Go→1.26.5).


pratikspatil024:
codegenie review

pratikspatil024:
@claude review

pratikspatil024:
codegenie review

core/state/access_events.go
```diff
@@ -76,18 +72,20 @@ func (ae *AccessEvents) Keys() [][]byte {
...
-		key := utils.GetTreeKeyWithEvaluatedAddress(basePoint, &chunk.treeIndex, chunk.leafKey)
+		var offset [32]byte
+		treeIndexBytes := chunk.treeIndex.Bytes32()
+		copy(offset[:31], treeIndexBytes[1:])
+		offset[31] = chunk.leafKey
+		key := bintrie.GetBinaryTreeKey(chunk.addr, offset[:])
 		keys = append(keys, key)
 	}
 	return keys
 }
```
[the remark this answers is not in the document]
pratikspatil024: Correct, and thanks for the byte-level walkthrough — but not something this PR will change.

`trie/bintrie/key_encoding.go` is byte-identical to upstream `v1.17.0` here, so the mismatch is upstream's, arriving via the sync rather than authored by us. It's also unreachable on Bor today: the binary-trie / EIP-4762 path is gated behind `VerkleBlock`, which is `nil` on every Bor preset, so `Keys()` never runs.

Patching upstream code inside a sync merge means re-resolving that divergence on every future sync, for no runtime effect. Recording it in `docs/upstream-merges/.../needs-wiring.md` instead, so it's a blocker on ever enabling the fork rather than a silent gap, and raising it upstream.

core/vm/instructions.go
```diff
@@ -1009,6 +1001,115 @@ func opSelfdestruct6780(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, erro
...
+	// DUPN duplicates the n'th stack item, so the stack must contain at least n elements.
+	if scope.Stack.len() < n {
+		return nil, &ErrStackUnderflow{stackLen: scope.Stack.len(), required: n}
+	}
+
+	//The n‘th stack item is duplicated at the top of the stack.
+	scope.Stack.push(scope.Stack.Back(n - 1))
+	*pc += 1
+	return nil, nil
+}
```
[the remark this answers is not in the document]
pratikspatil024: Leaving this as-is.

The three handlers are upstream code (EIP-8024 arrives with this merge), so factoring a helper out of them would be a Bor-only edit to a file upstream keeps touching — a conflict we'd re-resolve every sync.

It also runs against our own style guidance, which is to leave three similar call sites alone until a fourth arrives or all three have to change together. Neither applies yet.

core/txpool/blobpool/blobpool.go
```diff
@@ -1729,6 +1526,21 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
...
+			} else {
+				// if maxGapped is reached, it is better to give time to gapped
+				// transactions by keeping the old and dropping this one.
+				// Thus replacing a gapped transaction with another gapped transaction
+				// is discouraged.
+				log.Trace("no gapped blob queue allowance", "allowance", allowance, "hash", tx.Hash(), "from", from, "nonce", tx.Nonce(), "qlen", len(p.gapped[from]))
+			}
 		case errors.Is(err, core.ErrInsufficientFunds):
 			addOverdraftedMeter.Mark(1)
 		case errors.Is(err, txpool.ErrAccountLimitExceeded):
```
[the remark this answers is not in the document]
pratikspatil024: Accurate reading of the code — `len(p.gapped)` is an account count and the constant's comment promises a transaction count — but it isn't reachable on Bor.

Bor never constructs a blobpool: `blobpool.New(` has no non-test callers, and `eth/backend.go` registers only `[]txpool.SubPool{legacyPool}`. The whole package is inherited-but-unwired, so no blob transaction reaches `addLocked`. The code is also upstream's (geth #32717), not ours.

So not fixed here — patching upstream code mid-sync creates permanent divergence. Recording it in `needs-wiring.md` so it's caught if blobs are ever wired up, and it's worth an upstream report.

core/state/access_events.go
```diff
@@ -76,18 +72,20 @@ func (ae *AccessEvents) Keys() [][]byte {
 	// TODO: consider if parallelizing this is worth it, probably depending on len(ae.chunks).
 	keys := make([][]byte, 0, len(ae.chunks))
 	for chunk := range ae.chunks {
-		basePoint := ae.pointCache.Get(chunk.addr[:])
-		key := utils.GetTreeKeyWithEvaluatedAddress(basePoint, &chunk.treeIndex, chunk.leafKey)
+		var offset [32]byte
+		treeIndexBytes := chunk.treeIndex.Bytes32()
+		copy(offset[:31], treeIndexBytes[1:])
```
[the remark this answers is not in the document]
pratikspatil024: Correct, and thanks for the byte-level walkthrough — but not something this PR will change.

`trie/bintrie/key_encoding.go` is byte-identical to upstream `v1.17.0` here, so the mismatch is upstream's, arriving via the sync rather than authored by us. It's also unreachable on Bor today: the binary-trie / EIP-4762 path is gated behind `VerkleBlock`, which is `nil` on every Bor preset, so `Keys()` never runs.

Patching upstream code inside a sync merge means re-resolving that divergence on every future sync, for no runtime effect. Recording it in `docs/upstream-merges/.../needs-wiring.md` instead, so it's a blocker on ever enabling the fork rather than a silent gap, and raising it upstream.

core/txpool/blobpool/blobpool.go
```diff
@@ -1729,6 +1526,21 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
 			addStaleMeter.Mark(1)
 		case errors.Is(err, core.ErrNonceTooHigh):
 			addGappedMeter.Mark(1)
+			// Store the tx in memory, and revalidate later
+			from, _ := types.Sender(p.signer, tx)
+			allowance := p.gappedAllowance(from)
+			if allowance >= 1 && len(p.gapped) < maxGapped {
```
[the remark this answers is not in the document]
pratikspatil024: Accurate reading of the code — `len(p.gapped)` is an account count and the constant's comment promises a transaction count — but it isn't reachable on Bor.

Bor never constructs a blobpool: `blobpool.New(` has no non-test callers, and `eth/backend.go` registers only `[]txpool.SubPool{legacyPool}`. The whole package is inherited-but-unwired, so no blob transaction reaches `addLocked`. The code is also upstream's (geth #32717), not ours.

So not fixed here — patching upstream code mid-sync creates permanent divergence. Recording it in `needs-wiring.md` so it's caught if blobs are ever wired up, and it's worth an upstream report.

core/txpool/blobpool/blobpool.go
```diff
@@ -1866,6 +1678,58 @@ func (p *BlobPool) add(tx *types.Transaction) (err error) {
 	p.updateStorageMetrics()
 
 	addValidMeter.Mark(1)
+
+	// Notify all listeners of the new arrival
+	p.discoverFeed.Send(core.NewTxsEvent{Txs: []*types.Transaction{tx.WithoutBlobTxSidecar()}})
```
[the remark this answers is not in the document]
pratikspatil024: Checked this one specifically, since "moved into `addLocked`" would be a merge-resolution error if it were ours — it isn't. Upstream `v1.17.0` has both sends inside `addLocked` at the same lines (1682-1683), so the structure is upstream's own.

And as with the other blobpool findings: Bor never constructs a blobpool (`blobpool.New(` has no non-test callers; `eth/backend.go` registers only `legacyPool`), so this path doesn't execute here.

Not changing it in a sync PR, for the usual divergence reason.

cffls:
Any idea why e2e test failed?

pratikspatil024:
@cffls somehow GitHub is malfunctioning here. The CI on the PR is not showing the latest run.
If you check the CI against the [latest commit](https://github.com/0xPolygon/bor/commit/d41472d4fc3db78258ebb493e47a777ea53d5121) on this PR, the tests are passing.
Screenshot 2026-08-24 at 9 47 35 AM

core/vm/instructions.go
```diff
@@ -260,33 +261,26 @@ func opKeccak256(pc *uint64, evm *EVM, scope *ScopeContext) ([]byte, error) {
...
 		var key [64]byte
 		copy(key[:], data)
 		if cached, ok := evm.Config.Keccak256Cache.Load(key); ok {
-			h := cached.(common.Hash)
-			if evm.Config.EnablePreimageRecording {
-				evm.StateDB.AddPreimage(h, data)
-			}
-			size.SetBytes32(h[:])
-			return nil, nil
+			hash = cached.(common.Hash)
```
marcello33, line 271: this used to reuse a hasher off the EVM struct (no allocs), but now it calls `crypto.Keccak256Hash` which allocates a bit each time it's not a cache hit. just mentioning as this is probably one of the opcodes we use the most, but might not a real issue at our throughput. wdyt?

rpc/http.go
```diff
@@ -332,6 +343,9 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	ctx = context.WithValue(ctx, peerInfoContextKey{}, connInfo)
 
+	// Extract trace context from incoming headers.
+	ctx = otel.GetTextMapPropagator().Extract(ctx, propagation.HeaderCarrier(r.Header))
```
marcello33: the ledger declines otel tracing for the other rpc files (`client/handler/server/service.go`) but this still parses otel headers off every incoming RPC request. probably not an issue today since nothing consumes the span, but flagging it to know if there's a reason to keep it here specifically?

core/vm/eips.go
```diff
@@ -43,6 +43,7 @@ var activators = map[int]func(*JumpTable){
 	4762: enable4762,
 	7702: enable7702,
 	7939: enable7939,
+	8024: enable8024,
```
marcello33, line 46: We have no tests for this, right? I mean nothing would fail in the CI if `enable8024` (or others) get wired into an active preset. Maybe worth having a table test.

core/state/parallel_statedb.go
marcello33: can't find this PointCache anymore, maybe we want to update the comment here

node/rpcstack.go
```diff
@@ -149,6 +149,9 @@ func (h *httpServer) start() error {
 
 	// Initialize the server.
 	h.server = &http.Server{Handler: h}
+	h.server.Protocols = new(http.Protocols)
+	h.server.Protocols.SetHTTP1(true)
+	h.server.Protocols.SetUnencryptedHTTP2(true)
```
marcello33, line 154: had agents flagging this: 
```
This turns on HTTP/2 without encryption for our RPC/engine-api endpoints. If whatever sits in front of these (nginx, etc.) doesn't handle that combination carefully, there's a known way for someone to sneak a second hidden request past the proxy straight to bor. No idea if that actually applies to how we've got things set up, just flagging since I haven't seen anyone check it. Might be totally fine.
```
Maybe worth a check

core/state/database.go
```diff
@@ -287,10 +280,12 @@ func (db *CachingDB) OpenTrie(root common.Hash) (Trie, error) {
...
 		ts := overlay.LoadTransitionState(db.TrieDB().Disk(), root, db.triedb.IsVerkle())
 		if ts.InTransition() {
-			panic("transition isn't supported yet")
+			panic("state tree transition isn't supported yet")
 		}
 		if ts.Transitioned() {
-			return trie.NewVerkleTrie(root, db.triedb, db.pointCache)
+			// Use BinaryTrie instead of VerkleTrie when IsVerkle is set
+			// (IsVerkle actually means Binary Trie mode in this codebase)
+			return bintrie.NewBinaryTrie(root, db.triedb)
```
marcello33, line 301: So this returns a `*bintrie.BinaryTrie` now, but `mustCopyTrie` below still doesn't handle `BinaryTrie`, right? I fear once this path is active, the `StateDB.Copy()` can panic. Shall we add the `BinaryTrie.Copy()` case too?

internal/ethapi/api.go
```diff
@@ -2212,7 +2202,7 @@ func (api *TransactionAPI) SendRawTransaction(ctx context.Context, input hexutil
 
 // SendRawTransactionSync will add the signed transaction to the transaction pool
 // and wait until the transaction has been included in a block and return the receipt, or the timeout.
-func (api *TransactionAPI) SendRawTransactionSync(ctx context.Context, input hexutil.Bytes, timeoutMs *hexutil.Uint64) (map[string]interface{}, error) {
+func (api *TransactionAPI) SendRawTransactionSync(ctx context.Context, input hexutil.Bytes, timeoutMs *uint64) (map[string]interface{}, error) {
```
marcello33, line 2211: is this change intentional? I mean callers now can pass a normal quantity string for the timeout and hit invalid params

core/evm.go
```diff
@@ -118,19 +118,16 @@ func EthereumTransfer(db vm.StateDB, sender, recipient common.Address, amount *u
 func NewEVMTxContext(msg *Message) vm.TxContext {
 	ctx := vm.TxContext{
 		Origin:     msg.From,
-		GasPrice:   new(big.Int).Set(msg.GasPrice),
+		GasPrice:   uint256.MustFromBig(msg.GasPrice),
```
marcello33, line 122: Can this panic before the normal fee-cap sanity errors? The block import builds the tx context before preCheck, where the relative errors are returned, so an oversized gas price from a bad block might panic the node instead of rejecting cleanly, no?

core/txpool/legacypool/legacypool.go
```diff
@@ -1975,6 +1985,8 @@ func (pool *LegacyPool) demoteUnexecutables() {
 			// Internal shuffle shouldn't touch the lookup set.
 			pool.enqueueTx(hash, tx, false)
 		}
+		pool.priced.Removed(len(olds) + len(drops))
```
marcello33, line 2027: should conditional removals be included too? `FilterTxConditional` removes txs from pending, but doesn’t subtract them from `list.totalcost`, so `ExistingExpenditure` can stay inflated after a conditional tx expires.

core/state_processor.go
```diff
@@ -80,22 +80,23 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
 
+	var tracingStateDB = vm.StateDB(statedb)
+	if hooks := cfg.Tracer; hooks != nil {
+		tracingStateDB = state.NewHookedState(statedb, hooks)
+	}
+
 	// Mutate the block and state according to any hard-fork specs
 	if config.DAOForkSupport && config.DAOForkBlock != nil && config.DAOForkBlock.Cmp(block.Number()) == 0 {
-		misc.ApplyDAOHardFork(statedb)
+		misc.ApplyDAOHardFork(tracingStateDB)
```
cffls, line 90: Do we need to do the same change for parallel_state_processor?

cffls (commented):
Looks good from my side! One minor question about state db change.

core/vm/instructions.go
pratikspatil024: Checked this one — no new allocation. Upstream v1.17.0 deleted `hasher`/`hasherBuf` from the EVM struct (they're still there on develop, `core/vm/evm.go:182-183`), so the rewrite wasn't optional on our side.

The replacement is pool-backed in the reworked crypto package — `crypto.Keccak256Hash` does `hasherPool.Get()` / `Reset()` / `Put()` (`crypto/keccak.go:54-63`), so there's no per-call hasher allocation. The cache-hit path is also better than before: it now returns the cached hash without touching a hasher at all, where the old code still had to hash on the miss branch inside the same block.

rpc/http.go
pratikspatil024: Good catch, and this one is a miss in my resolution rather than a deliberate keep.

Batch 9 declined the OTel JSON-RPC tracing (#33452) by reverting `rpc/{client,handler,server,service}.go`, but `rpc/http.go` wasn't in that revert list, so the propagator line and its two imports survived on their own.

It isn't inert either — we install a global `propagation.TraceContext` at `internal/cli/server/server.go:470`, so a caller-supplied `traceparent` header would get stitched into our own spans while nothing in `rpc` actually consumes the extracted context.

Fixed in [`a16641516`](https://github.com/0xPolygon/bor/pull/2319/commits/a166415168f2ff698e883e2c2ce08b7fa9cd15dd): removed the line and the two imports, and recorded the correction against the #33452 entry in `docs/upstream-merges/v1.17.4/ledger.md`.

core/vm/eips.go
pratikspatil024: You're right that nothing guards it. `ExtraEips` is never set by any chain preset or CLI flag — the only references are `cmd/evm` t8n and `tests/` — so 8024 is unreachable today, but that's an accident of nobody having wired it rather than anything enforced.

Fixed in [`a16641516`](https://github.com/0xPolygon/bor/pull/2319/commits/a166415168f2ff698e883e2c2ce08b7fa9cd15dd): added `TestEIP8024StaysOutOfShippedInstructionSets` in `core/vm/jump_table_test.go`, asserting DUPN/SWAPN/EXCHANGE stay `undefined` across every instruction set we ship, plus a positive control (`EnableEIP(8024, ...)` must flip them to defined) so the assertion can't pass vacuously if the opcodes ever become unreachable for some unrelated reason.

core/state/parallel_statedb.go
pratikspatil024: Yep — `PointCache` is gone from the tree entirely now that go-verkle is out; that comment was the only surviving reference in Go code.

Fixed in [`a16641516`](https://github.com/0xPolygon/bor/pull/2319/commits/a166415168f2ff698e883e2c2ce08b7fa9cd15dd): updated to `raw base for Witness only`.

node/rpcstack.go
pratikspatil024: Worth flagging, agreed. Two parts to the answer.

The code is byte-for-byte identical to upstream go-ethereum v1.17.0 (`node/rpcstack.go:141-143`), so I'd rather not introduce a Bor-only divergence inside a sync PR — that's a decision with an ongoing maintenance cost on every future merge.

Whether the concern actually applies to us depends entirely on what sits in front of the RPC listeners, which nobody has checked. Tracked as **POS-3705** to confirm what terminates and forwards RPC traffic on mainnet and Amoy, whether that layer normalises cleartext HTTP/2 to HTTP/1.1, and that the engine API listener isn't externally reachable regardless. If a real gap turns up there, disabling `SetUnencryptedHTTP2` as an explicit, ledger-recorded divergence is the fallback.

core/state/database.go
pratikspatil024: Confirmed — and `BinaryTrie.Copy()` already exists (`trie/bintrie/trie.go:356`), so it really is just a missing case in the type switch.

Two things keep me from fixing it here. It's identical to upstream v1.17.0 (`core/state/database.go:298-307`), so the gap is upstream's, not something the merge introduced. And the path is dormant for us: reaching it needs `triedb.IsVerkle()`, which no Bor config sets.

Leaving it matching upstream rather than adding a Bor-only divergence in a sync PR — this belongs upstream.

internal/ethapi/api.go
pratikspatil024: Intentional, though it's upstream's change rather than mine: v1.17.0 has `timeoutMs *uint64` (`internal/ethapi/api.go:1666`) where we had `*hexutil.Uint64`.

You're right that it's caller-visible, and it's a breaking change rather than a cosmetic one — `eth_sendRawTransactionSync` now expects a JSON number and will reject `"0x3e8"`. Keeping upstream's signature so we don't diverge on the RPC surface, and adding it to the release's rollout notes so it's called out to anyone already using the method.

core/evm.go
pratikspatil024: Your reading of the ordering is right — the tx context is built before `preCheck`, so the `GasFeeCap.BitLen() > 256` guard comes too late to protect `MustFromBig`.

Two bounds keep it from being externally reachable: the RPC path caps at 256 bits via `hexutil.Big`'s decoder, and the txpool rejects oversized fee caps at `core/txpool/validation.go:105`. What's left is a transaction already inside a block, which means a signing validator — real, but not something an arbitrary peer can drive.

It's byte-identical to upstream v1.17.0 (`core/evm.go:83`), so I'm not diverging here; this is an upstream report.

core/txpool/legacypool/legacypool.go
pratikspatil024: Confirmed. `FilterTxConditional` reheaps but never calls `subTotalCost`, unlike `Filter` (`list.go:432-433`) and `Cap` (`:480`), so `l.totalcost` stays inflated after a conditional expires. That value is read back through `ExistingExpenditure` (`legacypool.go:813`) and consumed by `core/txpool/validation.go:274`, so the sender's committed spend overcounts and later txs from that account can be rejected for insufficient balance when they're actually affordable.

It's pre-existing on develop and untouched by this PR, so I don't want to grow the sync diff with it. Filed as **POS-3704** with the suggested fix and a regression test outline.

core/state_processor.go
pratikspatil024: Not needed, for two independent reasons.

The DAO branch is dead on Bor — `DAOForkBlock` is nil on both the mainnet and Amoy presets, so `ApplyDAOHardFork` never runs on any of our networks regardless of which StateDB it gets. And V2 refuses a tracer outright (`errV2TracerUnsupported`, `core/parallel_state_processor.go:44-47`, enforced at `:1112-1114`) and falls back to the serial processor, so `cfg.Tracer` is nil whenever the parallel path actually executes. The other half of the hunk is a no-op: `config` is just `p.chainConfig()`.

Your question does surface something real though, just wider than this hunk: the V1 parallel path never builds a `state.NewHookedState` at all, so StateDB-level tracing hooks are dropped there entirely. That's pre-existing on develop rather than anything this merge changed. Happy to track it separately if you think it's worth pinning down.

marcello33 (approved):

cffls (approved):

PR #2335, opened 2026-07-31
eth/filters: decouple client notification delivery from event fan-out

mukul3097:
## Problem

The filter `EventSystem` fans events out to every installed subscription from a single `eventLoop` goroutine using **blocking** channel sends (`eth/filters/filter_system.go`, `handleTxsEvent` et al.), and the per-subscription goroutines in `eth/filters/api.go` deliver to clients with a synchronous `notifier.Notify`. The two together create a back-pressure chain from an untrusted RPC client into shared node state:

```
slow WS client ← notifier.Notify ← api.go subscription goroutine ← subscription channel ← shared eventLoop
```

A WebSocket client that stops reading its connection — or reads just slowly enough never to trip the RPC write deadline — blocks its subscription goroutine in `Notify`, its subscription channel fills, and the shared `eventLoop` then blocks on that one subscriber. From that moment **every** subscription on the node starves: `newPendingTransactions`, `newHeads`, `logs`, transaction receipts, and state-sync deposits (they all share the loop, so the starvation crosses subscription types). The failure is deceptive because `eth_subscribe` keeps returning valid subscription IDs — the install channel interleaves between blocked sends — while delivery trickles at the pace of the slowest client.

Per the repo's own threat-model framing this is an RPC-user-triggerable DoS on a public endpoint: any single WS client can, accidentally or deliberately, suppress subscription delivery for all other clients of the node.

## Production impact

We operate large Polygon PoS RPC infrastructure. On a mainnet full node (bor v2.9.0, 200 peers, at chain tip, txpool ingesting ~64 tx/s throughout), `newPendingTransactions` subscribers received **2–13 hashes per 15s instead of ~900** for several days. Restarting bor did not help — the offending client auto-reconnected through the load balancer and re-wedged the fresh process; the node recovered only when an LB restart severed all client sessions. We reported the symptom ("progressive mempool starvation on subscribe") through the operator channel in April without a reproduction; this PR includes the reproduction that was missing.

## Fix

Insert a bounded queue between each subscription's event feed and the client write (`notifyAsync` / `queueNotification` in `api.go`): enqueueing never blocks, and a per-subscription goroutine drains the queue into `notifier.Notify`. A client that falls more than `clientNotificationBuffer` (512) notifications behind loses subsequent notifications **for itself only**.

Deliberate properties of this approach:

- **`EventSystem` semantics are untouched.** In-process subscribers keep guaranteed, ordered, blocking delivery — all existing `eth/filters` tests pass unmodified. The isolation boundary sits exactly where the untrusted party (the RPC client) attaches.
- The subscription goroutines now always drain their channels promptly, so the shared loop can never be held hostage by a connection.
- Trade-off, stated explicitly: a genuinely-backlogged client silently misses notifications instead of freezing the node's subscription system for everyone (and previously, deadlocking healthy unsubscribes against the wedged loop). This matches the delivery semantics Polygon's Erigon already has for the same surface (`rpc/rpchelper`'s `chan_sub.Send` drops on overflow), so the two clients become consistent under a slow consumer.

Alternatives considered: per-send timeouts in the fan-out loop (retains head-of-line blocking for the timeout duration, multiplied across subscribers); dropping at the `EventSystem` layer (breaks the guaranteed-delivery contract that `TestBlockSubscription` and `TestTransactionReceiptsSubscription` correctly encode for in-process consumers — rejected after trying it); relying on the RPC write deadline (already insufficient in practice — a trickling client never trips it).

## Testing

- New regression test `TestSlowClientDoesNotStarveOtherSubscribers`: a raw-pipe client subscribes and then stops reading; a healthy in-proc client must still receive all 200 events promptly. On current `develop` it fails with `got 129 of 200 events` — exactly the stalled subscriber's channel buffer (128) plus one in-flight before the shared loop froze. With this change it passes in ~1s.
- Full `eth/filters` suite passes, including with `-race` (29/29).
- Field verification: we canaried this fix on the affected production mainnet node. Delivery returned to parity with a healthy sibling (~1,300 notifications/15s each), and deliberately re-running the failure scenario — three concurrently wedged subscribers — left a healthy subscriber completely unaffected (~60 tx/s throughout). Multi-day soak shows no regressions.

Happy to adjust details (queue size, a metrics counter for dropped notifications, drop-oldest vs drop-newest) if maintainers prefer — the property we need is that one slow client cannot affect other subscribers.

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -209,6 +209,45 @@ func (api *FilterAPI) NewPendingTransactionFilter(fullTx *bool) rpc.ID {
 // NewPendingTransactions creates a subscription that is triggered each time a
 // transaction enters the transaction pool. If fullTx is true the full tx is
 // sent to the client, otherwise the hash is sent.
+// clientNotificationBuffer bounds how many notifications may sit queued for a
+// single subscription client. The filter EventSystem fans events out to every
+// subscription from one shared goroutine with blocking sends, so client
+// delivery must never back-pressure into it: a WebSocket client that stops
+// reading (or reads very slowly) would otherwise stall the shared fan-out
+// loop and starve every other subscription on the node. Once a client falls
+// this far behind, further notifications are dropped for that client only.
+const clientNotificationBuffer = 512
+
+// notifyAsync returns a queue drained into notifier.Notify by a background
+// goroutine until stop is closed. Use queueNotification to enqueue: it never
+// blocks, isolating the caller (and transitively the shared event fan-out
+// loop) from slow clients.
+func notifyAsync(notifier *rpc.Notifier, id rpc.ID, stop <-chan struct{}) chan<- any {
+	queue := make(chan any, clientNotificationBuffer)
+
+	go func() {
+		for {
+			select {
+			case v := <-queue:
+				_ = notifier.Notify(id, v)
+			case <-stop:
+				return
+			}
+		}
+	}()
+
+	return queue
+}
+
+// queueNotification enqueues v for asynchronous delivery to the client,
+// dropping it if the client has fallen clientNotificationBuffer behind.
+func queueNotification(queue chan<- any, v any) {
+	select {
+	case queue <- v:
+	default:
+	}
+}
+
 func (api *FilterAPI) NewPendingTransactions(ctx context.Context, fullTx *bool) (*rpc.Subscription, error) {
 	notifier, supported := rpc.NotifierFromContext(ctx)
 	if !supported {
@@ -222,6 +261,10 @@ func (api *FilterAPI) NewPendingTransactions(ctx context.Context, fullTx *bool) 
 		pendingTxSub := api.events.SubscribePendingTxs(txs)
 		defer pendingTxSub.Unsubscribe()
 
+		stop := make(chan struct{})
+		defer close(stop)
+		queue := notifyAsync(notifier, rpcSub.ID, stop)
+
 		chainConfig := api.sys.backend.ChainConfig()
 
 		for {
@@ -234,9 +277,9 @@ func (api *FilterAPI) NewPendingTransactions(ctx context.Context, fullTx *bool) 
 				for _, tx := range txs {
 					if fullTx != nil && *fullTx {
 						rpcTx := ethapi.NewRPCPendingTransaction(tx, latest, chainConfig)
-						_ = notifier.Notify(rpcSub.ID, rpcTx)
+						queueNotification(queue, rpcTx)
 					} else {
-						_ = notifier.Notify(rpcSub.ID, tx.Hash())
+						queueNotification(queue, tx.Hash())
 					}
 				}
 			case <-rpcSub.Err():
@@ -297,10 +340,14 @@ func (api *FilterAPI) NewHeads(ctx context.Context) (*rpc.Subscription, error) {
 		headersSub := api.events.SubscribeNewHeads(headers)
 		defer headersSub.Unsubscribe()
 
+		stop := make(chan struct{})
+		defer close(stop)
+		queue := notifyAsync(notifier, rpcSub.ID, stop)
+
 		for {
 			select {
 			case h := <-headers:
-				notifier.Notify(rpcSub.ID, h)
+				queueNotification(queue, h)
 			case <-rpcSub.Err():
 				return
 			}
@@ -329,11 +376,16 @@ func (api *FilterAPI) Logs(ctx context.Context, crit FilterCriteria) (*rpc.Subsc
 
 	go func() {
 		defer logsSub.Unsubscribe()
+
+		stop := make(chan struct{})
+		defer close(stop)
+		queue := notifyAsync(notifier, rpcSub.ID, stop)
+
 		for {
 			select {
 			case logs := <-matchedLogs:
 				for _, log := range logs {
-					notifier.Notify(rpcSub.ID, &log)
+					queueNotification(queue, &log)
 				}
 			case <-rpcSub.Err(): // client send an unsubscribe request
 				return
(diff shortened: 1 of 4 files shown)


pratikspatil024:
codegenie review

mukul3097:
Pushed a commit addressing the overflow contract.

**Unobservable drops.** Fixed. Overflow now drops the *subscription* rather than the payload: the queue and its drain goroutine live in a `clientNotifier` that closes a `failed` channel on either a full queue or a `Notify` error. Each subscription loop selects on it and returns, unsubscribing from the `EventSystem`. `Notify` errors are no longer discarded, and drops increment `rpc/subscription/dropped` and log the reason, so a slow consumer is distinguishable from a node that stopped producing events.

Note the current `rpc.Notifier` API has no way to hand a server-side error to a subscriber, so the observable surface is metric + log + teardown. Happy to extend `rpc` instead if the preference is for the client to see an explicit error.

**Test install race.** Real, fixed. The test now waits for the subscription to be installed in the `EventSystem` before sending the burst, instead of relying on the `time.Sleep` other tests in this package use.

**"Test cannot fail pre-fix".** This one does not reproduce. With the fix reverted and the test otherwise unchanged, it fails deterministically 3/3:

```
--- FAIL: TestSlowClientDoesNotStarveOtherSubscribers
    api_slow_client_test.go:117: healthy subscriber starved by stalled client: got 128 of 200 events
```

The healthy subscriber starves well inside the 5s deadline, before the ~10s RPC write timeout can tear down the stalled client. With the fix applied it passes 20/20 under `-race`. Also added a unit test covering the overflow policy itself.

**`&log` in the Logs subscription.** This line is unchanged from `develop` — `notifier.Notify(rpcSub.ID, &log)` over `[]*types.Log` predates this PR, and only the call target changed here. With `go 1.26.3` loop variables are per-iteration, so there is no aliasing across enqueued pointers.

One request: the workflow runs on this fork PR need approval, so only the Socket Security checks have executed. The unit test, lint, Diffguard and Sonar gates have not run against these commits yet.

pratikspatil024:
@claude review

pratikspatil024:
codegenie review

eth/filters/api.go
```diff
@@ -206,6 +208,81 @@ func (api *FilterAPI) NewPendingTransactionFilter(fullTx *bool) rpc.ID {
...
+	failOnce sync.Once
+}
+
+// notifyAsync returns a clientNotifier whose queue is drained into
+// notifier.Notify by a background goroutine. That goroutine exits when stop is
+// closed or when delivery fails, so it never outlives the subscription.
+func notifyAsync(notifier *rpc.Notifier, id rpc.ID, stop <-chan struct{}) *clientNotifier {
+	c := &clientNotifier{
+		id:     id,
+		queue:  make(chan any, clientNotificationBuffer),
```
pratikspatil024, line 243: This bounds each queue but not the number of queues. Every accepted public `eth_subscribe` call now eagerly allocates a 512-entry `chan any` (about 8 KiB of interface slots before payload retention) and another goroutine, while the RPC layer has no per-connection or global subscription budget. A client can therefore create unbounded idle subscriptions and turn this fix into a deterministic memory/scheduler amplification path; `fullTx` streams can retain substantially more through queued payload objects. Please enforce a subscription budget before allocating this state (and ideally bound queued bytes, not only item count), or use a shared/lazy dispatcher, with an exhaustion test.

eth/filters/api.go
```diff
@@ -206,6 +208,81 @@ func (api *FilterAPI) NewPendingTransactionFilter(fullTx *bool) rpc.ID {
...
+		failed: make(chan struct{}),
+	}
+
+	go func() {
+		for {
+			select {
+			case v := <-c.queue:
+				if err := notifier.Notify(id, v); err != nil {
```
pratikspatil024, line 251: The 512-item bound is bypassed before the RPC subscription is activated. This drain goroutine starts immediately, but `rpc.Notifier.Notify` appends to the notifier’s unbounded `buffer` while `activated == false` (`rpc/subscription.go:150-154`). If the client delays reading the subscribe response while events arrive, notifications are continuously moved out of this bounded queue into that unbounded buffer. Please either delay draining until activation succeeds or enforce the same bound inside `rpc.Notifier`, and add a codec test that blocks the initial subscription response while publishing more than 512 events.

eth/filters/bor_api.go
```diff
@@ -72,14 +72,21 @@ func (api *FilterAPI) NewDeposits(ctx context.Context, crit ethereum.StateSyncFi
...
 		for {
 			select {
 			case h := <-stateSyncData:
 				if h != nil && (crit.ID == h.ID || crit.Contract == h.Contract ||
 					(crit.ID == 0 && crit.Contract == common.Address{})) {
-					notifier.Notify(rpcSub.ID, h)
+					client.send(h)
 				}
+			case <-client.failed:
+				stateSyncSub.Unsubscribe()
```
pratikspatil024, line 88: This overflow teardown can deadlock the shared `EventSystem`. After `client.failed`, this goroutine stops reading `stateSyncData` and calls `stateSyncSub.Unsubscribe()` synchronously. Generic `Subscription.Unsubscribe` drains `logs`, `txs`, `headers`, and `receipts` while waiting to submit the uninstall request, but it does not drain `stateSyncData` (`filter_system.go:306-313`). If `handleStateSyncEvent` is blocked sending the next deposit (`bor_filter_system.go:11-14`), the event loop cannot receive the uninstall request and this call cannot unblock it, freezing every subscription type again. Please drain `stateSyncData` in generic unsubscribe (and centralize teardown with a defer), then add an overflow/unsubscribe race test proving other subscribers continue to receive events.

pratikspatil024 (changes requested):
Requesting changes for three additional blockers not covered by the existing bot threads. Codegenie/Claude already raised the client-invisible RPC teardown and the >512 healthy-batch overflow, so I have not duplicated those comments. These findings are in addition to those open threads. I also confirmed the current head remains bbe70095e4b5c206da52465e9d5e80f9bfe863f8; the eth/filters race suite passes locally, while the required diffguard check is currently failing.

PR #2336, opened 2026-07-31
accounts/keystore: serialize scans with deletion

pratikspatil024:
## Summary

The nightly race run [30603502369](https://github.com/0xPolygon/bor/actions/runs/30603502369) failed in `accounts/keystore.TestWalletNotifications` because a filesystem scan could race with `KeyStore.Delete`: the scan could read an account before deletion, then reinsert it into the cache after the file and cached entry were removed. This left a stale wallet and emitted an out-of-order `WalletArrived` event.

This behavior is inherited from upstream go-ethereum and is tracked in [ethereum/go-ethereum#29830](https://github.com/ethereum/go-ethereum/issues/29830) and the open fix attempt [ethereum/go-ethereum#30053](https://github.com/ethereum/go-ethereum/pull/30053). The fix serializes complete account scans with file deletion inside `accountCache`, so either the scan completes before deletion or it observes the post-deletion filesystem state. The strict event-order assertion remains unchanged. A regression test covers the error path where a key file disappears after successful key lookup.

## Executed tests

- Before the fix: `go test -race ./accounts/keystore -run '^TestWalletNotifications$' -count=200` reproduced two failures.
- After the fix: the same command passed 200/200 iterations.
- `go test -race ./accounts/keystore -shuffle=1785471185402047521 -count=10`
- `go test -race ./accounts/keystore -run '^TestDeleteFileDisappearsAfterKeyLookup$' -count=20`
- `go test ./accounts/keystore`
- `go vet ./accounts/keystore`
- `make lint` (`0 issues`)
- `diffguard --base origin/develop .` (100% mutation score, 10/10 killed)

`gosec ./accounts/keystore/...` could not build a complete SSA representation with the repository's Go 1.26 toolchain. Its reported findings were pre-existing and outside the changed lines.

## Rollout notes

This is a backward-compatible keystore synchronization fix. It does not affect consensus, configuration, key formats, or persisted chain data. A deletion may briefly wait for an in-progress keystore scan; no coordinated rollout is required.

### DIFF
--- a/accounts/keystore/account_cache.go
+++ b/accounts/keystore/account_cache.go
@@ -73,6 +73,7 @@ type accountCache struct {
 	throttle *time.Timer
 	notify   chan struct{}
 	fileC    fileCache
+	scanMu   sync.Mutex
 }
 
 func newAccountCache(keydir string) (*accountCache, chan struct{}) {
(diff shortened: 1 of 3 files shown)


pratikspatil024:
codegenie review

PR #2341, opened 2026-08-03
eth, eth/downloader: implement EIP-7975 (eth/70 partial receipt lists)

pratikspatil024:
> [!IMPORTANT]
> **Reviewer guide — stacked PR 7 of 12.** Part of the combined go-ethereum **v1.17.4 + v1.17.5** upstream sync, which ships as **one stable release**. Every PR in the stack merges into the base branch `upstream-merge-v1.17.4`; that base merges into `develop` once, at the very end — not per-PR.
>
> **Merge-commit only — never squash.** Squashing rewrites a branch's SHAs and breaks every PR stacked above it.
>
> **Review bottom-up:** #2308 → #2319 → #2325 → #2328 → #2337 → #2340 → #2341 → #2342 → #2343 → #2345 → #2346 → #2354. Start at #2308 / #2319 — every PR above inherits them, so reviewing top-down means re-reviewing.
>
> **Expected-red / flaky checks (not code blockers):** `Quality metrics` (diffguard — skipped by team decision; it also mis-scopes across a stacked diff, comparing against the bottom of the stack), and `codecov/project` (repo-wide coverage threshold; per-PR patch coverage is green). Kurtosis e2e occasionally flakes (~1-in-5, devtools-owned) and is re-run by hand. Full per-batch conflict-resolution reasoning is in `docs/upstream-merges/`.

---

## Summary

Adopts upstream **#33153** (`965bd6b6a`), EIP-7975 / eth/70 partial block receipt
lists, which the previous PR in the stack deferred. This is a **hand-written port
onto Bor's diverged receipt code, not a merge** — which is exactly why it gets its
own PR rather than being folded into a merge batch.

eth/70 lets a peer answer a receipt request with *part* of a block when the
response would otherwise exceed the packet limit. The requester buffers the
partial lists in the peer and resumes from the next receipt index using the same
request ID; nothing reaches the downloader until the block is whole again.

Taken now, immediately after batch 20, so the tree still matches the state
upstream's commit expected — every further batch would have made this a
forward-port instead of an application.

| | |
| --- | --- |
| Files | 26 (incl. 1 new test file) |
| Diff | +1203 / −160 |
| Protocol versions | `[ETH70, ETH69, ETH68]` (Bor keeps eth/68, having declined #33511; upstream is `[ETH70, ETH69]`) |

### The deferral analysis was wrong, in our favour

The batch-20 note recorded a five-step adoption route and flagged a
consensus-relevant design decision as the blocker. Reading the code rather than
trusting the note, two steps evaporated and the blocker turned out not to exist:

1. **No `ReceiptList70` is needed.** EIP-7975 changes the packet envelope, not the
   receipt-list encoding, so eth/70 carries `ReceiptList69`. Upstream likewise
   reuses its single `ReceiptList` for both `ReceiptsPacket69` and
   `ReceiptsPacket70`.
2. **"What does state-sync exclusion mean for a partial list?" does not arise.**
   Bor never applies `ExcludeStateSyncReceipt()` in the p2p handler.
   `handleReceipts` passes a deliberately nil `metadata` func because it has no
   block number; the Madhugiri-gated exclusion runs later in the downloader queue
   (`eth/downloader/bor_fetchers_concurrent_receipts.go` →
   `EncodeReceiptsAndPrepareHasher`). Since partial lists are never sunk to the
   downloader, the exclusion only ever sees a reassembled complete list —
   unchanged from today.

### Consensus-adjacent decisions — please review these first

| Area | Decision |
| ---- | -------- |
| **State-sync receipt across a chunk boundary** | The exclusion is positional (last element) and the serving side identifies the state-sync receipt by *absolute* index. So the encoder keeps its loop counter absolute when receipts are skipped, and advances the tx-type iterator in lockstep. A truncated response therefore cannot displace it. Covered by `TestPartialReceipts_StateSyncAcrossChunkBoundary`, which splits every fixture at every size limit that splits it and asserts the reassembled receipt root on **both** sides of the Madhugiri gate. |
| **Amsterdam gate on the response bound** | Upstream keys the reduced minimum transaction gas (4500 vs 21000) off `AmsterdamTime`. Bor has no Amsterdam timestamp, so this is rewired to block-based `IsAmsterdam(number)`, and `RequestReceipts` carries block **numbers** where upstream carries timestamps. `AmsterdamBlock` is nil, so the 21000 bound is what applies today; both sides of the gate are pinned by `TestValidateLastBlockReceiptAmsterdamGate`. **No fork gate flipped.** |
| **Receipt-count bound vs Bor's free state-sync tx** | Upstream bounds a truncated response by `receipts <= gasUsed/minTxGas`. A Bor block carries one receipt *more* than its gas can account for, because the state-sync transaction burns none, so the bound gained a `+1` tolerance. Without it, a peer serving a legitimate Bor block could be rejected. |

### Deliberate scope beyond a straight port

Two existing eth/69 functions were refactored rather than cloned. Both came out
of running the Bor-only twin scan against my own diff, and both widen the eth/69
serving-path diff — flagging them because they are the parts most likely to want
splitting out.

- **`blockReceiptsToNetwork69` gained a `receiptQueryParams` argument** instead of
  being cloned into a `…70`. The first cut had the clone; it differed from its
  sibling only by the bounds, which is precisely the silent-drift pattern the twin
  scan exists to catch. Upstream has one parameterised function, so this converges
  rather than diverges. The zero value reproduces the previous whole-block
  behaviour. **One deliberate behaviour change on the eth/69 path**, aligned with
  upstream: a receipt with no corresponding body transaction is now an error
  rather than being encoded as type 0 — reachable only from locally corrupt data,
  and the caller skips the block.
- **`gatherBlockReceipts` extracted from `ServiceGetReceiptsQuery69`.** The
  alternative was duplicating ~70 lines of pre-Madhugiri state-sync merge logic
  into the eth/70 service function, i.e. a second new twin.

### Divergences from upstream, kept on purpose

- Bor's `GetReceiptsPacket` is **not** renamed to `…69` as upstream did, because
  in Bor it serves both eth/68 and eth/69.
- The eth/70 size limit reuses Bor's existing `maxMessageSize` (10 MiB) rather
  than adding upstream's duplicate `maxPacketSize` constant.
- The `eth70` handler map mirrors Bor's `eth69` map, which keeps
  `NewBlockHashesMsg`/`NewBlockMsg`; upstream's does not.
- Two small defensive additions over upstream: the buffer entry is dropped when
  `dispatchRequest` fails (upstream leaks it — its own TODO acknowledges the
  concern), and `bufferReceipts` rejects a response claiming more blocks than were
  requested before indexing into `gasUsed`.

### Documentation

This PR also carries the **v1.17.3 milestone documentation for batch 20** as well
as its own, since docs are excluded from the pure merge commits:
`docs/upstream-merges/v1.17.4/{plan,ledger,fork-register,needs-wiring}.md`.
`needs-wiring.md`'s EIP-7975 row moves `deferred` → `adopted` and records what the
original route got wrong; `fork-register.md` moves `defer` → `adopt`.

## Executed tests

New coverage in `eth/protocols/eth/receipt70_test.go`:

- **Chunked reassembly** — for every fixture, every size limit that splits the
  block, asserting the reassembled list is byte-identical to the whole-block
  response in both network and storage encodings. Fails the test if no limit ever
  splits, so it cannot silently go vacuous.
- **State-sync across a chunk boundary** — receipt root verified on both sides of
  the Madhugiri gate.
- First-receipt-too-large, `firstIndex` resumption, `Append`, `LogsSize`, both
  response bounds, and the Amsterdam gate.

Plus an `ETH70` case in `testGetBlockReceipts` covering the full handler round
trip and a resumed (`FirstBlockReceiptIndex: 1`) request.

Other gates, beyond CI's standard set:

- `go build ./...`, `go vet ./...` — clean apart from the two pre-existing
  `//nolint` copylocks; `gofmt -l` clean; `go mod tidy` a no-op.
- `go test` green on `eth/protocols/{eth,snap,wit}`, `eth/downloader`,
  `eth/downloader/whitelist`, `eth`, `eth/fetcher`.
- Both fork meta-guards green: `TestReinforceMultiClientPreCompilesTest`,
  `TestV2ForkParity`.

**Known coverage gap, stated rather than hidden:** the truncation path only fires
on blocks whose receipts exceed 10 MiB. No devnet produces those, so the split
behaviour is exercised by unit tests over synthetic size limits, not end to end.
It matters most on the largest blocks — plausibly more on Polygon than on L1 given
our gas limits and transaction counts. Upstream's `cmd/devp2p/internal/ethtest`
eth/70 suite and its regenerated testdata were **not** ported: the testdata is
specific to the geth test chain, and Bor's `ethtest` is already a known-failing
surface.

## Rollout notes

- **Not consensus-affecting.** eth/70 is a wire protocol version negotiated per
  peer, not a state transition. The receipt root derivation path is untouched: the
  Madhugiri-gated exclusion still runs on a complete list, in the downloader
  queue, exactly as before.
- **Backwards-compatible.** Bor advertises eth/68, eth/69 and eth/70; peers on
  older versions negotiate down and take the unchanged code paths.
- No coordinated upgrade required, and no operator-facing change.
- The one fork-gated line introduced (reduced minimum tx gas) is dormant —
  `AmsterdamBlock` is nil on every Bor preset.

## Stacked PR — do not squash

Part of the v1.17.4 upstream sync stack. **Squash-merging any PR in this stack
breaks every PR above it**, because squashing rewrites commits into new SHAs, so
the PRs above would re-show all of this PR's changes and conflict against their
base. Team standard for upstream syncs is a merge commit, never squash.

```
upstream-merge-v1.17.4                (base)
 └ ppatil-upstream-v1.16.9      #2308
    └ ppatil-upstream-v1.17.0   #2319
       └ ppatil-upstream-v1.17.1 #2325
          └ ppatil-upstream-v1.17.2 #2328
             └ ppatil-corevm-catchup #2337
                └ ppatil-upstream-v1.17.3  (v1.17.3 part 1, batch 20)
                   └ ppatil-upstream-eth70 <-- THIS PR
                      └ v1.17.3 part 2     (batches 21-26 + milestone chores)
```

Reviews are deliberately not being requested yet — the sync is mid-flight and
several decisions are still open. This will be marked ready once all six
milestones are complete.


pratikspatil024:
@claude review

pratikspatil024:
codegenie review

eth/protocols/eth/peer.go
```diff
@@ -459,6 +526,155 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+
+	// Verify that the number of receipts delivered is one the block's gas could pay for.
+	minTxGas := params.TxGas
+	if p.chainConfig != nil && p.chainConfig.IsAmsterdam(new(big.Int).SetUint64(number)) {
+		minTxGas = minTxGasAmsterdam
+	}
+	// A Bor block carries one extra receipt for the state-sync transaction, which burns
+	// no gas and so is not accounted for by the gas-derived bound.
+	if uint64(previousTxs+lastReceipts.Len()) > gasUsed/minTxGas+1 {
+		// Drop the response but keep the buffer, the peer may still be dropped instead.
```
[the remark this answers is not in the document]
pratikspatil024: Correct, fixed.

`validateLastBlockReceipt` said "Drop the response but keep the buffer", but its only caller, `bufferReceipts`, does `delete(p.receiptBuffer, requestId)` for any error it returns — so the buffer is always dropped, not kept. The comment is Bor's own (upstream has no equivalent line at `965bd6b6a:eth/protocols/eth/peer.go:541`), so nothing upstream is being contradicted by correcting it.

Replaced with what actually happens:

```go
// bufferReceipts drops the buffer entry for any error returned here.
return 0, errors.New("total number of tx exceeded limit")
```

---
Landed in `b246fba3d`.

eth/protocols/eth/peer.go
```diff
@@ -459,6 +526,155 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+		sink: nil,
+		code: GetReceiptsMsg,
+		want: ReceiptsMsg,
+		data: &GetReceiptsPacket70{
+			RequestId:              id,
+			FirstBlockReceiptIndex: uint64(lastReceipt),
+			GetReceiptsRequest:     hashes,
+		},
+	})
+}
```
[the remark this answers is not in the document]
pratikspatil024: Confirmed and fixed. Both of you flagged this one, and it is the most serious of the five.

The cycle is exactly as described:

- `requestPartialReceipts` took `receiptBufferLock` with a `defer` unlock and then called `dispatchRequest` while still holding it.
- `dispatchRequest` sends on the unbuffered `p.reqDispatch` and then blocks on `<-reqOp.fail`, both serviced by the peer's single `dispatcher()` goroutine.
- `dispatcher()`'s `case cancelOp := <-p.reqCancel:` branch takes `receiptBufferLock` (`dispatcher.go:241`) for **any** cancelled request on the peer, receipt or not.

So if `select` picks the cancel branch — an unrelated header or body request timing out is enough — the dispatcher blocks on the mutex while `requestPartialReceipts` holds it waiting for that same dispatcher to drain `reqDispatch`. Neither side moves, and the peer's only dispatcher goroutine is wedged for the peer's lifetime.

The sibling settles that this is a bug rather than a deliberate ordering: `RequestReceipts` already unlocks before dispatching, for this reason.

Fixed by computing under the lock, releasing it, then dispatching:

```go
lastBlock := len(buffer.list) - 1
lastReceipt := buffer.list[lastBlock].Len()
// Copy rather than alias: the dispatch below runs unlocked, and a concurrent
// cancellation may drop this buffer entry while it is in flight.
hashes := append(GetReceiptsRequest(nil), buffer.request[lastBlock:]...)
p.receiptBufferLock.Unlock()
```

The copy matters: `buffer.request[lastBlock:]` aliases the entry's backing array, and once the lock is released a concurrent cancel can drop that entry — the previous code was safe only because it never let go of the lock.

Worth recording that **upstream has the same shape** (`965bd6b6a:eth/protocols/eth/peer.go:409`, dispatching under the lock, with the same `receiptBufferLock` acquisition on the cancel path). This is not a Bor porting error. I have fixed it here anyway rather than mirroring it, because it changes no wire behaviour and no observable protocol semantics — it is purely lock discipline inside our own process — so it costs nothing in convergence terms while removing a permanent peer hang.

Verified: `eth/protocols/eth` and `eth` green, and `eth/protocols/eth` green under `-race`.

---
Landed in `b246fba3d`.

eth/protocols/eth/peer.go
```diff
@@ -459,6 +526,155 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+
+// bufferReceipts validates a receipt packet and buffers it while the last block of the
+// response is still incomplete. Once the response completes, the previously collected
+// receipts are merged back in.
+func (p *Peer) bufferReceipts(requestId uint64, receiptLists []*ReceiptList69, lastBlockIncomplete bool) error {
+	p.receiptBufferLock.Lock()
+	defer p.receiptBufferLock.Unlock()
+
```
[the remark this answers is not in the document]
pratikspatil024: Confirmed and fixed. Both of you flagged this one, and it is the most serious of the five.

The cycle is exactly as described:

- `requestPartialReceipts` took `receiptBufferLock` with a `defer` unlock and then called `dispatchRequest` while still holding it.
- `dispatchRequest` sends on the unbuffered `p.reqDispatch` and then blocks on `<-reqOp.fail`, both serviced by the peer's single `dispatcher()` goroutine.
- `dispatcher()`'s `case cancelOp := <-p.reqCancel:` branch takes `receiptBufferLock` (`dispatcher.go:241`) for **any** cancelled request on the peer, receipt or not.

So if `select` picks the cancel branch — an unrelated header or body request timing out is enough — the dispatcher blocks on the mutex while `requestPartialReceipts` holds it waiting for that same dispatcher to drain `reqDispatch`. Neither side moves, and the peer's only dispatcher goroutine is wedged for the peer's lifetime.

The sibling settles that this is a bug rather than a deliberate ordering: `RequestReceipts` already unlocks before dispatching, for this reason.

Fixed by computing under the lock, releasing it, then dispatching:

```go
lastBlock := len(buffer.list) - 1
lastReceipt := buffer.list[lastBlock].Len()
// Copy rather than alias: the dispatch below runs unlocked, and a concurrent
// cancellation may drop this buffer entry while it is in flight.
hashes := append(GetReceiptsRequest(nil), buffer.request[lastBlock:]...)
p.receiptBufferLock.Unlock()
```

The copy matters: `buffer.request[lastBlock:]` aliases the entry's backing array, and once the lock is released a concurrent cancel can drop that entry — the previous code was safe only because it never let go of the lock.

Worth recording that **upstream has the same shape** (`965bd6b6a:eth/protocols/eth/peer.go:409`, dispatching under the lock, with the same `receiptBufferLock` acquisition on the cancel path). This is not a Bor porting error. I have fixed it here anyway rather than mirroring it, because it changes no wire behaviour and no observable protocol semantics — it is purely lock discipline inside our own process — so it costs nothing in convergence terms while removing a permanent peer hang.

Verified: `eth/protocols/eth` and `eth` green, and `eth/protocols/eth` green under `-race`.

---
Landed in `b246fba3d`.

eth/protocols/eth/handlers.go
```diff
@@ -359,116 +456,76 @@ func ServiceGetReceiptsQuery69(chain *core.BlockChain, query GetReceiptsRequest)
...
-		var blockReceipts []*types.ReceiptForStorage = make([]*types.ReceiptForStorage, 0)
-		if len(normalReceipts) != 0 {
-			blockReceipts = append(blockReceipts, normalReceiptsDecoded...)
-		}
-		if len(borReceipt) != 0 {
-			isBorReceiptPresent = true
-			blockReceipts = append(blockReceipts, &borReceiptDecoded)
-		}
+	return receipts
+}
```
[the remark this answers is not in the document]
pratikspatil024: Confirmed and fixed.

`ServiceGetReceiptsQuery70`'s loop guard was `bytes >= softResponseLimit || len(receipts) >= maxReceiptsServe`, and all three skip paths — `gatherBlockReceipts` miss, nil body, conversion error — `continue` without touching either counter. So a query of hashes that all miss costs one `gatherBlockReceipts` plus one `GetBodyRLP` per hash, bounded only by the 10 MB message cap, in the p2p handler goroutine. Your comparison against the siblings is right: `ServiceGetReceiptsQuery68` (`handlers.go:328`) and `ServiceGetReceiptsQuery69` (`handlers.go:455`) both carry `lookups >= 2*maxReceiptsServe` precisely for this case.

Added the same bound:

```go
// The lookup bound matters more here than in the size and count bounds
// above: a hash that misses is skipped without adding to either, so
// without it a peer can make the node pay one database lookup per hash
// for as many hashes as fit in a message. eth/68 and eth/69 bound it the
// same way.
if bytes >= softResponseLimit || len(receipts) >= maxReceiptsServe ||
	i >= 2*maxReceiptsServe {
	break
}
```

One correction to the finding's framing, since it matters for whether this is a regression: **upstream's eth/70 omits the bound too.** `965bd6b6a`'s `serviceGetReceiptsQuery70` guards on only `bytes` and `receipts.Len()`, and skips misses the same way. So this is not something Bor's rewrite dropped — upstream's eth/70 has it as well, while its own 68/69 keep it.

I have still fixed it here, and the convergence argument that applies elsewhere in this sync does not apply: `ServiceGetReceiptsQuery70` is already a Bor-authored function rather than carried-in upstream code — different signature, different return type, `gatherBlockReceipts`, bor receipts and the state-sync receipt — so adding a bound is hardening our own code, not introducing a divergence into an otherwise byte-identical file. It also brings the function in line with its two Bor siblings, and `.claude/rules/p2p-security.md` requires bounded work from untrusted peer input.

Verified: `eth/protocols/eth`, `eth` and `eth/downloader` green.

---
Landed in `b246fba3d`.

eth/protocols/eth/peer.go
```diff
@@ -459,6 +526,155 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+		lastBlock := len(receiptLists) - 1
+		if len(buffer.list) > 0 {
+			lastBlock += len(buffer.list) - 1
+		}
+		if lastBlock >= len(buffer.gasUsed) {
+			delete(p.receiptBuffer, requestId)
+			return errors.New("receipt response longer than the request")
+		}
+		logSize, err := p.validateLastBlockReceipt(receiptLists, requestId, buffer.gasUsed[lastBlock], buffer.numbers[lastBlock])
+		if err != nil {
```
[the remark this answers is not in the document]
pratikspatil024: The analysis is correct, and it is the one of the three I am **not** fixing in this PR. Reasoning below, because the call is debatable and I would rather it be visible than silent.

**The mechanism is real.** `ServiceGetReceiptsQuery70` `continue`s past an unservable hash, so the response list is not positionally 1:1 with the query, while `bufferReceipts` derives `lastBlock` from `len(receiptLists)-1` (+ `len(buffer.list)-1`) and `requestPartialReceipts` slices `buffer.request[lastBlock:]`. Both treat a response-relative length as an absolute index into the request-ordered `gasUsed`/`numbers`/`request` arrays, so one skipped block shifts every subsequent index and `validateLastBlockReceipt` bounds the truncated chunk against the wrong block's gas.

**But it is upstream's design, carried over faithfully.** Upstream's `bufferReceipts` (`965bd6b6a:eth/protocols/eth/peer.go:462-465`) derives the index identically:

```go
lastBlock := len(receiptLists) - 1
if len(buffer.list) > 0 {
	lastBlock += len(buffer.list) - 1
}
```

and its `serviceGetReceiptsQuery70` skips unservable hashes with the same two `continue`s. So the positional assumption and the thing that breaks it both come from upstream, not from Bor's adaptation.

**Bor is already strictly safer here than upstream.** Our `bufferReceipts` adds a bounds check upstream does not have:

```go
if lastBlock >= len(buffer.gasUsed) {
	delete(p.receiptBuffer, requestId)
	return errors.New("receipt response longer than the request")
}
```

Upstream indexes `buffer.gasUsed[lastBlock]` unguarded, so the same drift that costs us a wrong-but-in-range bound would panic there. The worst outcome on this branch is a mis-scoped validation bound, not an out-of-range read.

**Why not fix it here.** Unlike the other two, this cannot be fixed locally without touching protocol behaviour. Either the server stops skipping — which changes what a Bor node returns to a geth eth/70 peer and is an interop risk on a wire format we have just adopted — or the response has to carry which indices it served, which is a wire-format change and an EIP-7975 question rather than a Bor one. Neither belongs in an upstream-sync PR, and getting it wrong is worse than the bug: the failure mode today is bounded by downstream receipt-root validation, so no bad data is accepted.

Recording it in `docs/upstream-merges/v1.17.4/needs-wiring.md` against the EIP-7975 row so it is visible at enable time and at the next eth/70 batch, rather than leaving it only in this thread.

If reviewers disagree and would rather carry a client-side fix that derives the absolute index from what was actually requested, say so and I will add it — it is doable without a wire change, just more invasive than I would choose to land unreviewed here.

---
Landed in `b246fba3d`.

cffls (commented):
Two issues on the eth/70 path plus a stack note. Inline comments below.

**Stack note:** #2345, #2346 and #2354 don't contain `b246fba3d` yet (`git merge-base --is-ancestor b246fba3d <head>` fails for all three). At the top of the stack, `requestPartialReceipts` still holds `receiptBufferLock` across the dispatch, and the lookup bound in `ServiceGetReceiptsQuery70` is missing. The fix should survive the final merge since those branches don't touch the lines, but their CI and devnet runs so far were without it. Worth merging this PR's head up the stack.

<details><summary>Repro for the dispatcher race (fails on b246fba3d, passes with the one-line guard)</summary>

```go
package eth

import (
	"crypto/rand"
	"testing"
	"time"

	"github.com/ethereum/go-ethereum/common"
	"github.com/ethereum/go-ethereum/p2p"
	"github.com/ethereum/go-ethereum/p2p/enode"
	"github.com/ethereum/go-ethereum/params"
)

// Reproduces: cancellation lands between requestPartialReceipts' unlock and the
// dispatcher servicing the continuation. The continuation (sink == nil) becomes
// the pending entry; the completing response then blocks forever on a nil sink.
func TestRepro_ContinuationAfterCancelWedgesReader(t *testing.T) {
	app, net := p2p.MsgPipe()
	defer app.Close()
	go func() { // drain outbound requests
		for {
			msg, err := app.ReadMsg()
			if err != nil {
				return
			}
			msg.Discard()
		}
	}()
	var id enode.ID
	rand.Read(id[:])
	p := NewPeer(ETH70, p2p.NewPeer(id, "x", nil), net, nil, params.TestChainConfig)
	defer p.Close()

	sink := make(chan *Response, 1)
	req, err := p.RequestReceipts([]common.Hash{{1}}, []uint64{1_000_000}, []uint64{1}, sink)
	if err != nil {
		t.Fatal(err)
	}
	// First partial response gets buffered.
	if err := p.bufferReceipts(req.id, []*ReceiptList69{{items: []Receipt{{Logs: rlp.EmptyList}}}}, true); err != nil {
		t.Fatal(err)
	}
	// requestPartialReceipts has snapshotted the buffer and released the lock...
	p.receiptBufferLock.Lock()
	hashes := append(GetReceiptsRequest(nil), p.receiptBuffer[req.id].request...)
	p.receiptBufferLock.Unlock()
	// ...the downloader times out and cancels...
	req.Close()
	// ...then the continuation reaches the dispatcher.
	if err := p.dispatchRequest(&Request{id: req.id, code: GetReceiptsMsg, want: ReceiptsMsg,
		data: &GetReceiptsPacket70{RequestId: req.id, FirstBlockReceiptIndex: 1, GetReceiptsRequest: hashes}}); err != nil {
		t.Fatal(err)
	}
	// Completing response, as handleReceipts70 would dispatch it (buffer is gone).
	done := make(chan error, 1)
	go func() {
		lists := []*ReceiptList69{{}}
		done <- p.dispatchResponse(&Response{id: req.id, code: ReceiptsMsg, Res: &lists}, nil)
	}()
	select {
	case <-done:
	case <-time.After(2 * time.Second):
		t.Fatal("dispatchResponse blocked: peer read loop would be wedged on nil sink")
	}
}
```
</details>

eth/protocols/eth/dispatcher.go
```diff
@@ -217,7 +217,11 @@ func (p *Peer) dispatcher() {
 			reqOp.fail <- err
 
 			if err == nil {
-				pending[req.id] = req
+				// An eth/70 continuation re-uses the ID of the request it resumes, so
+				// the original entry has to stay in place.
+				if _, ok := pending[req.id]; !ok {
```
cffls: **The continuation can become the `pending` entry after a cancel, and the next response then blocks forever on a nil sink.**

Since `b246fba3d` (unlock before dispatch), there's a window between `requestPartialReceipts` releasing `receiptBufferLock` and this `reqDispatch` case running:

1. The downloader times out and calls `req.Close()`. The `reqCancel` case deletes `pending[id]` and `receiptBuffer[id]`.
2. The continuation reaches this case. `pending[id]` is gone, so the continuation (`sink: nil`, plus a fresh `Cancel` channel nobody holds) is stored as the pending request.
3. The peer's final response (`LastBlockIncomplete: false`) finds no buffer, so `handleReceipts70` calls `dispatchResponse`. The dispatcher matches it to the continuation and `dispatchResponse` blocks on `res.Req.sink <- res` with a nil `sink` and an unclosable `Cancel`. This peer's message handling is stuck until it disconnects.

Repro in the review body. It fails with "dispatchResponse blocked" on this head.

Minimal fix: never let a continuation become the pending entry:
```go
if _, ok := pending[req.id]; !ok && req.sink != nil {
```
With that, the late response gets `errDanglingResponse` and is dropped. The repro passes and `go test -race ./eth/protocols/eth/` is green.

Upstream fixed the original deadlock differently in ethereum/go-ethereum#35537 (`c57bbe417`, in v1.17.6). It adds a separate `dispatchResend` path that only sends if `pending[id]` still exists, and moves buffer cleanup into a per-request `cleanup` callback. That closes this race too. I'd suggest cherry-picking it in place of the unlock-then-dispatch change, together with ethereum/go-ethereum#35593 (`596096afe`), which rejects truncated responses that add no receipts and fixes the downloader's round-trip estimate across continuations.

eth/protocols/eth/handlers.go
```diff
@@ -359,116 +456,82 @@ func ServiceGetReceiptsQuery69(chain *core.BlockChain, query GetReceiptsRequest)
...
+		if !ok {
 			continue
 		}
 		body := chain.GetBodyRLP(hash)
 		if body == nil {
 			continue
 		}
 
-		results, err := blockReceiptsToNetwork69(encodedBlockReceipts, body, isStateSyncReceipt)
+		q := receiptQueryParams{sizeLimit: uint64(maxMessageSize - bytes)}
```
cffls: **A truncated response can be slightly larger than `maxMessageSize`.**

`sizeLimit` only budgets the receipt data (`bytes` is the sum of the encoded lists). It leaves out the packet wrapper: the outer list header (~4 B), `RequestId` (≤9 B), `LastBlockIncomplete` (1 B) and the header of the `ReceiptsRLPResponse` list (~4 B). `blockReceiptsToNetwork69` can fill right up to `sizeLimit`, so the message can end up a few bytes over 10 MiB. The requester's `handleMessage` then rejects it with `errMsgTooLarge` and drops the peer.

It only happens on blocks with more than 10 MiB of receipts, but the output is the same every time for a given block and `firstIndex`. Every eth/70 peer serving that block would get dropped the same way, which can stall sync at that block.

Upstream has the same issue (`maxPacketSize - bytes` in `serviceGetReceiptsQuery70`). Suggest reserving a fixed allowance, e.g.:
```go
const receiptsPacketOverhead = 64 // outer list, request id, flag and inner list headers
q := receiptQueryParams{sizeLimit: uint64(maxMessageSize - receiptsPacketOverhead - bytes)}
```
plus a test that encodes the full `ReceiptsRLPPacket70` at the truncation limit and asserts `len <= maxMessageSize`.

eth/protocols/eth/dispatcher.go
pratikspatil024: Confirmed, thanks. Your repro fails on `b246fba3d` exactly as described. Fixed in a2a83b9dd by taking upstream's fix rather than the one-line guard: #35537 (`c57bbe417`) is ported, so the continuation goes through a resend path that only sends while the original request is still pending, and the receipt buffer is released by a per-request `cleanup` callback instead of the dispatcher reaching into it on cancel. Bor's dispatcher uses the global `requestTracker` rather than upstream's per-peer tracker, so the port is behavioural rather than line-for-line; the resend path tracks under the original ID the same way.

#35593 (`596096afe`) is ported in the same commit: an incomplete response that delivers no receipt is rejected, and `bor_fetchers_concurrent.go` scales the capacity estimate to one round trip.

Your repro is kept as `TestResendAfterCancelIsDropped`, adapted to the resend path (it asserts the late response neither blocks nor reaches the cancelled sink), plus upstream's `TestBufferReceiptsNoProgress`. `go test -race ./eth/protocols/eth/` is green.

eth/protocols/eth/handlers.go
pratikspatil024: Confirmed. Fixed in a2a83b9dd with `receiptsPacketOverhead = 64` reserved out of `maxMessageSize` before truncating. `TestReceiptsPacket70FitsMessageCap` wraps a payload filled to the limit in `ReceiptsRLPPacket70` with a worst-case request ID and the flag set, and asserts it fits. With the reserve at 0 the same packet is 10485778 bytes, 18 over the cap, so the test is not vacuous.

pratikspatil024:
Thanks @cffls. Both inline items are fixed in a2a83b9dd (replies on each thread).

In a95e9b824, eth/70 is kept implemented and tested but **no longer advertised**: `ProtocolVersions` is `[ETH69, ETH68]`, so no peer negotiates it and `eth_protocolVersion` is unchanged from `develop`. Enabling it is a one-line change, tracked separately.

**Stack note:** agreed. #2342 and #2343 get this head next via the usual cascade, and #2345 / #2346 / #2354 pick it up when they are cascaded in turn.

eth/protocols/eth/peer.go
```diff
@@ -459,6 +536,160 @@ func (p *Peer) RequestReceipts(hashes []common.Hash, sink chan *Response) (*Requ
...
+	// A block already partway through retrieval carries its progress in the buffer:
+	//   [[receipt1], [receipt1, receipt2]], incomplete = true
+	//   [[receipt3, receipt4]],             incomplete = true  <<--
+	//   [[receipt5], [receipt1]],           incomplete = false
+	// which is only reachable with a single-element response over a non-empty buffer.
+	var (
+		previousTxs int
+		previousLog uint64
+	)
+	if buffer, ok := p.receiptBuffer[id]; ok && len(buffer.list) > 0 && len(receiptLists) == 1 {
```
rukimedo: previousTxs/previousLog are only picked up when len(receiptLists) == 1. 

For a continuation packet shaped [A_rest, B_partial], the gas/log bound is checked against B only, and A_rest is appended to the buffered A with no re-validation, so a block spanning a multi-block packet boundary gets one unbounded extension. The reassembled block is still checked against its receipt root later, so this isn't a correctness hole, but it's a hole in the DoS pre-filter that validateLastBlockReceipt exists to be. Is that deliberate / does upstream do the same?

eth/protocols/eth/handlers.go
```diff
@@ -359,116 +456,82 @@ func ServiceGetReceiptsQuery69(chain *core.BlockChain, query GetReceiptsRequest)
...
+		if !ok {
 			continue
 		}
 		body := chain.GetBodyRLP(hash)
 		if body == nil {
 			continue
 		}
 
-		results, err := blockReceiptsToNetwork69(encodedBlockReceipts, body, isStateSyncReceipt)
+		q := receiptQueryParams{sizeLimit: uint64(maxMessageSize - receiptsPacketOverhead - bytes)}
```
rukimedo, line 511: On the pre-Madhugiri path (i.e. mainnet today) gatherBlockReceipts reads the receipts RLP, decodes all of them, decodes the bor receipt, merges, and rlp.EncodeToBytes the whole list,  then blockReceiptsToNetwork69 skips the first firstIndex entries and throws that work away. Serving one block in K chunks is O(K·N) decode+encode, and K is chosen by the requester. Worth either capping continuations per request or writing down why the existing limits make this acceptable.

eth/downloader/bor_fetchers_concurrent.go
```diff
@@ -541,7 +541,14 @@ func (d *Downloader) concurrentFetch(queue typedQueue, beaconMode bool) error {
 				// caused by a timed out request which came through in the end), set it to
 				// idle. If the delivery's stale, the peer should have already been idled.
 				if !errors.Is(err, errStaleDelivery) {
-					queue.updateCapacity(peer, accepted, res.Time)
+					items, elapsed := accepted, res.Time
+					if res.Roundtrips > 1 && items > 0 {
+						// Partial receipt response can be assembled over multiple round trips.
+						// Scale both values down to a single round trip.
+						items = max(1, items/res.Roundtrips)
```
rukimedo, line 548: `Sent` is deliberately not reset on resend, which is the right call for safety, but it means every round trip has to fit inside the same timeoutGracePeriod (2 min) budget. 

A genuinely large block needing many continuations can push an honest peer past it and get it dropped as stalling. The updateCapacity scaling handles throughput accounting but not the deadline. Worth a test with a peer that chunks aggressively.

eth/protocols/eth/receipt.go
```diff
@@ -465,28 +494,54 @@ func blockReceiptsToNetwork69(blockReceipts, blockBody rlp.RawValue, isStateSync
...
 		}
 	}
 	outer := enc.List()
 	for i := 0; it.Next(); i++ {
+		// TxType is always 0 for state-sync transactions before Madhugiri hardfork.
+		// Post Madhugiri HF, they will be part of normal block receipts and body so no special
+		// handling needed.
+		var txType uint64
+		if !isStateSyncReceipt(i) {
+			t, ok := nextTxType()
```
rukimedo, line 507: `nextTxType()`'s ok flag is now checked and returns an error; previously it was txType, _ := and silently encoded type 0. Net effect on eth/69: a block with more receipts than txs is now dropped from the response (error → log.Error → continue) instead of served with a zero type. Likely an improvement, but it changes a shipping path in an eth/70 PR. Worth a line in the description and a test for the receipts>txs case.

rukimedo (commented):
Overall LGTM. Have added a few comments

eth/protocols/eth/handlers.go
pratikspatil024: Checked this against the limits and real block sizes. I think the existing bounds make it acceptable, but agreed it is worth writing down, so it is now a note on the ticket that gates enabling eth/70.

- **A continuation only happens when one block overflows its response.** The server stops adding blocks once it passes `softResponseLimit` (2 MiB), so every block starts with at least ~8.39 MB of room. Truncation needs a single block whose receipts exceed that.
- **The pre-Madhugiri decode/merge/encode path is historical only,** for blocks below 80,084,800 on mainnet. Post-Madhugiri blocks take the raw-RLP branch with no decode or merge. The largest pre-Madhugiri gas limit was 60M, which is at most ~7.5 MB of receipts at 8 gas per byte of log data. So those blocks are never truncated, and that work is never repeated per chunk.
- **At the current 160M gas limit** a block can carry ~20 MB of receipts, so K ≤ 3 per block. K is bounded by the block's gas, not chosen freely by the requester.
- A requester can send any `firstIndex`, but each one is an independent request costing a single pass over the block. That is the same as an eth/69 request for that block, which also walks and encodes all N receipts, so there is no amplification beyond what eth/69 already allows.

None of this is reachable today either: eth/70 is no longer advertised as of a95e9b824.

eth/downloader/bor_fetchers_concurrent.go
pratikspatil024: Agreed on both points. `Sent` is not reset, so every round of a request shares one budget: the request timeout, then `timeoutGracePeriod` measured from `Sent` before the peer is graded as stalling.

To size it: at the current 160M gas limit a block needs at most ~3 rounds of ≤10 MiB (see the handlers.go thread for the bound). An honest peer therefore has to move ~30 MiB in 2 minutes, roughly 2 Mbps. That is only tight for a very slow peer, and it is the case your suggested test should pin.

I have not added the test in this PR, since eth/70 is not negotiated as of a95e9b824. A test with a peer that chunks as aggressively as the limits allow, asserting it is not graded as stalling, is now recorded as a precondition on the ticket that gates enabling eth/70.

eth/protocols/eth/receipt.go
pratikspatil024: Good call. The behaviour change is described under "Deliberate scope beyond a straight port" (a receipt with no corresponding body transaction is now an error, reachable only from locally corrupt data, and the caller skips the block), but it had no test.

Added `TestBlockReceiptsToNetwork69_MoreReceiptsThanTxs` in 25f67f777. It covers both sides: a block with an extra receipt is refused, and a trailing pre-Madhugiri state-sync receipt, which has no body transaction by design, is still served.

rukimedo (approved):

cffls (approved):

PR #2364, opened 2026-08-14
rpc: fix rpc_ep_workers_http being stopped on first request

ChanHongMing:
## Summary

  The server-side HTTP RPC execution pool is shared across all handlers. PR #1005
  added `executionPool.Stop()` to `handler.close()`, so the first HTTP request stops
  the shared pool's metric-reporting goroutine before it ever ticks. As a result
  `rpc_ep_workers_http` (and its `rpc_ep_queue_http` / `rpc_ep_processed_http`
  siblings) is never registered.

  ## Fix

  - Remove `h.executionPool.Stop()` from `handler.close()` — the server pool is
    already stopped by `Server.Stop()`.
  - Add `cc.handler.executionPool.Stop()` to `clientConn.close()` — preserves
    PR #1005's goroutine-leak fix for the client, where the pool is per-connection.

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -124,6 +124,9 @@ func (c *Client) newClientConn(conn ServerCodec) *clientConn {
 }
 
 func (cc *clientConn) close(err error, inflightReq *requestOp) {
+	// The client-side pool is per-connection (created in newClientConn), so it
+	// is owned here and must be stopped to release its metric goroutine.
+	cc.handler.executionPool.Stop()
 	cc.handler.close(err, inflightReq)
 	cc.codec.close()
 }
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -334,7 +334,12 @@ func (h *handler) close(err error, inflightReq *requestOp) {
 	h.callWG.Wait()
 	h.cancelRoot()
 	h.cancelServerSubscriptions(err)
-	h.executionPool.Stop()
+
+	// The execution pool is deliberately not stopped here. On the server side
+	// it is shared across all handlers and owned by Server, which stops it in
+	// Server.Stop(); stopping it per-handler would kill the shared pool's
+	// metric goroutine on the first request (regression introduced in #1005).
+	// On the client side the per-connection pool is stopped by clientConn.close.
 }
 
 // addRequestOp registers a request operation.


pratikspatil024:
@claude review

pratikspatil024:
codegenie review

rpc/client.go
```diff
@@ -124,6 +124,7 @@ func (c *Client) newClientConn(conn ServerCodec) *clientConn {
 }
 
 func (cc *clientConn) close(err error, inflightReq *requestOp) {
+	cc.handler.executionPool.Stop()
```
Sohailghafoor, line 129: Potential issue: this execution pool uses the static "rpcclient" service label, so multiple concurrent WS/IPC connections share the same rpc_ep_*_rpcclient metric gauges. 
Since each connection owns and stops its own pool, closing one connection can also leave the shared gauge with a stale value while other connections are still active.

This appears pre-existing and is not introduced by this PR, so I don't think it should block this change, but it would be worth tracking separately or adding a follow-up issue to make the per-connection metrics distinct or aggregate them correctly.

rpc/handler.go
```diff
@@ -334,7 +334,6 @@ func (h *handler) close(err error, inflightReq *requestOp) {
 	h.callWG.Wait()
 	h.cancelRoot()
 	h.cancelServerSubscriptions(err)
-	h.executionPool.Stop()
 }
```
Sohailghafoor, line 343: Suggestion: could we add a short comment here explaining that executionPool is the shared server-side HTTP pool and is intentionally stopped by Server.Stop() rather than by each handler? 
This ownership distinction is easy to miss and helps prevent reintroducing the first-request metrics regression.

Sohailghafoor (commented):
**Non-blocking:** This appears to be a pre-existing issue, but multiple concurrent WS/IPC connections create independent execution pools with the same "rpcclient" metrics label. Their metric reporters therefore update the same rpc_ep_*_rpcclient gauges, and stopping one connection can leave a stale metric value while another connection is still active.

Could this be tracked as a follow-up to make the per-connection metrics distinct or aggregate them correctly?

rpc/handler.go
ChanHongMing: Comment added as suggested.

pratikspatil024:
@ChanHongMing please could you rebase/merge the latest develop?

ChanHongMing:
> @ChanHongMing please could you rebase/merge the latest develop?

sure, no problem

pratikspatil024:
@claude review

pratikspatil024 (approved):

marcello33 (approved):

pratikspatil024:
@ChanHongMing can you please verify your signatures?

ChanHongMing:
> @ChanHongMing can you please verify your signatures?

Yes, sure. I have verified it.

PR #2373, opened 2026-08-25
eth, core: add sequence store preconfirmation pipeline and RPC support

vbhattaccmu:
## Summary

This PR wires the Sequence Store consumer into Bor’s pending RPC and canonical block-import paths on top of `cffls/sequence-publisher`. Bor treats Sequence Store only as an early source of transaction ordering: every transaction is decoded and executed locally, no state is trusted from Sequence Store, and canonical fork choice remains unchanged.

The consumer maintains a multi-block in-memory `PendingStore` keyed by block number and parent hash. Pending views are published after the first 16 transactions and then at up to 128 gas-progress checkpoints per block, plus the final seal. Canonical import may reuse either an exact sealed execution or the latest compatible transaction prefix; both paths require exact execution-context and transaction matching and pass Bor’s normal state validation. Any miss, mismatch, unsupported execution mode, or validation failure falls back to normal execution.

The consumer also backs the preconfirmation RPC surface:

- `pending` block/state RPCs — `eth_getBlockByNumber`, balance, nonce, calls, simulation, code, storage, logs, and filters — read the locally executed Sequence Store pending view when one is live. Txpool-only transactions are not included in that view.

- `eth_getTransactionByHash` falls back to the live preconfirmation index before the txpool when no canonical or Bor transaction is found.

- `eth_getTransactionReceipt` falls back to the preconfirmation receipt index after canonical and Bor receipt lookup miss. Speculative receipts set `blockHash: null` and `preconfirmation: true`; clients must not treat them as canonical until `preconfirmation` is absent and `blockHash` is populated.

- `eth_sendRawTransaction` submits locally first, then forwards successfully accepted transactions to the configured preconfirmation relay when enabled. Relay failure is logged without rejecting the local submission.

- `eth_sendRawTransactionSync` returns the first available speculative or canonical receipt. It retains the normal chain-event path for canonical inclusion and, while preconfirmation is enabled, polls the receipt index every 100ms. The default timeout remains 20s and callers may override it within the configured maximum.

- `eth_feeHistory` preserves the continuous canonical portion of a request when the pending view is more than one height ahead of canonical head, rather than attempting to read unavailable intermediate heights.

Canonical transaction and receipt data takes precedence over speculative data. Invalidated preconfirmations are recorded through `bor_getInvalidPreconfBlocks`, and canonical mismatches are persisted atomically with the canonical head update.

`PendingStore` intentionally has no aggregate byte limit. Ordered transaction input is limited to 32 MiB per active block, the speculative transaction/receipt index retains at most 256 heights, and invalidation RPC responses are capped at 1024 records. Invalidation records are not pruned.

Sequence Store entries do not currently carry individual producer signatures. Before sealing, attribution relies on the authenticated publisher channel at Sequence Store ingress. Once `BlockSeal` arrives, Bor verifies the signed Bor header and producer authorization before publishing a sealed pending view or reusable execution.

## Executed tests

```text
make test TEST_TIMEOUT=60m
PASS
eth/sequencer coverage: 93.9%

make test-integration TEST_TIMEOUT=60m
PASS
tests/bor: 619.600s

go test -race ./eth/sequencer
PASS

go test ./eth/sequencer ./eth ./eth/filters ./internal/ethapi ./graphql ./rpc
PASS

make lint
PASS — 0 issues

git diff --check
PASS

Diffguard quality metrics
PASS
Cognitive complexity: 0 violations
Mutation thresholds: PASS
```

No Kurtosis, chaos, Amoy, mainnet, or upgrade-devnet run has been performed.

## Rollout notes

This change touches canonical block import but is intended to remain consensus-neutral. Cached results are accepted only when the incoming block, ordered transactions, execution context, transaction root, receipt root, bloom, gas usage, and state root match, followed by Bor’s normal `ValidateState` check. Otherwise, Bor executes the block through the existing canonical path.

The feature is disabled by default and requires explicit Sequence Store configuration. Nodes without it remain protocol-compatible. On feature-enabled nodes, clients consuming `eth_getTransactionReceipt` must check `preconfirmation: true` and `blockHash: null` before treating a receipt as final. A coordinated rollout is recommended before enabling cached import broadly because the import path is consensus-sensitive even though it does not change consensus rules.

Cached reuse is disabled for sprint-boundary and state-sync blocks, witness or stateless execution, tracing, preimage recording, extra EIPs, enforced parallel execution, and pipelined import. Those blocks continue through normal processing.

No database migration or resync is required. The only persistent addition is the `InvalidPreconf-*` rawdb keyspace. These records are intentionally retained without pruning, so operators enabling the consumer should account for gradual disk growth. Speculative blocks, receipts, state, and reusable executions remain memory-only and disappear on restart.


eth/sequencer/pending_types.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/ethereum/go-ethereum/ethdb"
+)
+
+const (
+	pendingTotalLimit     = 512 * 1024 * 1024
+	pendingInputLimit     = 32 * 1024 * 1024
+	pendingRPCLimit       = 64 * 1024 * 1024
+	pendingExecutionLimit = 256 * 1024 * 1024
+	pendingViewLimit      = 128 * 1024 * 1024
+	pendingMaxTxs         = 4096
```
cffls: Not sure if we should enforce this limit.

eth/sequencer/consumer.go
cffls: In this implementation, we will only be able to execute preconf txns whose parent block `N` has already been imported. In case where two consecutive blocks arrives to sequence store (`N+1` and `N+2`) before the nodes peers broadcast them, the node should be able to continue to execute `N+2`. Otherwise, the latency will be bounded by the p2p propagation again.

core/rawdb/accessors_preconf.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return records
+}
+
+func PruneInvalidPreconfs(db ethdb.KeyValueStore, newest uint64) error {
+	if newest < InvalidPreconfRetention {
+		return nil
+	}
+	return DeleteInvalidPreconf(db, newest-InvalidPreconfRetention)
+}
```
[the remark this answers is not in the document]
vbhattaccmu: resolved in c730d3868cc2

eth/api_backend.go
```diff
@@ -341,6 +357,9 @@ func (b *EthAPIBackend) HistoryPruningCutoff() uint64 {
 }
 
 func (b *EthAPIBackend) GetReceipts(ctx context.Context, hash common.Hash) (types.Receipts, error) {
+	if block, receipts, _ := b.sequencerPending(); block != nil && block.Hash() == hash {
+		return receipts, nil
+	}
 	return b.eth.blockchain.GetReceiptsByHash(hash), nil
 }
```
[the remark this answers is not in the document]
vbhattaccmu: resolved in c730d3868cc2

eth/api_backend.go
```diff
@@ -478,12 +500,23 @@ func (b *EthAPIBackend) GetCanonicalTransaction(txHash common.Hash) (bool, *type
...
 func (b *EthAPIBackend) TxIndexDone() bool {
 	return b.eth.blockchain.TxIndexDone()
 }
 
 func (b *EthAPIBackend) GetPoolNonce(ctx context.Context, addr common.Address) (uint64, error) {
+	if _, _, statedb := b.sequencerPending(); statedb != nil {
+		return statedb.GetNonce(addr), statedb.Error()
+	}
 	return b.eth.txPool.PoolNonce(addr), nil
 }
```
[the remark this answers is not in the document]
vbhattaccmu: resolved in c730d3868cc2

core/blockchain.go
```diff
@@ -3004,6 +3042,26 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
...
+	}
+	if err := rawdb.WriteInvalidPreconf(batch, block.NumberU64(), reason); err != nil {
+		log.Error("Failed to batch preconfirmation invalidation", "number", block.NumberU64(), "err", err)
+	}
+	if block.NumberU64() < rawdb.InvalidPreconfRetention {
+		return
+	}
+	if err := rawdb.DeleteInvalidPreconf(batch, block.NumberU64()-rawdb.InvalidPreconfRetention); err != nil {
+		log.Error("Failed to batch preconfirmation invalidation pruning", "number", block.NumberU64(), "err", err)
+	}
```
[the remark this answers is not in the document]
vbhattaccmu: resolved in c730d3868cc2

core/rawdb/accessors_preconf.go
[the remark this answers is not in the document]
vbhattaccmu: resolved in c730d3868cc2

eth/api_backend.go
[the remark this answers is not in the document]
vbhattaccmu: resolved in c730d3868cc2

eth/sequencer/pending_store.go
```diff
@@ -0,0 +1,410 @@
...
+	key := pendingKey{number: block.NumberU64(), parent: block.ParentHash()}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	entry := s.entries[key]
+	if entry == nil || entry.phase != PendingSealed || entry.Sealed == nil || entry.claimedHash != (common.Hash{}) {
+		preconfReuseMiss.Inc(1)
+		return nil, false
+	}
+	sealed := entry.Sealed
+	if sealed.HeaderHash != block.Hash() || sealed.TxRoot != block.TxHash() || !sameTransactions(entry.RPCView.Block, block) {
```
cffls, line 380: Should also check state root, gas used, and all other fields in block header I guess?

core/blockchain.go
```diff
@@ -1496,6 +1497,42 @@ func (bc *BlockChain) ProcessBlock(block *types.Block, parent *types.Header, wit
 	return result.receipts, result.logs, result.usedGas, result.statedb, result.vtime, result.err
 }
 
+func (bc *BlockChain) claimPreconf(block *types.Block, requireWitness bool, compatible bool) (*PreconfExecution, time.Duration) {
+	if bc.preconfProvider == nil || bc.logger != nil || !compatible {
+		return nil, 0
+	}
+	cached, ok := bc.preconfProvider.ClaimPreconf(block, requireWitness)
```
cffls: If I understand correctly, it looks like the cached result will only be used if the block has been fully executed in preconf. It means that if a block from peer arrives in the middle of preconf execution for the same block, even if the preconf already completed 99% of the transaction, we won't be able to reuse it again. Not sure how likely this is going to happen. If it happens frequently, we might need to allow blockchain.go to wait for preconf to complete (for this block) and use its result.

core/types.go
```diff
@@ -65,3 +65,15 @@ type ProcessResult struct {
...
+type PreconfExecution struct {
+	StateDB *state.StateDB
+	Result  *ProcessResult
+}
+
+type PreconfProvider interface {
+	ClaimPreconf(block *types.Block, requireWitness bool) (*PreconfExecution, bool)
+	RejectPreconf(block *types.Block)
+	PreconfInvalidation(block *types.Block, receipts types.Receipts) (string, bool)
+	CompletePreconf(block *types.Block, committed bool)
```
cffls: Would be good to add descriptions about what each method does.

cffls: What's the difference between RejectPreconf and PreconfInvalidation? They seem to do the same thing.

eth/sequencer/pending_store.go
```diff
@@ -0,0 +1,410 @@
+package sequencer
```
cffls, line 1: Would it be better to name this file `pending_store.go`?

eth/sequencer/consumer_flow.go
```diff
@@ -0,0 +1,122 @@
+package sequencer
```
cffls: Seems okay to merge this file to consumer.go?

cffls (commented):
Looks like in the right direction overall! A few questions and comments. One missing component is back track the sequence store if the node is just started, but that could be implemented later in a new PR.

eth/sequencer/pending_types.go
vbhattaccmu: Addressed in 49aa2cba5. Removed the transaction-count limit and the complete cache-budget mechanism, including size estimates, reservations, accounting, and memory-based eviction. Pending entries are retained until lifecycle removal. The remaining input-size guard only protects gRPC record ingestion and does not limit cache retention.

eth/sequencer/consumer.go
vbhattaccmu: Addressed in 49aa2cba5. N+2 now executes immediately from N+1’s speculative post-state when N+1 is the current speculative tip, without waiting for its canonical/P2P import. Added consecutive N+1/N+2 execution coverage.

eth/sequencer/consumer_flow.go
vbhattaccmu: Removed consumer_flow.go in 8c4974103. Consumer control flow now lives in consumer.go, while execution-specific helpers remain in exec.go to keep the files within the repository size checks.

eth/sequencer/pending_store.go
vbhattaccmu: Addressed in 49aa2cba5. HeaderHash is the complete block-header hash, so it commits every RLP-encoded field, including state root, gas used, bloom, and extra data. Added tests mutating state root, gas used, and extra data; each produces a cache miss.

core/blockchain.go
vbhattaccmu: Reviewed in 49aa2cba5. Canonical import remains nonblocking, and reuse occurs only after execution is fully sealed. Waiting for an in-flight preconfirmation could delay P2P import, so a bounded wait should be introduced separately if metrics demonstrate frequent overlap.

eth/sequencer/pending_store.go
vbhattaccmu: Addressed in 49aa2cba5. Renamed pending.go to pending_store.go.

eth/sequencer/consumer_flow.go
vbhattaccmu: Addressed in 49aa2cba5. Removed consumer_flow.go. Consumer control flow now lives in consumer.go, while execution-specific helpers remain in exec.go to keep the files within repository size checks.

core/types.go
vbhattaccmu: > Would be good to add descriptions about what each method does.

Addressed in commit 49aa2cba5. Added descriptions for every PreconfProvider method:
- ClaimPreconf reserves an exact sealed execution for canonical import.
- RejectClaimedPreconf discards a claimed execution when state validation fails.
- CheckPreconfInvalidation compares the pending RPC view with the normally executed canonical block and receipts.
- CompletePreconf removes a committed execution or releases an aborted claim.

> What's the difference between RejectPreconf and PreconfInvalidation? They seem to do the same thing.

`RejectClaimedPreconf` is called when a cached execution fails validation. It disables cache reuse and falls back to normal execution. `CheckPreconfInvalidation` runs after normal execution. It checks whether the preconfirmation differed from the canonical result and should be recorded as invalid.

eth/filters/api.go
```diff
@@ -494,6 +490,10 @@ func (api *FilterAPI) GetLogs(ctx context.Context, crit FilterCriteria) ([]*type
 
 	var borLogsFilter *BorBlockLogsFilter
 
+	if isPendingRange(crit) {
+		return api.getPendingLogs(ctx, crit)
+	}
```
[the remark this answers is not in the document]
vbhattaccmu: Addressed in 3f37f5928. GetLogs now routes crit.Pending through the pending-view path even when no explicit block range is supplied. getPendingLogs continues to reject incompatible block hashes or ranges. Added regression coverage for both { "pending": true } and explicit pending ranges.

eth/sequencer/exec.go
```diff
@@ -161,6 +173,87 @@ func (env *blockEnv) checkSeal(sealed *types.Header) error {
...
+		s.consumer.logsFeed.Send(logs)
+	}
+}
+
+func (s *session) sealResult(sealed *types.Header) (*types.Block, *ReusableExecution, bool) {
+	assembled, reusable, err := s.env.finalizeSeal(s.consumer.chain, sealed)
+	if err == nil {
+		return assembled, reusable, true
+	}
+	preconfSealFailure.Inc(1)
```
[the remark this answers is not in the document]
vbhattaccmu: Addressed in 8ad9ce0c7. An expected errCachedFinalizationUnavailable no longer increments preconfSealFailure when the fallback seal check succeeds. Unexpected finalization errors and genuine fallback seal mismatches still increment it.

eth/sequencer/consumer.go
```diff
@@ -118,6 +137,10 @@ func (c *Consumer) run(ctx context.Context) {
 		}
 
 		log.Warn("Sequence stream session ended", "err", err)
+		logs := c.pendingStore().invalidateFrom(0, "session_lost")
```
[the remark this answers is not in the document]
vbhattaccmu: Addressed in 8ad9ce0c7. Stream-loss invalidation now defers removal of entries in PendingImporting. If canonical import succeeds, the entry is removed without recording a false invalidation. If import aborts, the deferred invalidation is applied once and removed logs are emitted.

core/blockchain.go
cffls: Agreed that canonical import should remain nonblocking, but the point is that we should be allowed to reuse what has been already computed by the preconf. 

Consider a block with 100 txns. By the time it arrives at block import in blockchain.go, if preconf already executed 80 txns, at this moment we should just take the statedb from preconf and continue the block import at txn 81, instead of restarting from a fresh statedb. I would suggest testing this in kurtosis with large transaction volumes and see if the latency will become significantly higher in such scenario.

vbhattaccmu: Addressed in commit 5e9ff525b.

Canonical import can now reuse the latest completed preconfirmation transaction boundary without waiting for the full speculative block. In the 100-transaction example, it reuses the StateDB after transaction 80 and executes transactions 81–100. Reuse requires an exact transaction prefix and matching block context. Normal seal and state validation still run, with fallback to canonical execution on any mismatch. The exact 80-of-100 scenario, fallback behavior, and concurrency paths are covered by tests. High-volume Kurtosis testing is still pending.

eth/sequencer/pending_types.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+type PendingStore struct {
+	mu         sync.RWMutex
+	generation uint64
+	entries    map[pendingKey]*pendingEntry
+	active     pendingKey
+	hasActive  bool
+	db         ethdb.Database
+}
```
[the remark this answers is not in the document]
vbhattaccmu: Addressed in 96a1ca377

eth/sequencer/consumer.go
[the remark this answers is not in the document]
vbhattaccmu: Addressed in 96a1ca377

eth/sequencer/pending_rpc.go
```diff
@@ -0,0 +1,232 @@
...
+	view := entry.RPCView
+	s.mu.RUnlock()
+	block := view.Block
+	receipts := receiptsFromView(block, view)
+	stateDB, err := view.State.NewStateDB()
+	if err != nil {
+		return nil, nil, nil
+	}
+	return block, receipts, stateDB
+}
```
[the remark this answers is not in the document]
vbhattaccmu: addressed in 406cdd2ff

eth/filters/pending_logs.go
```diff
@@ -0,0 +1,85 @@
...
+	receipts, err := api.sys.backend.GetReceipts(ctx, header.Hash())
+	if err != nil {
+		return nil, err
+	}
+	var logs []*types.Log
+	for _, receipt := range receipts {
+		logs = append(logs, receipt.Logs...)
+	}
+	return returnLogs(filterLogs(logs, nil, nil, crit.Addresses, crit.Topics)), nil
+}
```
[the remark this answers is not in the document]
vbhattaccmu: Fixed in 9730d92b1. getPendingLogs now uses PendingBlockAndReceipts() to obtain the pending block and receipts from one coherent snapshot. The original header-plus-hash lookup remains only as a fallback for backends that do not support the snapshot accessor.

core/rawdb/accessors_preconf.go
```diff
@@ -0,0 +1,49 @@
...
+
+func WriteInvalidPreconf(db ethdb.KeyValueWriter, number uint64, reason string) error {
+	return db.Put(invalidPreconfKey(number), []byte(reason))
+}
+
+func ReadInvalidPreconfs(db ethdb.Iteratee, limit uint64) []InvalidPreconfRecord {
+	if limit == 0 || limit > InvalidPreconfQueryLimit {
+		limit = InvalidPreconfQueryLimit
+	}
+	iterator := db.NewIterator(invalidPreconfPrefix, nil)
```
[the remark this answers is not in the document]
vbhattaccmu: We should not add pruning because retaining all invalidated block records is the intended requirement.

core/rawdb/accessors_preconf.go
```diff
@@ -0,0 +1,104 @@
...
+func invalidPreconfKey(number uint64) []byte {
+	key := make([]byte, len(invalidPreconfPrefix)+8)
+	copy(key, invalidPreconfPrefix)
+	binary.BigEndian.PutUint64(key[len(invalidPreconfPrefix):], ^number)
+	return key
+}
+
+// PrepareInvalidPreconf adds an invalidation to batch and prunes records beyond
+// the newest InvalidPreconfQueryLimit heights.
+func PrepareInvalidPreconf(db ethdb.Iteratee, batch ethdb.KeyValueWriter, number uint64, reason string) error {
```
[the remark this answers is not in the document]
vbhattaccmu: Agreed. Fixed in 08fa679dd. PrepareInvalidPreconf no longer prunes or silently drops records, including older invalidations written out of order. All invalidation records are retained; ReadInvalidPreconfs only limits the number returned by the RPC query.

eth/gasprice/feehistory.go
```diff
@@ -217,6 +222,9 @@ func (oracle *Oracle) resolveBlockRange(ctx context.Context, reqEnd rpc.BlockNum
 		// Absolute number resolved.
 		reqEnd = rpc.BlockNumber(resolved.Number.Uint64())
 	}
+	if pendingBlock != nil && pendingBlock.NumberU64() > headBlock.Number.Uint64()+1 {
+		blocks = 1
+	}
```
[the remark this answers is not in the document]
vbhattaccmu: Agreed. Fixed in 08fa679dd. The range is now reduced to the pending block only when the entire requested range falls inside the speculative gap. If the request includes canonical blocks, the original range is preserved so available historical fee data is returned up to the first missing speculative block. Focused tests cover both cases.

internal/ethapi/preconf_receipt.go
```diff
@@ -0,0 +1,39 @@
...
+			continue
+		}
+		copyLog := *entry
+		copyLog.BlockHash = common.Hash{}
+		copyReceipt.Logs[i] = &copyLog
+	}
+	blockNumber := receipt.BlockNumber.Uint64()
+	result := marshalReceipt(&copyReceipt, common.Hash{}, blockNumber, types.LatestSigner(backend.ChainConfig()), tx, int(receipt.TransactionIndex), false)
+	result["blockHash"] = nil
+	return result
```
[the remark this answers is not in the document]
vbhattaccmu: fixed in 61c85c829

core/types.go
cffls: Thanks for providing the descriptions!

eth/sequencer/pending_helpers.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil
+	}
+	requests := make([][]byte, len(result.Requests))
+	for index := range result.Requests {
+		requests[index] = append([]byte(nil), result.Requests[index]...)
+	}
+	logs := make([]*types.Log, len(result.Logs))
+	for index, entry := range result.Logs {
+		copy := *entry
+		copy.Topics = append([]common.Hash(nil), entry.Topics...)
```
cffls: `cloneProcessResult` (here) and `cloneReceipt` (line 83) both do `append([]common.Hash(nil), entry.Topics...)`. For a log with zero topics (e.g. a `LOG0`), `entry.Topics` is empty, so appending to a `nil` base returns a `nil` slice, which marshals to `"topics": null`.

The canonical receipt for the same tx renders `"topics": []` (its log carries a non-nil empty slice), so the preconf and post-import receipts for one transaction disagree on this field:

```
bor_getPreconfTransactionReceipt        -> "topics": null
eth_getTransactionReceipt (post-import) -> "topics": []
```

Reproduced on a kurtosis devnet by calling a contract that emits a `LOG0`; every other receipt field matched exactly between the two. Low severity, but a strict JSON client that expects `topics` to always be an array can fail to decode the preconf form, and it's a visible shape change across the "upgraded transparently once imported" boundary.

Suggest seeding with a non-nil empty slice on both lines so the copy is stable for empty input, e.g. `append([]common.Hash{}, entry.Topics...)` (the `nil` -> `{}` change keeps the deep copy for the non-empty case). `Data` doesn't have this issue since empty bytes marshal to `"0x"`.

eth/sequencer/exec.go
```diff
@@ -161,6 +266,212 @@ func (env *blockEnv) checkSeal(sealed *types.Header) error {
...
+	case err := <-result:
+		return err
+	case <-time.After(preconfSealVerifyTimeout):
+		return errSealVerificationDeferred
+	}
+}
+
+func (s *session) sealResult(sealed *types.Header) (*types.Block, *ReusableExecution, bool, bool) {
+	headers := &speculativeHeaderChain{ChainHeaderReader: s.consumer.chain, headers: s.verified}
+	if err := s.consumer.verifyPreconfSeal(headers, sealed); err != nil {
```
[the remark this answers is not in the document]
vbhattaccmu: fixed in 1fd61e89b

eth/filters/pending_logs.go
```diff
@@ -0,0 +1,144 @@
...
+	return crit.FromBlock != nil && crit.FromBlock.Int64() == rpc.PendingBlockNumber.Int64() ||
+		crit.ToBlock != nil && crit.ToBlock.Int64() == rpc.PendingBlockNumber.Int64()
+}
+
+type pendingLogsBackend interface {
+	PendingBlockAndReceipts() (*types.Block, types.Receipts)
+}
+
+func (api *FilterAPI) getPendingLogs(ctx context.Context, crit FilterCriteria) ([]*types.Log, error) {
+	if !onlyPendingRange(crit) {
```
cffls, line 81: Flagging an `eth_getLogs` interface mismatch vs the pending-logs usage we discussed for indexer integrations. The expected pattern is the existing log query with just `toBlock` changed to `pending` (fromBlock stays a concrete block number).

That query is rejected here. `getPendingLogs` requires `onlyPendingRange(crit)`, which returns `false` whenever `fromBlock` is a concrete number, so `{fromBlock: <n>, toBlock: "pending"}` fails with `errInvalidBlockRange`. `{toBlock: "pending"}` with no `fromBlock` also fails (`onlyPendingRange` requires `crit.Pending && ToBlock==nil`). Only `{pending: true, ...}` and `{fromBlock: "pending"}` are accepted — and they return just the *current pending block's* logs, not a range up through pending.

So the drop-in ("keep your query, just change toBlock to pending") doesn't hold; the real integration is a two-query merge: `{pending: true}` for the pending block **plus** the normal canonical `[last_seen, latest]` polling. That's a reasonable design and matches the opt-in `pending` flag in the RPC-changes doc.

eth/sequencer/pending_helpers.go
vbhattaccmu: Fixed in 7bfc0707d. Both cloneProcessResult and cloneReceipt now copy topics using a non-nil empty slice, so a LOG0 receipt consistently returns "topics":[] before and after canonical import. A regression test covers the JSON representation.

eth/filters/pending_logs.go
vbhattaccmu: Fixed in 7bfc0707d. eth_getLogs now accepts both {fromBlock:<n>,toBlock:"pending"} and {toBlock:"pending"}. It combines canonical logs through the current head with the available speculative pending range, while retaining {pending:true} for pending-only queries. It also validates snapshot continuity and retries if the canonical anchor changes during the request. This affects only the RPC query path, not transaction execution or pending-view publication.

PR #2376, opened 2026-08-27
core, txpool, miner, eth/gasprice, internal/ethapi: reserved blockspace follow-ups

kamuikatsurgi:
## Summary

Follow-up hardening on top of the core Reserved Blockspace protocol change (#2302): value-only-balance txpool admission and fallback-fee handling (POS-3671), an aggregate reserved-sender occupancy cap to prevent normal-sender starvation (POS-3681), gas-price-oracle/`eth_feeHistory` exclusion of fee-free reserved transactions plus a new `normalGasUsedRatio` field (POS-3675/3676), `chain/reserved/*` and `worker/reserved/*` observability metrics (POS-3679), and execution-path produce/import determinism coverage at the fork boundary across the serial and both BlockSTM processors (POS-3672). Also carries the wire-format reconciliation needed when the Austin hard fork changed the header extra-data shape mid-implementation, and two correctness fixes found via live devnet testing (see below). Also fixes a txpool-only correctness/security gap found during a same-day security review of the occupancy-cap change: a same-nonce transaction replacement could grow to `txMaxSize` while skipping the reserved-occupancy cap check entirely, letting a reserved sender occupy several times their intended pool share (`ad75b96ad`).

A combined delivery-and-test-status primer for both this branch and the base protocol branch, written for the still-open production go/no-go decision:
**[Reserved Blockspace - Delivery & Test Status](https://primer-polygon.up.railway.app/d/01a04295-372c-745f-a37e-9d4fc3d22a2f)**

## Executed tests

- Full unit + `-race` suites across every touched package, and all 9 `tests/bor` reserved integration tests green.
- Two full live devnet trials on a 9-node Kurtosis topology covering every node kind (producers, witness producer, stateless validators/RPC with and without BlockSTM parallel import, archive) - full write-up: [Reserved Blockspace Devnet Trial](https://claude.ai/code/artifact/3d4c5c48-ef5f-4fbe-a573-7af04a292305).
  - Round 1 reproduced the spec's core behaviors live, then found a critical issue while exercising registry governance for the first time on a live chain: growing the registry could permanently wedge stateless nodes, and with stateless validators in the mix this cascaded into a full network production halt via lost Heimdall span-rotation quorum (`registryreader.BuildSnapshot` read the registry against a state copy whose witness never reached the real per-block witness - `cf9f3e82c`). Also found `feeMode=1` clients were mining fee-free identically to `feeMode=0`, contradicting the contract's own documented semantics (`124087f55`).
  - Round 2 rebuilt from both fixes and re-ran the identical, riskier topology (stateless validators included) through the same trigger, a rapid-mutation stress pass, and two full node cold-restarts, with zero divergence - plus a deeper pass on the registry contract's own access-control and boundary semantics (client-admin delegation, exact quota-cap boundaries, invalid-input reverts, a 21-client/67.5M-gas registry with no measurable timing impact).
- Not yet covered: a naturally-occurring reorg around a reserved-tx block, and a standing automated e2e/upgrade CI harness (both scripted-but-manual so far, not checked into CI).

## Rollout notes

Consensus-affecting (gated behind `Bor.ReservedBlockspaceBlock`, same as the base branch). No operator-facing config changes beyond the existing `ReservedMaxOccupancyPercent` txpool knob (default 50%, CLI-wired). Both devnet-found issues above are fixed on this branch; production readiness still depends on the smart-contract team's audited registry contract and genesis-contracts parity, neither of which is part of this PR - see the primer's "still outstanding" section.

**Node upgrade ordering:** this PR adds a new header field (`ReservedCapacity`, alongside the existing `ReservedGasUsed`) to `BlockExtraData`/`BlockExtraDataPostAustin`. Like any hard-fork field addition, every node needs the upgraded binary before any of them start producing past the fork block - a pre-upgrade node attempting to parse a post-fork header carrying this field will fail. Nothing is live on any network yet, so this is purely a reminder for whoever runs the next devnet round or upgrade rehearsal: roll out the binary to all nodes first, in line with the existing `Bor.ReservedBlockspaceBlock` hard-fork gate.


core/txpool/legacypool/legacypool.go
```diff
@@ -2360,6 +2542,21 @@ func (pool *LegacyPool) isReserved(addr common.Address) bool {
...
+// pool is quota-unaware, so this is a superset of what will actually execute
+// fee-free; an overflowing tx that doesn't fit is caught at execution instead
+// (see the design note on isReserved). addr is the caller's already-recovered
+// sender; unlike reservedTx (the priced heap's bare-tx predicate), this is
+// never called without one, so it has no signer-recovery fallback.
+func (pool *LegacyPool) effectiveCost(addr common.Address, tx *types.Transaction) *big.Int {
+	if pool.isReserved(addr) {
+		return tx.Value()
+	}
+	return tx.Cost()
```
marcello33, line 2617: q on this one: `effectiveCost` gives value only pricing just because the sender is a registered reserved client, without checking if this specific tx's gas actually fits their quota. 
I understand that the comment above says this is intentional, but doesn't that mean a reserved sender can get fee-free pricing AND the eviction protection on a tx that was never going to execute fee-free anyway? Feels like it opens the door to spamming the pool pretty cheaply. Might be worth a quota check at admission time too, wdyt?

core/txpool/legacypool/legacypool.go
```diff
@@ -1008,12 +1047,33 @@ func (pool *LegacyPool) add(tx *types.Transaction, async bool) (replaced bool, e
...
+	// transaction would occupy a genuinely new slot (as opposed to a same-
+	// nonce replacement, which leaves combined occupancy unchanged) that
+	// pushes aggregate reserved occupancy over its cap. Runs unconditionally,
+	// independent of overall pool fullness — unlike the Underpriced exemption
+	// below, which only applies once the pool is already globally full.
+	if reserved && pool.isNewReservedSlot(pendingList, from, tx) {
+		if pool.reservedOccupancy+1 > pool.reservedOccupancyCap() {
+			stage0Duration = time.Since(stage0Time)
+			return false, ErrReservedOccupancyExceeded
+		}
```
marcello33, line 1089: The `reservedOccupancyCap()` is computed against `GlobalSlots`+`GlobalQueue`, which is the same denominator the pool uses for its real slot based fullness check (with `numSlots(tx)`). However, every place that bumps `reservedOccupancy` just does +1/-1 per tx, never `numSlots(tx)`. So a reserved sender sending big calldata txs (multiple slots each) would only count as "1" against a cap that's measured in slots? Keep me honest here, but could someone blow past the intended 50% pool share pretty easily, right? Maybe worth a check

consensus/bor/registryreader/reader.go
```diff
@@ -113,21 +145,22 @@ func BuildSnapshot(r Reader, statedb *state.StateDB, number uint64, hash common.
...
+// in free fee mode (see FeeModeFree for why non-free modes are excluded).
 func resolveClients(r Reader, statedb *state.StateDB, number uint64, hash common.Hash, addrs []common.Address, effectiveAt uint64) (map[common.Address]Client, error) {
 	clients := make(map[common.Address]Client, len(addrs))
 	for _, a := range addrs {
 		c, err := r.ReservedClientForAddress(statedb, number, hash, a)
 		if err != nil {
 			return nil, err
 		}
-		if !c.Active || c.EffectiveFrom > effectiveAt {
+		if !c.Active || c.EffectiveFrom > effectiveAt || c.FeeMode != FeeModeFree {
```
marcello33, line 157: looks like non-free (`feeMode!=0`) clients still get read from the registry every single block before getting filtered out. Not a big deal today since it's gated behind whoever controls the whitelist, but if that list grows with a bunch of non-free addresses, seems like unnecessary per-block work for addresses that can never be reserved anyway. Maybe we should filter earlier in the read path? Not sure how much this matters in practice tho

eth/gasprice/gasprice.go
```diff
@@ -320,3 +346,39 @@ func (oracle *Oracle) getBlockValues(ctx context.Context, blockNum uint64, limit
...
+func (oracle *Oracle) reservedTxHashes(ctx context.Context, block *types.Block) map[common.Hash]struct{} {
+	receipts, err := oracle.backend.GetReceipts(ctx, block.Hash())
+	if err != nil || len(receipts) != len(block.Transactions()) {
+		log.Debug("Gasprice oracle degraded to no reserved-tx exclusion", "block", block.NumberU64(), "err", err)
+		return nil
+	}
+
+	var reserved map[common.Hash]struct{}
+	for _, receipt := range receipts {
+		if isReservedReceipt(receipt) {
```
marcello33, line 367: Here and in `feehistory`, I think this might hit on the state-sync txs, because it's inferring "reserved" from `EffectiveGasPrice == 0`, but the state-sync tx also reports 0. Could we end up skewing `eth_feeHistory` a bit on blocks that have both? Seems like `internal/ethapi/api.go` uses the actual index table instead, which seems safer. Shall we reuse that here too?

core/types/block.go
```diff
@@ -599,6 +617,45 @@ func (h *Header) GetValidatorBytes(chainConfig *params.ChainConfig) []byte {
...
+	}
+
+	if len(h.Extra) < ExtraVanityLength+ExtraSealLength {
+		return nil, nil, false
+	}
+
+	payload := h.Extra[ExtraVanityLength : len(h.Extra)-ExtraSealLength]
+
+	if chainConfig.Bor != nil && chainConfig.Bor.IsAustin(h.Number) {
+		var blockExtraData BlockExtraDataPostAustin
```
marcello33, line 639: If we add the `ReservedCapacity` as a new field, does this mean an old node can't parse a header once that field shows up? Probably fine since nothing's live yet and we have a HF, but might be worth an addition in the PR description / docs reminding whoever runs the next devnet round that every node needs the new binary before any of them start producing.

core/blockchain.go
```diff
@@ -2612,9 +2627,12 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
...
 		}
 
-		// Write all chain data to ancients.
+		// Write all chain data to ancients. These blocks arrive via receipt
+		// sync rather than local execution, so there is no reserved-tx
+		// classification to carry; the nil entries make writeAncientBlock
+		// record that explicitly.
 		td := bc.GetTd(first.Hash(), first.NumberU64())
-		writeSize, err := rawdb.WriteAncientBlocks(bc.db, blockChain, receiptChain, borReceipts, td)
+		writeSize, err := rawdb.WriteAncientBlocks(bc.db, blockChain, receiptChain, borReceipts, make([]rlp.RawValue, len(blockChain)), td)
```
marcello33, line 2635: Do blocks that come in via receipt sync get `reservedTxIndexes` hardcoded to `nil`? I mean it would make sense, as no local execution is there to derive it from, but doesn't that mean `EffectiveGasPrice` for reserved txs on those blocks is just permanently wrong, with no way to backfill? Might be a non-issue if we never actually run that sync mode across the reserved-blockspace range, but wanted to double check.

core/txpool/legacypool/legacypool.go
kamuikatsurgi: yeah this is intentional, matches the spec directly - §8.1 waives the balance check unconditionally for registered senders, no quota check. AccountSlots/AccountQueue plus the new occupancy cap already bound this even without an admission-time quota check - the spec's own §8.2 calls out a per-client cap as the natural follow-up if spam shows up in practice, which is basically what the occupancy cap already is. don't think we need this right now.

core/txpool/legacypool/legacypool.go
kamuikatsurgi: good catch, real bug - fixed. was bumping by flat 1 per tx while the cap's computed in slots. also found a nastier version of the same bug while fixing it: same-nonce replacements skipped the cap check entirely (assumed occupancy-neutral, which held under flat counting but not once tx size can differ), so a sender could replace cheap txs with maximal-size ones at trivial cost and blow past the cap 4x, repeatably. fixed both, added tests confirming they fail on the old code and pass now.

consensus/bor/registryreader/reader.go
kamuikatsurgi: don't think this is fixable in a meaningful way - you need the read to know the fee mode in the first place, so there's no way to filter earlier than right after it comes back. cost's the same for every whitelisted address regardless of mode, a non-free client doesn't add anything extra, it just doesn't make it into the map.

eth/gasprice/gasprice.go
kamuikatsurgi: you're right, state-sync's effectiveGasPrice is hardcoded to 0 so it does match. checked the actual impact though and it's a no-op today - state-sync's gasUsed is 0 (no weight in feehistory either way) and its tip is 0 (already below the ignoreUnder floor in the sampler regardless). real imprecision, zero practical effect right now. fixing it properly means widening OracleBackend's interface for a currently-inert bug, so I'd rather leave it documented and pick it up if it ever actually matters.

core/types/block.go
kamuikatsurgi: good point, added a note to the PR description about upgrade ordering. no code change needed here, this is just normal hard-fork rollout behavior.

core/blockchain.go
kamuikatsurgi: confirmed, real gap but scoped narrowly - only fallback-fee-in-quota reserved txs are affected (zero-fee ones come out correct by accident since their own fee fields are already 0). also inherent to how receipt-sync works today, the reserved-tx table is bor-local and never gossiped so there's nothing to backfill from without a wire-protocol change. think this is best left as a documented limitation for now rather than fixed here.

PR #2398, opened 2026-09-09
eth/downloader/whitelist: bound fork-check slice preallocation

kriss39:
## Summary

Bound the initial capacity of `blocksChecked` in `checkForkCorrectness` to the same 2x-default size already used for the fork validation cache.

`max-blind-fork-validation-limit` remains the actual traversal limit. This change only avoids eagerly preallocating a slice directly from unusually large configured values; the slice can still grow normally when more entries are actually visited.

A regression test covers a very large configured limit and verifies that fork correctness checking continues normally while preserving the configured traversal limit.

## Executed tests

A focused Go runtime reproducer was used to confirm the oversized-capacity behavior. The full Bor test suite was not run locally because the repository could not be cloned from this environment; CI is expected to run the repository's standard checks.

## Rollout notes

Backwards-compatible, non-consensus behavior change. Default configuration behavior is unchanged. No coordinated upgrade is required; this only bounds an internal initial allocation.

### DIFF
--- a/eth/downloader/whitelist/service.go
+++ b/eth/downloader/whitelist/service.go
@@ -29,6 +29,10 @@ var (
 	DefaultMaxForkCorrectnessLimit = uint64(256)
 )
 
+func forkValidationInitialCapacity(limit uint64) uint64 {
+	return min(limit, 2*DefaultMaxForkCorrectnessLimit)
+}
+
 type Service struct {
 	db ethdb.Database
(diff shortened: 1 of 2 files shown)


pratikspatil024 (approved):
LGTM, thanks!

vbhattaccmu (approved):

marcello33 (approved):

PR #2401, opened 2026-09-10
core, eth: detect and self-heal missing contract code on stateless verification

lucca30:
## Problem

A WIT2 stateless node reads contract code from its **local disk** — witnesses
carry no code. When a called contract's bytecode is absent (a bytecode-heal /
fast-forward completeness gap), `StateDB` records a sticky error and serves the
read as phantom-zero, so execution silently continues against wrong state. The
miss then surfaces only as a misleading `ErrGasUsedMismatch` — or, when the
divergent execution hits "gas limit reached", as that `Process` error —
masquerading as consensus divergence.

Observed on mainnet-soak: block `93472327` diverged (gasUsed `46,716,714` vs
`47,007,392`) because code `0xf98d41f7…e56e64` was missing from local disk, and
the true cause was invisible in the logs.

## Fix

This PR both **detects** the condition and **recovers** from it.

**Detect** (`core`, `core/state`) — check `db.Error()` in `ExecuteStateless`
immediately after `Process`, **before** the `Process`-error branch (a failed
state read invalidates the result, so it is the true cause regardless of which
downstream symptom fires first). Reject with a typed
`ErrStatelessIncompleteState` wrapping a `state.MissingCodeError{Addr, Hash}`
(the "code is not found" message is unchanged, so witness-regen fixtures still
match), and bump a `chain/stateless/incomplete_state` meter. The typed error
lets a caller distinguish a missing **code** (fetchable, content-addressed) from
a missing **trie node** (an incomplete witness, not fetchable).

**Recover** (`eth/protocols/snap`, `eth/downloader`) — on a
`*state.MissingCodeError` from a stateless batch import, fetch that
content-addressed blob from a snap peer via `Syncer.FetchByteCodes` (verify each
blob against its hash, fail over across peers, usable outside a running `Sync`
cycle), persist it, and retry the batch. Bounded by `maxStatelessCodeHeals` so a
blob no peer can serve degrades to the existing safe failure instead of looping.

Together this is **mechanism-agnostic**: it recovers any missing code at
execution time regardless of how the gap arose (heal incompleteness, crash
window, or fast-forward), turning the "code arrived later" recovery from
eventual into immediate.

## Test

- `TestExecuteStatelessRejectsMissingCode` drives the real `ExecuteStateless`
  over a mainnet witness fixture, removes each contract code the block reads, and
  asserts `ErrStatelessIncompleteState` naming the hash (and that the hash
  round-trips via `errors.As`) — never a gas mismatch, a `Process` symptom, or a
  silent pass. Writing it caught a real ordering bug: a missing code first
  surfaced as "gas limit reached" from `Process`, which is why the `db.Error()`
  check must precede the `Process`-error branch.
- `eth/protocols/snap` — `TestFetchByteCodesOnDemand` (well-behaving peer;
  unknown hash never fabricated) and `TestFetchByteCodesFailsOverAndVerifies`
  (fails over past a peer returning bytes that don't match the requested hash).
- `eth/downloader` — `TestRecoverMissingStatelessCode` (a servable miss is
  fetched, verified and persisted; a non-code error and an unservable blob are
  both left to the safe failure path, writing nothing).

## Scope / risk

Stateless-verification path only. A valid block whose code is fully present is
unaffected: `db.Error()` is nil and the existing `ValidateState` path runs
unchanged. Every fetched blob is verified content-addressed
(`keccak(code) == hash`) before it is persisted, so a malicious peer cannot
inject wrong code — it simply fails over. Recovery is bounded and safe-stops on
an unservable code.

> Consolidated from the previously stacked #2401 (detect) + #2402 (recover) into
> a single squashed commit for one review/merge unit.


cffls (approved):
LGTM, thanks!

pratikspatil024 (approved):

PR #2404, opened 2026-09-13
core: delete hash to number mapping during SetHead chain rewind

### LINKED ISSUE #2386: core: SetHead to a height behind the divergence does not clear a wedged sidechain state
## Summary

Issue #2224 reported two problems. PR #2293 fixes the first one, the false-positive `sidechain ghost-state attack` rejection on empty sibling blocks. This issue tracks the second one, which #2293 does not address: a node that already wedged does not recover from a short `debug_setHead`.

## What the trace shows

From the #2224 operator session on node v2, `debug_setHead 0x59f` (block 1439) rewound the chain as expected (excerpt, intermediate lines trimmed):

```
WARN  Rewinding blockchain to block            target=1439
INFO  Truncating from head                     type=state ohead=1442 tail=427 nhead=1441
INFO  Truncating from head                     type=state ohead=1441 tail=427 nhead=1440
INFO  Rewound to block with state              number=1439 hash=280067..874280
INFO  Loaded most recent local block           number=1439 hash=280067..874280 td=2316 age=47m16s
```

The warn-and-drop loop continued after the rewind. The operator had to rewind well past the divergence before the node resynced.

## Why the cause is still open

#2224 records a best guess: the side-block record for the old 1441 hash survives `SetHead` because "only the canonical pointer was rewound". Reading the rewind path, that explanation does not hold on its own.

`HeaderChain.setHead` gathers every hash at each height with `ReadAllHashes`, then deletes each header and its TD, not only the canonical entry:

https://github.com/0xPolygon/bor/blob/c788d1a97cb389312d56daeda50b659fa962b226/core/headerchain.go#L705-L726

On the first iteration it also sweeps every height above the target:

https://github.com/0xPolygon/bor/blob/c788d1a97cb389312d56daeda50b659fa962b226/core/headerchain.go#L690-L699

The trace confirms the loop ran rather than exiting early: `ohead` moves 1442, then 1441, so the body executed for both heights above the target. The trace also shows `tail=427`, so blocks 1439 to 1442 sat in the active key-value store and not the chain freezer, which rules out the documented freezer no-op:

https://github.com/0xPolygon/bor/blob/c788d1a97cb389312d56daeda50b659fa962b226/core/headerchain.go#L713-L718

So the header and TD for the diverged 1441 should have been deleted. Something else reintroduces or re-reaches that block on the next sync attempt, and the recorded guess does not identify it.

## What would help

Determine what actually re-triggers the check after a short rewind. Candidates to rule in or out, none of them yet confirmed:

- Peer-supplied headers arriving again and re-entering `insertSideChain` before the node reconciles with the canonical chain.
- Downloader or fetcher state that outlives the rewind within the same process.
- Data the rewind leaves behind that is not the header or TD.

A reproduction against current `develop` would settle it. #2224 has the scenario and the captured artifacts.

## Expected behavior

A `debug_setHead` to a height behind the divergence recovers the node. If a deeper rewind is genuinely required, the operator gets a clear reason why. 

## Notes

Reproduction steps, the full timeline, and the artifacts are in #2224. The fix in #2293 stops a node from reaching this state, so this issue matters for nodes that wedged before that fix ships, and for any other path that leaves a diverged side block behind the head.

Tyagiquamar:
Fixes #2386.

### Root Cause Analysis

When a node wedged on a sidechain divergence (e.g. at height 1441) calls \debug_setHead\ to rewind to a height behind the divergence (e.g. height 1439):
1. \HeaderChain.setHead\ rewinds the block headers and deletes header/difficulty mappings for displaced heights.
2. However, persisted milestone keys (\LastMilestone\, \LockField\, \FutureMilestoneField\) and in-memory downloader milestone states higher than the rewind target were left intact in the database.
3. Upon sync resumption, incoming canonical headers were validated against the stale milestone locks from the displaced sidechain fork.
4. The check failed with a milestone mismatch (\checkForkCorrectness\ returning \alse\), causing the node to continuously reject valid peer chains and fall back into a warn-and-drop loop.

### Changes Included

1. **Database Purge Helper**: Added \
awdb.PurgeStaleMilestonesFromDb(db, head)\ in \core/rawdb/milestone.go\ to delete \LastMilestone\, \LockField\, and filter \FutureMilestoneField\ entries strictly higher than \head\.
2. **Chain Rewind Purge**: Integrated \
awdb.PurgeStaleMilestonesFromDb\ inside \HeaderChain.setHead\ in \core/headerchain.go\ so all chain rewinds clean persisted milestone locks.
3. **Downloader State Reset**: Ensured \EthAPIBackend.SetHead\ in \th/api_backend.go\ calls \.eth.handler.downloader.PurgeMilestonesAfter(number)\ before \SetHead\ to reset in-memory downloader milestone locks and fork validation caches.
4. **Unit Test**: Added \TestPurgeStaleMilestonesFromDb\ in \core/rawdb/accessors_chain_test.go\ verifying that rewinding past milestone boundaries cleans persisted locks and future milestone lists.


Tyagiquamar:
@pratikspatil024 Cleaned up the branch and rebased onto the latest develop. The PR now only contains the intended 4 files for purging stale milestones and downloader locks on SetHead rewind (plus unit tests). Verified locally in Docker with \go test -v ./core/rawdb/ -run TestPurgeStaleMilestonesFromDb\ passing.

pratikspatil024:
diffguard in the CI is failing, could you please fix it

Tyagiquamar:
@pratikspatil024 Addressed the diffguard failure: expanded unit test coverage with dedicated subtests for full milestone list deletion and non-stale entry retention, and cleaned up unused conditional branches in SetHead. All mutation checks now pass.

pratikspatil024:
@claude review

PR #2406, opened 2026-09-15
sequencer: stop paying for a store that is not answering

pratikspatil024:
## Summary

Addresses the second finding from John's chaos campaign ([report](https://primer-polygon.up.railway.app/d/01a08de0-dce9-76ca-b476-f84d2d069743), F2 and F3): **the sequence store sits on the block-production critical path**, against a design brief that says "Nothing in this service can affect block production". The producer waited on the store on every block even where the answer it waited for could not arrive.

| fault | campaign | reproduced here | with this PR |
| --- | --- | --- | --- |
| gateway frozen (read path, F3) | 0.47 block/s | **0.48** | **0.98**, seal 2 ms |
| ingress frozen (write path, F2) | 0.74–0.78 block/s | **0.75**, seal 503 ms | **0.95–0.97**, seal 53–192 ms |
| no fault | 1.00 block/s | 1.00, seal 4 ms | 1.00, seal 4 ms |

**One flag caused both findings.** `Publisher.unreachable` meant "do not wait on the store" and every hot-path wait already checked it — but it had a single setter (publish-transport failure) and three clearers, one of them wired to *every successful read* (`markReachable` is the reader's `onRead`). So it was false during an ingress outage, because the healthy gateway kept clearing it, and false during a gateway outage, because nothing but the publish transport ever set it. Each outage went on paying the other path's waits.

It is now two signals:

- **`writeDown`** — the publish path. Set on transport failure, and cleared **only by an arriving ack**, which is the only evidence an ack can arrive. A stream session ending is not evidence: while an outage holds a build, the held-build watchdog ends one every 400 ms from a local timer, and clearing on that left the flag false for all but a handful of blocks (measured: the fix engaged on 7 of 46 blocks before this, 54 of 58 after).
- **`readBreaker`** — the read path, on the reader, so both client call sites funnel through it. Two consecutive unanswered reads open it; reads are then refused outright, with one probe per interval looking for the path again. Callers need no new handling: each already treats an unreadable tail as its safe branch.

**A note on the report's naming.** It attributes the write-path cost to "a 500 ms `ConfirmSeal` barrier". Those are two separate things here: `sequenceBarrierTimeout` (the pre-seal barrier) is **120 ms**, and the 500 ms is `sealGateTimeout`, the post-seal gate. The devnet agrees with the code — seal elapsed under ingress freeze was 503 ms, matching the report's own "503–521 ms". The measurements are sound; the fix just lands somewhere other than where the name points.

**Giving up on the ack still resolves what the wait would have.** `ConfirmSeal` takes one pass of the store's own record and then the chain — a rival's block at our height is the rejection notice and needs no store read — rather than returning `SealUnknown` blind. `storedVerdict` runs first and not the chain alone, because an ungated build carries height 0, and asking the chain about height 0 compares genesis against an empty hash and refuses a seal nobody contested.

**Probe budget.** A probe asks one question — does the path answer — so it gets 150 ms rather than the full read budget. Probing once a second at a full budget was, by itself, the whole of the throughput still missing: 12 probes × 1.25 s across a 60 s outage is a 25 % loss against a 27 % measured. A path too slow to answer in 150 ms keeps the breaker open, which is the right answer anyway.

**Six new metrics** under `sequencer/`, one per declined wait (`publish/barrierskipped`, `gate/writedownskip`, `gate/readdownskip`, `read/skipped`, `read/breakeropened`, `read/breakerclosed`). The campaign's complaint was not only the cost but that `sequencer_publish_barriertimeout` incremented "on nearly every block, with no WARN anywhere" — a degraded store is now visible.

## Executed tests

- `eth/sequencer` and `miner` green; `-race` clean on `eth/sequencer`. `go vet`, `gofmt`, `gofumpt` and `golangci-lint` (v2.11.4, repo config) clean.
- **Diffguard passes** with CI's flags: tier-1 logic **100 %** (gate 80), complexity, sizes, dependency structure and dead code clean. The three survivors are a defensive nil guard and two log/metric statements (tier-3, ungated).
- **Every new test was confirmed to fail without its fix**, so none passes vacuously:
  - `TestFrozenReadPathIsNotPaidForEveryBlock` — 8 blocks spend **10.02 s** against a stalled read path with the breaker disabled, **1.26 s** with it.
  - `TestSuccessfulReadDoesNotClearTheWriteFlag` — fails the moment a read is allowed to clear `writeDown` again.
  - `TestWriteDownGiveUpDoesNotRefuseAnUngatedSeal` — returns `SealRefused` for an uncontested seal if `storedVerdict`'s height-0 short-circuit is removed.
  - `TestReadBreakerOpensOnSilentGenerationFetches` — fails if the per-height fetch stops feeding the breaker.
- The stalled-store stub **completes the gRPC handshake and then never answers**, because a listener that merely accepts TCP does not reproduce this: gRPC fails an RPC immediately on a channel that never reached READY, so the call returns `Unavailable` in microseconds instead of hanging. My first version of that harness did exactly that and the test passed against the unfixed code. The campaign saw the same distinction from the other side — a killed envoy (fail fast) cost 0.71 block/s where frozen gateways (hang) cost 0.47.
- **Reproduced end to end on a kurtosis devnet**, 4 validators + 1 RPC at 1 s blocks with the store in-enclave, `docker pause` on each service in turn for 60 s, run once per image with only `el_image` differing. The unfixed run landing within 0.01–0.03 block/s of the campaign's own numbers is the main evidence the rig is faithful.

**The devnet caught two defects the unit tests could not**, both because those tests set the flag by hand: the 400 ms held-build watchdog clearing `writeDown`, and the probe interval costing a full read budget every second. Worth knowing for anything else in this area — a flag-level test proves the short-circuit fires, not that the flag is ever in the state that fires it.

## Rollout notes

Not consensus-affecting, no coordinated upgrade. Confined to the sequence-store publisher, which is off by default. No new operator knobs, no database or wire changes. Six new metrics.

**No safety property changes.** Every site touched already returned "proceed" once its budget expired; this changes when the node learns nothing, not what it does about it. The gate keeps the chain check it always had, and gains one it previously skipped on this path.

**F3 is closed: 0.48 → 0.98 block/s, sealing at baseline latency.** The breaker alone took it to 0.78, and the residual was the seal gate spending its 500 ms: during a read-path outage the publisher cannot anchor — `reconcile` → `readTail` is a read — so no ack arrives, while `writeDown` is legitimately false because the ingress is healthy.

The gate now gives up in that state too. The signal is the read breaker, not staleness: only *silent* reads trip it, so a contended store that still answers — `NOT_FOUND` included — leaves the gate fully armed, and an open breaker means the store cannot inform us rather than that it says we lost. (An earlier draft of these notes proposed skipping when unanchored and talked itself out of it, correctly: `endStale` unanchors and a STALE means contention, which is exactly when the gate earns its keep. The breaker is the signal that draft was looking for; thanks to Jerry for pushing on it.)

The justification is baseline-equivalence, and it is worth stating plainly. The gate is an equivocation self-check layered on top of consensus, not a safety floor — production runs today with no sequence store at all, where two producers at one height fork and consensus resolves it. When the store is down its coordination is gone regardless of what the gate does, so sealing immediately degrades to exactly that accepted behaviour, and buys back the whole slot. Measured: 21 of 47 blocks in the earlier F3 window paid the full budget and settled `SealUnknown` — the wait was resolving nothing.

**One exception keeps its wait.** A gate with `refuseOnTimeout` set is the case where the refusal does not depend on an answer still owed: `sealWithoutPublishLocked` sets it from a read that already found the height closed in the store with content this block does not carry. Skipping there would broadcast over an acked generation. That refusal needs no store round trip, so keeping it is free.

**And an ack already delivered still wins,** because `storedVerdict` runs before anything else. What the skip gives up is an ack *in flight* at the instant we stop waiting — the sliver the baseline argument covers. It cannot be avoided by holding the build instead: `holdNewLocked` gates the entries, but the hold clears at seal time, so a held build does flush and can be acked.

**The 4 s seal outlier from the earlier run is explained and gone.** It was `contestedGateTimeout`: a refused read fails instantly where it used to time out, so the reconcile loop retried faster and contested more gates. Skipping the gate when the read path is down removes the wait it was waiting in — the read-path window's seal max is now 11 ms against 4002 ms before.


cffls:
**On the F3 residual (0.48 → 0.78): I think we can close it, not just improve it — by skipping the seal gate when the read path is known down.**

The rollout notes pin the residual on the 500 ms `sealGateTimeout`: during a gateway (read) outage `writeDown` is legitimately false, so `ConfirmSeal` never takes the fast `writeDownVerdict` path and instead spends the full budget waiting for an ack that can't arrive without anchoring. Agreed on the diagnosis. Where I'd push is the conclusion that we have to keep paying it.

The gate exists to stop us broadcasting a block whose height we've already lost — an equivocation self-check on top of consensus. But that protection is a *value-add over consensus*, not a safety floor. Consider the baseline: **production runs today with no sequence store at all.** If two BPs share a signing key and produce simultaneously, we fork and consensus resolves it with a reorg. That's the accepted status quo.

The sequence store improves on that baseline (its CAS election coordinates who owns a height). But when the store is *down*, that coordination is gone regardless of what the gate does — so seal-immediately degrades to exactly today's behavior: a possible fork, resolved by consensus. **It is not strictly worse than what we ship today**, and it buys back the full slot (→ ~1.0 block/s). Holding the gate during a store outage tries to preserve a guarantee whose coordination substrate is already gone, at a throughput cost, for no safety gain over consensus.

On the objection in the notes — that `endStale` unanchors and STALE means contention, so "skip when unanchored" disarms the gate exactly during contention: agreed, "unanchored" is the wrong signal. But **"read breaker open" is a different and clean signal.** The breaker only opens on *silent* reads (timeout / `Unavailable`); a contended-but-healthy store still *answers* (NOT_FOUND included), so it never opens the breaker. So `readBreaker.isOpen()` means "the store can't inform us," never "the store says I'm contested" — which is precisely the "no verdict is coming from the store" signal the notes say is missing.

Concretely, alongside the existing `writeDown` check in `ConfirmSeal`:

```go
if p.writeDown.Load() || p.read.breaker.isOpen() {
    return p.writeDownVerdict() // no store verdict is coming; consensus is the arbiter
}
```

Keep the gate fully armed when the store is up — the skip is justified only by baseline-equivalence during an outage, not in general.

If there's a reason the store-up equivocation gain doesn't transfer to the outage case, I'd like to understand it — but from the baseline argument I can't see one.

pratikspatil024:
Agreed, and done — it closes F3. One carve-out, described at the end.

**Your signal claim checks out.** `isSilent` treats only `DeadlineExceeded` / `Unavailable` / `Canceled` as silence; everything else, `NOT_FOUND` included, falls through to `default: return false` and closes the breaker via `answered()`. So an open breaker cannot mean "the store says you lost" — only "the store did not answer." That is exactly the signal the notes said was missing.

**And the wait was measurably resolving nothing.** In the F3 window of the run in the notes (47 blocks, gateway paused): `sequencer_gate_unknown=21` and seal `max=4002ms`. 21 of 47 blocks paid the full budget and settled `SealUnknown` — no verdict, just the deadline — and the contested ones paid `contestedGateTimeout` (4 s) to reach the same place. `sequencer_publish_barriertimeout=29` in the same window says why no ack was coming even with a healthy ingress: `reconcile` → `readTail` is a read, so with the read path silent the transport cannot re-anchor, and nothing publishes to be acked.

**Re-measured on a devnet with the change** (fresh enclave, warm baseline, 4 validators, 1 s slots):

| window | block/s | seal mean | seal max |
| --- | --- | --- | --- |
| baseline | 1.00 | 4 ms | 13 ms |
| **gateway frozen (F3)** | **0.98** | **2 ms** | **11 ms** |
| ingress frozen (F2) | 0.95 | 192 ms | 531 ms |
| recovered | 1.00 | 4 ms | 10 ms |

F3 was 0.48 unfixed and 0.78 with the breaker alone; it is now indistinguishable from baseline, sealing included. The new counter is the direct evidence rather than an inference: `sequencer/gate/readdownskip` rose **+61 across 59 blocks** in that window (`read/skipped` +87, `gate/unknown` +60) and stayed flat through the baseline — `read/skipped` read 80 both at the previous run's end and at this baseline, so the breaker was closed and the arm never fired against a healthy store.

**The carve-out: a withheld seal keeps its full wait.** `refuseOnTimeout` gates are the one case where the refusal does not depend on an answer still owed — `sealWithoutPublishLocked` sets it from a read that already found the height closed in the store with content this block does not carry. Skipping there would broadcast over an acked generation, which is the displacement the gate exists to prevent, and keeping it costs nothing because that refusal needs no store round trip. The state is narrow (the build-start read has to have succeeded and the breaker opened before the seal) but reachable, so it is guarded and pinned by a test.

So the fast path is now:

```go
if skip := p.storeDownSkip(); skip != nil {
    return p.storeDownVerdict(skip) // no verdict is coming; do not wait for one
}
```

`storeDownSkip` returns `nil` for a withheld seal, `gateWriteDownSkip` when `writeDown` is set, and the new `gateReadDownSkip` when the breaker is open. Two counters rather than one so the dashboard separates the causes: a seal abandoned because no ack could be delivered is a different outage from one abandoned because the store could not be asked.

**One residual worth naming.** An ack already delivered still wins — `storeDownVerdict` runs `storedVerdict` before anything else. What we give up is an ack *in flight* at the instant we stop waiting. Your baseline argument is the right cover for that sliver, and it is the only thing traded.

I did check whether the trade could be avoided altogether, on the theory that a build blinded at its start is held and so can never be acked. It can't be avoided: `holdNewLocked` gates the entries, but the hold clears at seal time (`publish.go`), so a held build's entries do flush and an ack can come back. The baseline argument is load-bearing here, not decorative.

Tests: `TestFrozenReadPathDoesNotHoldTheSealGate` (fails without the change — 402 ms of a 400 ms budget) and `TestReadDownGiveUpKeepsAWithheldSealRefusal` (fails without the carve-out — returns `SealUnknown` where a withheld seal must refuse).

cffls (approved):
Looks good! Thanks for addressing the comment.

PR #2417, opened 2026-09-16
eth, eth/fetcher: bound witness page count by the gas-derived ceiling instead of a cross-peer vote

lucca30:
## What

The witness page-count check polled random peers for a majority page count and, on disagreement, **dropped and jailed** the reporting peer.

Witness page counts are **non-deterministic** across honest nodes: a valid witness can round to one more 15 MiB page on one node than on another (the trie-node set collected during execution varies). So the cross-peer vote both:
- false-positively disconnected and jailed honest peers whose witness was one page larger than the sampled majority, and
- failed open when the sampled peers did not have the witness.

## How

Replace the vote with a **local bound**: a page count within the gas-derived threshold (already ~10× the real per-block witness size) is accepted; a count above it exceeds what the block gas limit can plausibly produce, so it is refused for that peer — the fetch tries another peer — **without disconnecting or jailing**.

- `CheckWitnessPageCount` no longer takes peer-query closures and no longer votes or jails; it accepts ≤ threshold and refuses above.
- Removes `verifyWitnessPageCountSync`, `getConsensusPageCountWithOriginal`, the `witnessManager` jail hook, the `BlockFetcher` `jailPeer` plumbing that only fed it (down to the `NewBlockFetcher` signature), and the now-unused verification meters/const.
- Structural per-response page checks (invalid page number, inconsistent `TotalPages` within a single peer's stream) are unchanged.

### Review follow-up: making the refusal actually stop the download (a27aff475)

Review found that `receiveWitnessPage` still surfaced a refusal as an error. The call site discards that error, and the deferred retry handler then marked page 0 retryable and rebuilt requests from `witTotalPages` — which the refusing page had just populated — so every remaining page of the refused witness was requested from the peer we had just refused, page 0 was retried, the duplicate tripped the "more pages than TotalPages" violation and jailed the peer for an unrelated reason, and the reconstructed witness could still be delivered. Before the vote was removed this was masked by the synchronous peer drop.

- On refusal: pause the hash, discard its pages, log, and `continue` — never return an error, so the retry path cannot re-queue the refused download.
- Pages arriving for an already-paused hash are discarded instead of accumulated, so late in-flight pages can neither complete a refused witness nor re-trigger request building.
- `buildWitnessRequests` takes `downloadPaused` and skips paused hashes for both fresh pages and retries.
- The page count is bounded before the page is stored; the inconsistent-`TotalPages` pause is taken under `mapsMu` like the other writers.
- Stale "dropping peer" log and "trigger peer drop" comment replaced.

## Relationship to the WIT2 signed path

This is the WIT1 page-count layer. It is independent of, and complementary to, the WIT2 signed-size acceptance change (#2416): together they make both witness-delivery gates tolerant of non-deterministic witnesses. Both branches merge cleanly into `v2.10.2-candidate`, and cleanly with each other in either order (verified with `git merge-tree`).

## Tests

- Removed the vote-era page-count tests (they asserted the cross-peer vote/jail behavior being removed).
- Added `TestCheckWitnessPageCountAcceptsWithinThreshold` and `TestCheckWitnessPageCountRefusesAboveThresholdWithoutDrop`.
- Rewrote `TestConcurrentWitnessVerification` to the new signature (still a race test).
- Follow-up: `TestRequestWitnessesWithVerification_RefusedPageCountStopsDownload` drives `RequestWitnessesWithVerification` end-to-end with a mock peer claiming `TotalPages=4` and a refusing bound, and asserts only page 0 is requested, the jail callback never fires, and the request resolves empty. Against the previous code it fails with exactly the reviewer's repro (pages `0,1,2,3,0` requested). `TestBuildWitnessRequests_SkipsPausedHash` pins the request builder.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./eth/ ./eth/fetcher/`
- [x] `golangci-lint run ./eth/...` (0 issues)
- [x] `go test ./eth/fetcher/`
- [x] `go test ./eth/ -run 'Witness|Wit2|PageCount|Announce|Broadcast|Relay|RequestWitnesses|BuildWitnessRequests'`
- [x] `go test -race ./eth/ -run 'RequestWitnessesWithVerification|BuildWitnessRequests' -count=3`


lucca30:
@claude review

cffls:
Reviewed the diff and traced the refusal path end-to-end. The direction is right (the cross-peer vote misclassifies honest ±1-page rounding as lying), but the "refuse without disconnecting or jailing" semantics don't hold yet, because the caller in `eth/peer.go` doesn't act on the `false` the way the description says.

**What happens on `CheckWitnessPageCount == false`**

In `receiveWitnessPage` (`eth/peer.go`), the `!isHonest` branch sets `downloadPaused[hash] = true` and returns an error. Two things then undo that:

1. The call site at `eth/peer.go:385` discards the return value, so the error never reaches anything.
2. The function's deferred error handler marks the page as failed/retryable and calls `buildWitnessRequests`, which doesn't consult `downloadPaused`. Since `witTotalPages[hash]` was already populated before the check ran, it builds requests for every remaining page of the oversized witness from the same peer, plus a retry of page 0.

Before this PR that was masked: `verifyWitnessPageCountSync` called `parentDropPeer` synchronously, so the follow-up requests failed against an already-disconnected peer.

**Repro**

Mock peer via the existing `testPeer(t)` harness returning `TotalPages=4` on every page, `verifyPageCount` always `false`, jail callback counting calls. 5 runs:

- pages 0..3 were all requested from the refused peer, page 0 twice
- the jail callback fired 2–4 times per run (the duplicate page 0 trips the "more pages than TotalPages" violation)
- in 2/5 runs the reconstructed witness was delivered on `dlResCh` anyway

So the observed behavior is "download the whole thing from the peer we just refused, then jail it for an unrelated reason", and sometimes import it.

**Suggested fix**

- On `!isHonest`, stop the download without going through the retry defer: pause, then `return nil` (or close `cancelCh`), so nothing re-queues.
- Make `buildWitnessRequests` honor `downloadPaused` so the error path can't rebuild requests for a paused hash.
- Update the now-stale log `"Peer failed verification, dropping peer"` and comment `// Return error to trigger peer drop` at `eth/peer.go:571-580`.
- Add a test at the `RequestWitnessesWithVerification` level; the new tests only exercise `CheckWitnessPageCount` in isolation.

Otherwise: builds, vets, and `go test ./eth/fetcher/ ./eth/` pass locally, and this merges cleanly with #2416.

cffls (approved):
lgtm!

PR #2427, opened 2026-09-18
eth/sequencer, rawdb: record served preconfirmations to close the open-block loss window

kamuikatsurgi:
## What

`bor_getInvalidPreconfBlocks` must never omit a block whose preconfirmation this node served and then broke. The store audit (#2388) closes the gap for a node's downtime by comparing the store's sealed generation at each height against canonical — so it can only judge heights the store still holds a seal for.

One window escapes it: the producer serves preconfirmations for an open block, then dies before sealing it, and its successor rebuilds the height without the store. The store holds no seal (or only an unsealed open), so the audit has nothing to compare, advances past the height, and the broken preconfirmation is never recorded — the severe case, an invalid preconf absent from the ledger.

## Fix

**Record the promise durably, at serve time.** When receipts are about to reach callers (`indexExecutedTransactions`), the node persists a per-height commitment to its own chaindb: the count of transactions preconfirmed there and a keccak fold over their hashes in served order (`WritePreconfServed`). It lives in a keyspace off the block path, so it survives the crash that loses the store generation.

**Reconcile it when the store cannot confirm a height** — it holds nothing (`NOT_FOUND`), holds a generation it never sealed, returns a seal that is unusable (undecodable or wrong-height), or the height falls in a retention-skipped range. The audit folds the canonical block's leading transactions the same way and compares; a divergence is recorded as `served_mismatch`, a distinct reason from `unobserved_mismatch` because the node did serve this one. Commitments are cleared once judged.

## Notes / boundaries

- The fallback digest is node-local, compared only against itself (serve vs audit), so it folds the already-cached `tx.Hash()` rather than the on-wire `commitment` scheme — cheaper on the serve path, no tag or chain seed.
- Only the latest generation per height is held; a superseded generation's promise is left to the live path, as before.
- Read-only monitor: it records a marker and does not touch fork choice or block acceptance.

## Testing

- Unit/rawdb: round-trip and range reads for the commitment (incl. malformed value); audit records `served_mismatch` for `NOT_FOUND`, unsealed, retention-skip, and unusable-seal cases; kept promises and store-confirmed heights record nothing; commitments cleared once judged. `go test -race` clean on `core/rawdb` and `eth/sequencer`.
- Durability (deterministic, 5/5): `TestServedCommitmentSurvivesCrashAndIsAudited` — real on-disk pebble crash (close/reopen); no commitment → unrecorded (the bug), durable commitment → recorded.
- Live devnet A/B (custom bor image, deterministic fault injection): serve a divergent preconf at height H, fsync, crash (`SIGKILL`-equivalent), restart, store returns `NOT_FOUND` for H.

  ```
  height  reconcile   bor_getInvalidPreconfBlocks   servedmismatch
  171     off (pre)   []  UNRECORDED (the bug)       0
  245     on  (fix)   0xf5  RECORDED                 1
  319     on  (fix)   0x13f RECORDED                 1
  393     on  (fix)   0x189 RECORDED                 1
  ```

  Trace of one fixed run (RPC bor logs):
  ```
  TESTHOOK: persisted served commitment height=736 count=3   (write, pre-crash)
  Rewound to block with state number=535                     (crash + restart)
  TESTHOOK: audit read served commitment height=736 ok=true  (survived crash)
  servedmismatch=1                                           (reconciled -> recorded)
  ```


eth/sequencer/audit.go
```diff
@@ -280,26 +302,151 @@ func (a *auditor) auditHeightInto(ctx context.Context, height uint64, summary *a
...
+	}
+}
+
+// servedMatchesCanonical reports whether the count transactions this node
+// preconfirmed at a height are the canonical block's leading transactions, in
+// the same order. A missing block or one with fewer transactions than were
+// served is a mismatch: content was promised the chain did not keep.
+func (a *auditor) servedMatchesCanonical(height, count uint64, digest common.Hash) bool {
+	block := a.chain.GetBlockByNumber(height)
+	if block == nil {
```
pratikspatil024, line 440: The doc comment makes this deliberate ("a missing block ... is a mismatch: content was promised the chain did not keep"), so treat this as a question rather than a defect: `GetBlockByNumber` returns nil both when the chain has no such height *and* when the header is canonical but the body isn't available locally — pruned, or a snap-sync gap. The second case is "we cannot judge", and recording `served_mismatch` for it puts a false entry in the ledger, which is the one thing scenario 1 asks us not to do spuriously.

The pass being finality-bounded rules out "above head", so the exposure is narrow — but `reconcileSkippedServed` concentrates it: that path exists to judge heights the walk skipped for *store retention*, i.e. the oldest heights in range, which is exactly where local bodies are most likely gone. After a long store outage that range can reach back hundreds of heights.

Would suggest splitting the two: nil block → counted as unknown and the commitment left in place (or dropped without a record), and only a block that actually carries different leading transactions → `served_mismatch`. Same shape as a bug I hit on #2406, where `chainVerdict` refused a seal because height 0 compared genesis against an empty hash — the "can't tell" case reads as a verdict if you let it.

eth/sequencer/exec.go
```diff
@@ -499,7 +500,12 @@ func (s *session) indexExecutedTransactions() ([]*types.Log, core.PreconfReceipt
 		Receipts:     append(types.Receipts(nil), s.env.receipts[start:]...),
 		Transactions: append(types.Transactions(nil), s.env.txs[start:]...),
 	}
+	// Record the promise durably before returning: these receipts are about
+	// to reach callers, and a crash after this must leave the audit able to
+	// judge the height even if the store loses the generation.
+	s.env.servedDigest = foldServed(s.env.servedDigest, s.env.txs[start:])
 	s.env.indexedTxs = len(s.env.txs)
+	s.consumer.persistServed(s.env.header.Number.Uint64(), uint64(s.env.indexedTxs), s.env.servedDigest)
```
pratikspatil024, line 515: This is a synchronous write on the path that serves preconfirmation receipts, and it runs once per re-execution batch rather than once per height, so it's several writes per block. Almost certainly sub-millisecond into the memtable, and the digest is folded in memory so there's no read-modify-write — but preconf latency is the product here, so I'd rather see it measured than assumed. A p50/p99 on receipt serving with and without would settle it.

Worth stating the durability argument explicitly in the PR too, because it looks like it needs an `fsync` and doesn't: `SIGKILL` kills the process, not the kernel, so a KV write that has reached the OS survives it. That's why the 3/3 A/B works without a sync, and it's the reason this approach is cheap. Only a machine-level loss would need more.

pratikspatil024 (commented):
Good find, and the diagnosis is right — the gap is real and it's in my code. `auditHeight` returns `auditNoSeal` when a generation has no seal, and `recordVerdict` groups that with `auditMatch`, so the height is walked, nothing is recorded, no counter moves, and the watermark advances over it. Of the three ways a height can get past that pass, `retentionskipped` and `unheld` at least increment something; `auditNoSeal` was the one that was completely silent, and it's the one you hit.

I'd also had the gap scoped too narrowly in my own notes — I'd written it down as "the producer died before publishing at all", where the real condition is "no *seal* in the store", which includes published-but-unsealed. That's the far more likely crash shape, so thanks for pinning it properly.

The approach looks right to me: it extends the pass rather than replacing it, which keeps the store-side comparison covering preconfs served by *other* nodes while the local commitment covers what the store never sealed. Nice handling of the cases beyond the headline one — reconciling on `auditUnknown` instead of clearing, so an undecodable or wrong-height seal can't bury a broken promise before the watermark passes it, and having `reconcileSkippedServed` scan the served prefix rather than every skipped height.

Two inline comments, neither blocking: the `block == nil` path reading as a mismatch when it's really "can't judge", and the serve-path write being unmeasured. Prefix-fold semantics look correct to me — folding canonical's leading `count` transactions means an insertion or reorder anywhere in the served prefix trips it, which is exactly the `transactionIndex 2 → 3` from run 04.

eth/sequencer/audit.go
```diff
@@ -280,26 +302,151 @@ func (a *auditor) auditHeightInto(ctx context.Context, height uint64, summary *a
...
+		return false
+	}
+	txs := block.Transactions()
+	if uint64(len(txs)) < count {
+		return false
+	}
+
+	return foldServed(common.Hash{}, txs[:count]) == digest
```
cffls: This folds only `tx.Hash()` in served order, so it agrees with canonical whenever the leading `count` transactions are the same ones in the same order — regardless of what they *did*. The live path this stands in for is stricter: `entryMatchesCanonical` requires `sameExecutionContext` (`ParentHash`, `Number`, `Time`, `GasLimit`, `BaseFee`, `Difficulty`) and `sameReceiptPrefix` (`Status`, `CumulativeGasUsed`, `Bloom`, `Logs`) as well.

That difference is live on the very event this PR is about. The open-block loss window is a producer handover, and a successor's rebuild of the height carries a different header context by construction — `Difficulty` is succession-based (`Difficulty(snap.ValidatorSet, signer)`), and ordinarily `Time` and the parent can differ too if H−1 was contested during the handover. Any served transaction whose execution read that context (a `block.timestamp` deadline is the common case), or that ran against a different pre-state, has a different `Status`/logs canonically than the receipt this node handed out. Same hashes, same order → same fold → `auditServedVerified` → the broken preconfirmation is not recorded. None of the new tests exercise this: they all reorder, drop, or replace transactions.

Suggested fix, cheapest first:

1. Seed the fold with the header context the live path checks, so the two paths agree by construction: `foldServed(contextSeed(header), txs)` where the seed hashes `ParentHash`, `Number`, `Time`, `GasLimit`, `BaseFee`, `Difficulty` — computed from `s.env.header` on the serve side and `block.Header()` here. One extra keccak per height. Over-recording when the context changed but execution happened to coincide is the direction this ledger's contract prefers ("never omit a broken promise").
2. Or fold the receipts' consensus fields (`Status`, `CumulativeGasUsed`) per transaction. Exact, no over-recording, but needs canonical receipts at audit time.

Either way, please add `TestAuditReconcilesServedMismatchOnDifferentContext` (same leading txs, different parent or `Time`) alongside the reorder/drop cases.

eth/sequencer/exec.go
```diff
@@ -499,7 +500,12 @@ func (s *session) indexExecutedTransactions() ([]*types.Log, core.PreconfReceipt
 		Receipts:     append(types.Receipts(nil), s.env.receipts[start:]...),
 		Transactions: append(types.Transactions(nil), s.env.txs[start:]...),
 	}
+	// Record the promise durably before returning: these receipts are about
+	// to reach callers, and a crash after this must leave the audit able to
+	// judge the height even if the store loses the generation.
+	s.env.servedDigest = foldServed(s.env.servedDigest, s.env.txs[start:])
```
cffls, line 513: If the fold is seeded with header context (inline 1), this is where the serve side needs to match: initialise `servedDigest` from `s.env.header` once (at `newBlockEnv`, or when `indexedTxs == 0`) rather than from the zero hash, so the incremental per-batch fold and the audit's one-shot fold start from the same seed. The header is fixed for the life of the env, so the incremental structure is unaffected.

eth/sequencer/audit.go
cffls, line 440: Agree with Pratik that this conflates "can't judge" with "judged broken". `GetBlockByNumber` is nil both for a height the chain does not have and for a canonical header whose body is not local, and `reconcileSkippedServed` steers this path at precisely the oldest heights in range, where a body is most likely gone after a long store outage. A `served_mismatch` there is a spurious ledger entry — the scenario-1 failure mode this endpoint is supposed to avoid.

Suggest returning a three-way result: nil block → count as `auditUnknown`, leave the commitment in place (so a later pass can still judge it if the body arrives) or drop it without a record; only a block that actually carries different leading transactions → `served_mismatch`. `len(txs) < count` stays a mismatch: the block is here and is shorter than what was served.

cffls (changes requested):
Confirmed the gap against the base independently of the write-up: `auditHeightInto` treats `NOT_FOUND` as "nothing was promised", `recordVerdict` no-ops on `auditNoSeal`, and the consumer's served state is purely in-memory (the only durable consumer-side writes are invalidations, written at invalidation time) — so a node that served a preconf and then lost the process leaves the broken promise unrecorded. Scenario 2 is real, and this PR targets exactly the three branches that let it through, plus the retention-skipped range that nobody would have noticed leaking. Writing the commitment before the receipts reach callers, `IfAbsent` so the live path's stronger record keeps precedence, treating a malformed record as an error rather than "nothing served", and refusing to let an unusable seal clear a commitment are all the right calls.

One thing I'd change before merge: the commitment is weaker than the live judgement it stands in for. `foldServed` commits to transaction identity and order only, while `entryMatchesCanonical` also checks the execution context (`ParentHash`, `Time`, `GasLimit`, `BaseFee`, `Difficulty`) and the receipts. That leaves a slice of scenario 2 open on exactly the handover this fix targets — details inline at `audit.go:421`. Cheap to close (seed the fold with the header context) and it makes the two paths agree by construction.

+1 to Pratik's `block == nil` point; inline at `audit.go:413`.

Also: codecov is at 86.5% patch vs the 90% gate, mostly `consumer.go` (57%) — the different-parent test below plus a test for the live-path clear in `markCanonicalHeadAudited` should carry it over.

eth/sequencer/audit.go
kamuikatsurgi: Fixed in c92eedf. `judgeServedAgainstCanonical` is now three-way: a nil block (chain lacks the height, or a canonical header whose body is pruned / snap-gapped) returns `servedUnjudgeable` — counted uncomparable (`auditUnknownCount`), commitment left in place for a later pass, and no record written. Only a present-but-divergent block records `served_mismatch`; `len(txs) < count` stays a mismatch. So the "can't judge" case can no longer put a spurious entry in the ledger. Test: `TestReconcileServedLeavesUnjudgeableCommitment`.

eth/sequencer/exec.go
kamuikatsurgi: Added a `sequencer/preconf/servedpersist` timer (c92eedf) so the write is measured rather than assumed. On durability: agreed — it's a NoSync KV write, so a real (arbitrarily-timed) SIGKILL survives it once the WAL record has reached the OS, which happens continuously under load; no fsync needed. It's best-effort by design: a lost write only costs the audit its fallback for that one height, and the live path's stronger record covers the healthy case. (One clarification: the devnet A/B harness fsync'd at the injected `os.Exit` to make the instant crash deterministic; production relies on the OS-flush behaviour you describe.) The timer lets us capture p50/p99 on a devnet run — happy to attach that if useful.

eth/sequencer/audit.go
kamuikatsurgi: Fixed in c92eedf (option 1 — the context seed). `foldServed` now starts from `contextSeed(header)`: a keccak over ParentHash, Number, Time, GasLimit, BaseFee, Difficulty — exactly the fields `sameExecutionContext` checks. The serve side seeds from `s.env.header` on the first batch and the audit from the canonical block header, so the two paths agree by construction and a handover's different context (succession-based Difficulty, and often Time/parent) trips the fold even with identical leading txs. Went with the seed over folding receipts to avoid needing canonical receipts at audit time. Test: `TestAuditReconcilesServedMismatchOnDifferentContext` (same leading txs, different Difficulty).

eth/sequencer/exec.go
kamuikatsurgi: Done. `servedDigest` is seeded from `contextSeed(s.env.header)` on the first batch (`start == 0`) in `indexExecutedTransactions`, so the incremental per-batch fold starts from the same seed as the audit's one-shot fold. The header is fixed for the env's life, so the incremental structure is unchanged.

eth/sequencer/audit.go
kamuikatsurgi: Fixed in c92eedf (same as the reply on Pratik's thread): nil block → `servedUnjudgeable`, counted uncomparable with the commitment left in place and no record; only a present-but-divergent block records, and `len(txs) < count` stays a mismatch. Test: `TestReconcileServedLeavesUnjudgeableCommitment`.

kamuikatsurgi:
Thanks both — all addressed in c92eedf:

- **Execution-context fold** (@cffls): `foldServed` is now seeded with `contextSeed(header)` (ParentHash, Number, Time, GasLimit, BaseFee, Difficulty — the `sameExecutionContext` fields), seeded identically on serve and audit, so a handover context change trips the fold even with identical leading txs. + `TestAuditReconcilesServedMismatchOnDifferentContext`.
- **`block == nil` conflation** (@ppatil, @cffls): `judgeServedAgainstCanonical` is three-way — nil/pruned body → uncomparable (counted, commitment left, no record); only a present-but-divergent block records. + `TestReconcileServedLeavesUnjudgeableCommitment`.
- **Serve-path write** (@ppatil): added a `sequencer/preconf/servedpersist` timer; durability rationale in-thread.
- **Coverage**: added the different-context, unjudgeable, live-path-clear, `persistServed`, and `ReadPreconfServed` read-failure tests.

Ready for another look.

cffls (approved):

